PR #31, opened 2023-07-25
PORT-4343 resolving error for unknown default file

yairsimantov20:
# Description

What - The initializePortResources failed if there is an unknown file in the  resources folder
Why - Ocean raise an error for unknown file types without checking if this file is known
How - Moved the check of the file name before the file type

## Type of change


Tankilevitch (approved):
LGTM

PR #38, opened 2023-07-26
Allow populating config with env vars

yairsimantov20:
# Description

What - We need to enable a way to pass the config with env vars
How - users can pass the env vars in a SNAKE_CASE formatting and __ for a nested key

## Type of change


.github/workflows/release-integrations.yml
```diff
@@ -4,6 +4,11 @@ on:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      LATEST:
```
dvirsegev, line 8: should be always the latest no?

integrations/gitlab/pyproject.toml
```diff
@@ -10,7 +10,7 @@ aiofiles = "^0.6.0"
 python-gitlab = "^3.14.0"
 pathlib = "^1.0.1"
 jsonschema = "^4.17.3"
-port-ocean = {version = "0.1.1", extras = ["cli"]}
+port-ocean = {version = "0.1.2rc3", extras = ["cli"]}
```
dvirsegev, line 13: are you sure?

dvirsegev (approved):
LGTM

.github/workflows/release-integrations.yml
yairsimantov20: when im building manually i want the option to build images for development

integrations/gitlab/pyproject.toml
yairsimantov20: will be change after this pr
i need to publish 0.1.2

PR #42, opened 2023-07-30
Using the current ocean version for ocean new deps

yairsimantov20:
# Description

What - The current scaffold ocean version is outdated
Why - The ocean version is a static string
How - Using the current ocean version

## Type of change


danielsinai (approved):
🦢 Nice Catch!

PR #50, opened 2023-08-03
Log traceback of raised exceptions in resync methods

Tankilevitch:
# Description

What - Log traceback of raised exceptions in resync methods
Why - Until this fix when error occurred in the resync method it was really hard to understand the cause of the failure as only the error was logged without any information on what caused it 
How - Add `logger.exception` in the `resync_error_handling` method

💡 **Had to add a workaround for resolving `Exception` subclass due to an open issue in loguru** 

## Type of change


port_ocean/core/integrations/mixins/utils.py
```diff
@@ -20,6 +21,8 @@ def resync_error_handling() -> Generator[None, None, None]:
     except StopAsyncIteration:
         raise
     except Exception as error:
+        # logger.exception is used to log the exception with the traceback
+        logger.exception("Failed to execute resync function")
```
yairsimantov20, line 25: There might be an error with this

yairsimantov20 (changes requested):

port_ocean/core/integrations/mixins/utils.py
Tankilevitch: handled it

yairsimantov20 (approved):
🌊

PR #52, opened 2023-08-06
Pass resource config inside the event context

Tankilevitch:
# Description

What - Pass resource config inside the event context
Why - In case multiple resource kinds were specified inside the app config, the resync method couldn't know for which resource block ocean referred to, by passing the resource config inside the context, the integration gets the resource config that is relevant without the need for any extra logic 
How - Create event context per resource before running the resync method

## Type of change


yairsimantov20 (approved):
🌊 Looks good

PR #54, opened 2023-08-07
Fix entity jq parsing

yairsimantov20:
# Description

What - If the identifier was not found in the jq it would set a "None" as the identifier which is invalid because we dont want to create the wrong entity
Why - The fields went through a parsing that made them string if they are not
How - Changed the typing and remove the parsing

## Type of change

### DIFF
--- a/changelog/bugfix.md
+++ b/changelog/bugfix.md
@@ -1,1 +1,1 @@
-Fixed an issue that caused the jq None values for relations to become a string of None
+Fixed an issue that caused the jq `None` values for relations to become a string with the value `"None"` instead of being interpreted as `null` in JSON
--- a/changelog/bugfix.md
+++ b/changelog/bugfix.md
@@ -1,0 +1,1 @@
+Fixed an issue that caused the jq None values for relations to become a string of None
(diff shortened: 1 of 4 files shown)


port_ocean/core/models.py
```diff
@@ -7,32 +7,13 @@
...
 
 class Entity(BaseModel):
-    identifier: str
-    blueprint: str
-    title: str | None
-    team: str | list[str] = []
+    identifier: Any
+    blueprint: Any
+    title: Any
+    team: str | None | list[Any] = []
```
MPTG94, line 13: Why list[Any]? teams are supposed to be strings

changelog/bugfix.md
```diff
@@ -0,0 +1 @@
+Fixed an issue that caused the jq None values for relations to become a string of None
```
MPTG94, line 1:
```suggestion
Fixed an issue that caused the jq `None` values for relations to become a string with the value `"None"` instead of being interpreted as `null` in JSON
```

MPTG94 (approved):
LGTM, please note my commetns

PR #55, opened 2023-08-07
Pagerduty

PeyGis:
# Description
Added who is on call to the service. I also changed the service URL to use the actual data from Pagerduty

What - Improving on service blueprint
Why - The previous implementation did not have oncall to the service blueprint
How - Made API call to Pagerduty's who is on call

## Type of change


integrations/pagerduty/.port/resources/port-app-config.yaml
```diff
@@ -10,6 +10,8 @@ resources:
           blueprint: '"pagerdutyService"'
           properties:
             status: .status
+            url: .html_url
+            oncall: .oncall_user[0].user.email
```
danielsinai, line 14: Can you verify that if the oncall user is empty it's inserting null?

integrations/pagerduty/clients/pagerduty.py
```diff
@@ -131,3 +137,50 @@ async def create_webhooks_if_not_exists(self) -> None:
...
+        escalation_policy_ids = [
+            service["escalation_policy"]["id"] for service in data[data_key]
+        ]
+
+        oncall_users = await self.get_oncall_user(",".join(escalation_policy_ids))
+        all_data = []
+
+        for service in data[data_key]:
+            escalation_policy_id = service["escalation_policy"]["id"]
+
```
danielsinai, line 176: What will happen if my service doesn;t have an escalation policy?

danielsinai (changes requested):
Great job, left some questions & comments 🌊

integrations/pagerduty/.port/resources/port-app-config.yaml
PeyGis: This is one of the blueprint entity data when the oncall_user is unavailable

```json
{
  "identifier": "PWJAGSD",
  "title": "Port Internal Service Updated",
  "blueprint": "pagerdutyService",
  "team": [],
  "properties": {
    "status": "critical",
    "url": "https://getport-io.pagerduty.com/service-directory/PWJAGSD"
  },
  "relations": {},
  "createdAt": "2023-08-07T07:56:01.687Z",
  "createdBy": "hBx3VFZjqgLPEoQLp7POx5XaoB0cgsxW",
  "updatedAt": "2023-08-07T08:14:54.954Z",
  "updatedBy": "hBx3VFZjqgLPEoQLp7POx5XaoB0cgsxW"
}

```

integrations/pagerduty/clients/pagerduty.py
PeyGis: From the UI, it appears that the escalation policy is always **required** when creating a service
image

Also, from the [Create Service API](https://developer.pagerduty.com/api-reference/7062f2631b397-create-a-service), the escalation_policy is a required property

integrations/pagerduty/.port/resources/port-app-config.yaml
PeyGis: Also, we see that even when the oncall_user is empty, the jq mapping returns null

image

danielsinai (approved):

PR #57, opened 2023-08-08
Support list of entities for generators

yairsimantov20:
# Description

What - The current generators generators are realy slow
Why - The generator yield one object at a time
How - allowing returnning a list of entities

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)


changelog/breaking.md
```diff
@@ -0,0 +1 @@
+Changed the on_resyn generator to expect a list instead of a single item
```
MPTG94, line 1:
```suggestion
Changed the `on_resync` generator to expect a list instead of a single item
```

MPTG94: Shouldn't this say: Updated the `on_resync` generator to support both a single item and a list of items

changelog/improvement.2.md
```diff
@@ -0,0 +1 @@
+Displaying the extra object only for debug log level
```
MPTG94, line 1:
```suggestion
Changed the extra object messages log level from INFO to DEBUG
```

changelog/improvement.md
```diff
@@ -0,0 +1 @@
+Removed a confusing error log at the integration installation that says the integration does not exists
```
MPTG94, line 1:
```suggestion
Removed a wrongful error log at the integration installation that says the integration does not exists
```

MPTG94 (approved):
LGTM

PR #60, opened 2023-08-09
reverted a the move of the httpx client

yairsimantov20:
# Description

What - We dont want users to play with the requests to port
Why - It might not be that bad to let the clients handle their own httpx clients
How - reverted a change

## Type of change


Tankilevitch (approved):
For a followup pr I suggest we create another global httpx client for the integration that will be accessible to the integration (could be a feature flag or something) that way we encourage the integration developer to re-use the client but still it doesn't effect the port httpx client

yairsimantov20:
> For a followup pr I suggest we create another global httpx client for the integration that will be accessible to the integration (could be a feature flag or something) that way we encourage the integration developer to re-use the client but still it doesn't effect the port httpx client

https://getport.atlassian.net/browse/PORT-4460

PR #71, opened 2023-08-11
optional webhook

yairsimantov20:
# Description

What - the app host is failing the integration to start if not set
Why - the property is required and sometimes the gitlab does not have a webhook for the groups in gitlab
How - made it so the feature is optional

## Type of change


integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -23,7 +25,22 @@ async def handle_webhook(group_id: str, request: Request) -> dict[str, Any]:
 
 @ocean.on_start()
 async def on_start() -> None:
-    setup_application()
+    logic_settings = ocean.integration_config
+    if not logic_settings.get("app_host"):
+        logger.debug(
```
MPTG94, line 30: This log message such be at warning level

MPTG94 (approved):
LGTM

PR #73, opened 2023-08-13
Handle delete events from change log where there is no after

Tankilevitch:
# Description

What - Handle where the integration receive an event of deletion from the port's Kafka
Why -
How - Check whether there is after

## Type of change


yairsimantov20 (approved):
🌊 LGTM

PR #81, opened 2023-08-17
Improved integration template

yairsimantov20:
# Description

What - Removed the ocean version from the scaffolding parameters & wrapped the config.yaml injection with "" in the integration template
Why - The user could override the ocean version to a different value from the current ocean version & the template config.yaml created incorrectly 
How - Using Jinja2 extensions class & fixed the config.yaml

## Type of change


port_ocean/cli/cookiecutter/extensions.py
```diff
@@ -0,0 +1,12 @@
...
+from jinja2.parser import Parser
+
+from port_ocean import __version__
+
+
+class VersionExtension(Extension):
+    tags = {"version"}
+
+    def parse(self, parser: Parser) -> nodes.Node | list[nodes.Node]:
+        return nodes.Const(__version__, lineno=next(parser.stream).lineno)
```
Tankilevitch, line 12: I am less familiar with how this fits in the general flow, I would suggest throwing a comment there with why & how

Tankilevitch (approved):

PR #104, opened 2023-08-28
[Feature][Integration][GitLab] Add support for file search checks

Tankilevitch:
# Description

What - Add support for `search://` filter  for project 
Why -
How -

## Type of change


integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
 from port_ocean.core.handlers.port_app_config.models import PortAppConfig
 
 
-class GitManipulationHandler(JQEntityProcessor):
+class FileEntityProcessor(JQEntityProcessor):
+    prefix = FILE_PROPERTY_PREFIX
+
+    def _search(self, data: Dict[str, Any], pattern: str) -> Any:
+        project_id, ref = self._validate_project_scope(data)
+        # relying on that the code is being called after event initialization and as part of the GitLab service
```
MPTG94, line 24:
```suggestion
        # assuming that the code is being called after event initialization and as part of the GitLab service
```

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+    def _parse_search_pattern(self, pattern: str) -> Tuple[str, str]:
+        """
+        :param pattern: e.g. search://scope=blobs&&query=filename:port.yml
+        :return: scope, search_pattern
+        """
+        # remove prefix
+        pattern = pattern.replace(self.prefix, "")
+        if len(pattern.split(self.separation_symbol)) == 1:
+            raise ValueError(
+                f"Search pattern {pattern} does not contain separation symbol {self.separation_symbol}"
```
MPTG94, line 88:
```suggestion
                f"Search pattern {pattern} does not contain separation symbol: {self.separation_symbol}"
```

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+            scope = splitted_patten[0]
+            query = self.separation_symbol.join(splitted_patten[1:])
+        # default case
         else:
-            return super()._search(data, pattern)
+            scope, query = pattern.split(self.separation_symbol)
+
+        # return the actual scope and query values
+        # e.g. scope=blobs -> blobs, query=filename:port.yml -> filename:port.yml
+        return scope.split("=")[1], "=".join(query.split("=")[1:])
```
MPTG94, line 101: If I understand this correctly, for checks with multiple rules, the query will end up being a `=`-delimited list, is that what's supposed to happen? Can you add an example of an expected query output for a case with multiple conditions?

MPTG94 (approved):
Take a look at my comments to make sure there are no mistakes

integrations/gitlab/gitlab_integration/git_integration.py
Tankilevitch: I join the list back to string after ignoring the first index which is the `query=` part of the string

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+    elif pattern.startswith(SEARCH_PROPERTY_PREFIX):
+        return SearchEntityProcessor
+    else:
+        return JQEntityProcessor
+
+
+class GitManipulationHandler(JQEntityProcessor):
+    def _search(self, data: Dict[str, Any], pattern: str) -> Any:
+        entity_processor = get_entity_processor_by_pattern(pattern)
+        return entity_processor(self.context)._search(data, pattern)
```
yairsimantov20, line 122: I think its a bit over complicated 
Lets talk about it

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+        logger.info(f"Searching for {query} in Project {project_id}, scope {scope}")
+
+        match = False
+        if project:
+            match = bool(project.search(scope=scope, search=query))
+        return match
 
-            file_path = pattern.replace(FILE_PROPERTY_PREFIX, "")
-            return self._get_file_content(gitlab_service, project_id, file_path, ref)
+    def _parse_search_pattern(self, pattern: str) -> Tuple[str, str]:
```
yairsimantov20, line 79: split

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
 
-class GitManipulationHandler(JQEntityProcessor):
+class FileEntityProcessor(JQEntityProcessor):
+    prefix = FILE_PROPERTY_PREFIX
+
+    def _search(self, data: Dict[str, Any], pattern: str) -> Any:
+        project_id, ref = self._validate_project_scope(data)
+        # assuming that the code is being called after event initialization and as part of the GitLab service
+        # initialization
+        project_client: Project | None = event.attributes["PROJECTS_CACHE_KEY"][
```
yairsimantov20, line 26: Use the var for the projects cache key

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
 from pydantic import Field
+from gitlab.v4.objects import Project
 
 from port_ocean.context.event import event
 from port_ocean.core.handlers import JQEntityProcessor
 from port_ocean.core.handlers.port_app_config.models import PortAppConfig
 
 
-class GitManipulationHandler(JQEntityProcessor):
+class FileEntityProcessor(JQEntityProcessor):
```
yairsimantov20, line 19: Why are they inheriting this?
They aint using the jq

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+
+        # return the actual scope and query values
+        # e.g. scope=blobs -> blobs, query=filename:port.yml -> filename:port.yml
+        return scope.split("=")[1], "=".join(query.split("=")[1:])
+
+    @staticmethod
+    def _validate_project_scope(data: Dict[str, Any]) -> Tuple[int, str]:
+        if (project_id := data.get("id")) and (ref := data.get("default_branch")):
+            return project_id, ref
+        raise ValueError("Project id and ref are required")
```
yairsimantov20, line 107: Just make it as function on the top level

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+
+        file_path = pattern.replace(self.prefix, "")
+        return (
+            self._get_file_content(project_client, file_path, ref)
+            if project_client
+            else None
+        )
+
     @lru_cache()
     def _get_file_content(
```
yairsimantov20, line 39: Im not sure the lru is working anymore because ur recreating the class each time

integrations/gitlab/gitlab_integration/git_integration.py
yairsimantov20, line 19: Im not sure there is a need for a full class for this
They can be simple functions

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+    def _parse_search_pattern(self, pattern: str) -> Tuple[str, str]:
+        """
+        :param pattern: e.g. search://scope=blobs&&query=filename:port.yml
+        :return: scope, search_pattern
+        """
+        # remove prefix
+        pattern = pattern.replace(self.prefix, "")
+        if len(pattern.split(self.separation_symbol)) == 1:
+            raise ValueError(
+                f"Search pattern {pattern} does not contain separation symbol: {self.separation_symbol}"
```
yairsimantov20, line 88: I think it can be described better

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+        """
+        project_id, _ = self._validate_project_scope(data)
+        scope, query = self._parse_search_pattern(pattern)
+
+        # assuming that the code is being called after event initialization and as part of the GitLab service
+        # initialization
+        project: Project | None = event.attributes[PROJECTS_CACHE_KEY][project_id]
+        logger.info(f"Searching for {query} in Project {project_id}, scope {scope}")
+
+        match = False
```
yairsimantov20, line 74: Why are u returning false if it does not exists?
When will it happen?

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+        return scope.split("=")[1], "=".join(query.split("=")[1:])
+
+    @staticmethod
+    def _validate_project_scope(data: Dict[str, Any]) -> Tuple[int, str]:
+        if (project_id := data.get("id")) and (ref := data.get("default_branch")):
+            return project_id, ref
+        raise ValueError("Project id and ref are required")
+
+
+def get_entity_processor_by_pattern(pattern: str) -> typing.Type[JQEntityProcessor]:
```
yairsimantov20, line 110: This can be in the search instead
No need for a separate function

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -135,32 +138,46 @@ def create_webhooks(self) -> list[int | str]:
...
+
+        filtered_projects = {
+            project.id: project
+            for project in projects
+            if self.should_run_for_project(project.path_with_namespace)
+        }
+        logger.debug(
+            f"Found {len(filtered_projects)} projects after filtering. Projects: {[proj.path_with_namespace for proj in filtered_projects.values()]}"
+        )
+        event.attributes.setdefault(PROJECTS_CACHE_KEY, {}).update(filtered_projects)
```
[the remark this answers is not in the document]
Tankilevitch: it returns all the projects per service mapping, so I don't want to override but rather update 

I changed it to a dictionary where the key is project id and value the project it self it allows easier cache search

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -1,47 +1,125 @@
...
+        Handles entity mapping for search:// pattern
+        :param data: project data
+        :param pattern: e.g. search://scope=blobs&&query=filename:port.yml
+        :return: True if the search pattern matches, False otherwise
+        """
+        project_id, _ = self._validate_project_scope(data)
+        scope, query = self._parse_search_pattern(pattern)
+
+        # assuming that the code is being called after event initialization and as part of the GitLab service
+        # initialization
```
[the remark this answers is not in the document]
Tankilevitch: no I mean event initialization which means that resync / event handling happened and therefor initialized the get_projects / get_project in the GitLabService

integrations/gitlab/gitlab_integration/git_integration.py
Tankilevitch: In the future we might want it to apply jq query after the special search so I think it will be useful

integrations/gitlab/gitlab_integration/git_integration.py
Tankilevitch: changed default to None, 
I get a list as a response, when no match I get empty list, when matched I get list with the matches, and then I apply bool over the response

integrations/gitlab/gitlab_integration/git_integration.py
MPTG94: How does the list look after the second join? how will it look specifically is there is more than one search clause? (eg. `query=filename:port.yml&&filecontains:requests` - I just made that up but I want to see that it makes sense)

Tankilevitch: So lets take for example the following: `query=pandas==2.6.0 filename:requirements.txt` which is the way you combine two conditions. query all requirements.txt files that contains `pandas==2.6.0` string. 

the output will be:
`pandas==2.6.0 filename:requirements.txt`

MPTG94: Ok got it 👍

yairsimantov20 (approved):
Approved and will schedule a meeting

PR #129, opened 2023-09-12
Gitlab owned flag

danielsinai:
# Description

What - Added owned + visibility and masking in logs
Why - To enable non owner tokens to get projects
How - Enriched the selector

## Type of change


integrations/gitlab/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+0.1.16 (2023-09-13)
+===================
+
+### Improvements
+
+- Added owned & visibility flags to project configurations (#1)
+- Masking token at the startup log (#2)
```
yairsimantov20, line 16: add a bump changelog

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -89,6 +92,12 @@ def _search(self, data: Dict[str, Any], pattern: str) -> Any:
 class GitlabPortAppConfig(PortAppConfig):
     spec_path: str | List[str] = Field(alias="specPath", default="**/port.yml")
     branch: str = "main"
+    filter_owned_projects: bool | None = Field(
+        alias="filterOwnedProjects", default=True
+    )
+    project_visibility_filter: str | None = Field(
+        alias="projectVisibilityFilter", default=None
+    )
```
yairsimantov20, line 100: i think the format should match
decide if you want to start the var with`filter` or end it with it

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -4,14 +4,17 @@
...
 from gitlab.v4.objects import Project
 
 from port_ocean.context.event import event
 from port_ocean.core.handlers import JQEntityProcessor
-from port_ocean.core.handlers.port_app_config.models import PortAppConfig
+from port_ocean.core.handlers.port_app_config.models import (
+    PortAppConfig,
+)
+
+PROJECTS_CACHE_KEY = "__cache_all_projects"
```
yairsimantov20, line 17: that shouldnt be in this file

i understand why you did it but there are other ways to solve such an issue. You did it becuae you couldnt import the GitlabConfiguration class to cast the configuration to it

if you are trying to import something that is only a type without having a circular error you can do it by checking if the code is in typechcking mode with this code

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from gitlab_integration.git_integration import GitlabPortAppConfig
    
```

Then you can use the class by wrapping its usage with `"`

it will look something like this

`typing.cast(obj, "GitlabPortAppConfig")`


You can see something similar in `context/resource.py`

yairsimantov20 (changes requested):

yairsimantov20 (approved):
🌊

PR #133, opened 2023-09-25
Add OpenCost Integration

PeyGis:
# Description

Added data exporter for OpenCost resources using Ocean

## Type of change


talsabagport (changes requested):
Looks good 👍  Left a small comment.

integrations/opencost/main.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def init_client() -> OpenCostClient:
+    return OpenCostClient(
+        ocean.integration_config["app_host"],
+        ocean.integration_config.get("window", "today"),
+    )
+
+
+async def process_cost_item(item: dict[str, Any]) -> list[dict[str, Any]]:
```
talsabagport, line 19: The function is not supposed to be async, and then you can get rid of the `asyncio.gather`

PeyGis: Noted. I've updated the PR

talsabagport (approved):

PR #156, opened 2023-10-15
Changed print in the http error handling to log info

yairsimantov20:
# Description

What - Sonarqube didnt log for http error when paginating requests
Why - used print instead of the logger
How - changed to print

## Type of change

### DIFF
--- a/integrations/sonarqube/client.py
+++ b/integrations/sonarqube/client.py
@@ -1,4 +1,5 @@
 from typing import Any, Optional, AsyncGenerator, cast
+
 import httpx
 from loguru import logger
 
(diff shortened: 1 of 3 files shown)


talsabagport (approved):

integrations/sonarqube/client.py
```diff
@@ -97,7 +98,7 @@ async def send_paginated_api_request(
             return all_resources
 
         except httpx.HTTPStatusError as e:
-            print(
+            logger.info(
```
talsabagport, line 101:
```suggestion
            logger.error(
```

dvirsegev (approved):

PR #199, opened 2023-10-31
[Ocean] Add option to run integration once and also on docker

Tankilevitch:
# Description

- Added ONCE event listener to trigger the integration once
- Added `--once` flag to the sail command.
- Added the event listener type as a property to the ocean context, so integrations could decide on what to do on startup depends on the event listener type.
- Changed the configuration priority from `envs -> config file -> init`  to  `init -> envs -> config file`

What -
Why -
How -

## Type of change


CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+## 0.4.0 (2023-10-31)
+
+### Features
+
+- Added support for running ocean integrations once and new ocean sail options to support it. As part of it we added ImmediateEventListener.
```
yairsimantov20, line 14: Dont forget to fix the changelog

port_ocean/cli/commands/sail.py
```diff
@@ -52,4 +63,9 @@ def sail(
     print_logo()
 
     console.print("Setting sail... ⛵️⚓️⛵️⚓️ All hands on deck! ⚓️")
+
+    if once:
+        console.print("Setting event listener to IMMEDIATE")
+        os.environ["OCEAN__EVENT_LISTENER"] = '{"type": "IMMEDIATE"}'
```
yairsimantov20, line 69: Can u pass it to the run instead of setting environs?

port_ocean/cli/commands/sail.py
```diff
@@ -33,14 +34,24 @@
...
             If not specified, the default value is false.""",
 )
+@click.option(
+    "-O",
+    "--once",
+    "once",
+    type=bool,
+    is_flag=True,
+    help="""specify the option to run the integration once and exit.""",
+)
```
yairsimantov20, line 48: I think there is no need for that now
Just pass the correct event listener type

port_ocean/core/event_listener/immediate.py
```diff
@@ -0,0 +1,60 @@
...
+
+from port_ocean.core.event_listener.base import (
+    BaseEventListener,
+    EventListenerEvents,
+    EventListenerSettings,
+)
+from port_ocean.utils import repeat_every
+
+
+class ImmediateEventListenerSettings(EventListenerSettings):
```
yairsimantov20, line 14: Maybe the event listener thpe should be ONCE

port_ocean/core/event_listener/immediate.py
```diff
@@ -0,0 +1,60 @@
...
+    This class inherits from `EventListenerSettings`, which provides a foundation for creating event listener settings.
+    """
+
+    type: Literal["IMMEDIATE"]
+
+    def to_request(self) -> dict[str, Any]:
+        return {}
+
+
+class ImmediateEventListener(BaseEventListener):
```
yairsimantov20, line 26: Maybe we should not trigger the on start function
Because some integrations will create webhooks

yairsimantov20 (changes requested):
Few comments

port_ocean/core/event_listener/immediate.py
Tankilevitch: passed the event listener type to the start event context, so that each integration can decide on its behavior based on the event listener type

port_ocean/cli/commands/sail.py
Tankilevitch: it is problematic as currently the config loader knows how to handle envs and file, adding also parameters can make it less readable and is probably an over kill.

port_ocean/context/ocean.py
```diff
@@ -45,6 +45,10 @@ def integration_config(self) -> dict[str, Any]:
     def port_client(self) -> "PortClient":
         return self.app.port_client
 
+    @property
+    def event_listener_type(self) -> str:
```
yairsimantov20, line 49:
```suggestion
    def event_listener_type(self) -> Literal['' | '' | '']:
```

yairsimantov20 (approved):
🌊 GJ

Issue #224, opened 2023-11-09
[Bug] [Integration] [OpsGenie]


wscoble:
### Ocean Framework Version

Linux ocean-opsgenie-opsgenie-deployment-*** 5.15.117 #1 SMP Thu Jul 6 03:52:13 UTC 2023 x86_64 GNU/Linux
🌊 Ocean version: 0.4.2

### Integration Version

latest
sha256:798c81817034a3b3a8b36df247cd572c58fc014049e5fc5c2bb946f0c214f3bc

### Steps to reproduce

Run integration with 40+ OpsGenie Incidents, 429 responses will be abundant.

### What did you expect to see?

Maybe one 429 then a sleep and retry from the integration.

### What did you see instead?

20 or so 429 HTTP errors reported in the logs. The integration did not attempt to resync to remedy the issue.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #226, opened 2023-11-11
[Ocean] Add helper for integrations to handle failures with Third Party APIs & Fix 401 against Port API when token expired

Tankilevitch:
# Description

- Add retry decorator on 401 for requests against Port ( reason we don't use the HTTPX Retry Class in this case is because we need to re-run the `token` property that re-initializes the token )
- Add HTTPX retry client, both for port client as well as for integrations to use to handle different http errors, for better durability  - #224  uses implementation from - https://github.com/encode/httpx/issues/108#issuecomment-1434439481
- Fix `repeat_every` to calculate failed as repetition runs as well ( For Once Event Listener )

## Type of change

### DIFF
--- a/port_ocean/clients/port/authentication.py
+++ b/port_ocean/clients/port/authentication.py
@@ -81,7 +81,4 @@ class PortAuthentication:
                 self.client_id, self.client_secret
             )
 
-        if random.choices([1, 0]):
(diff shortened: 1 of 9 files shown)


MPTG94 (approved):
LGTM take a look at my comments

port_ocean/clients/port/authentication.py
```diff
@@ -72,6 +72,10 @@ async def headers(
     @property
     async def token(self) -> str:
         if not self._last_token_object or self._last_token_object.expired:
+            msg = "Token expired, fetching new token"
+            if not self._last_token_object:
+                msg = "No token found, fetching new token"
```
yairsimantov20, line 77:
```suggestion
           logger.info("Generating new Port api token")
```

yairsimantov20 (approved):
🌊

PR #236, opened 2023-11-21
Argocd blueprint mismatch

PeyGis:
# Description
The gitRepo blueprint property is normally an HTTPS URL. But some clients may have a [private repository](https://argo-cd.readthedocs.io/en/stable/user-guide/private-repositories/#ssh-private-key-credential) which uses a different URL format, resulting in errors when syncing the data to Port. This PR fixes the bug by updating all URL formatted blueprint properties to string

What -
Why -
How -

## Type of change


yairsimantov20 (approved):
🌊

PR #245, opened 2023-11-28
Port 5333 create httpx client for the integration

yairsimantov20:
# Description

What :
- Integrations handle the httpx async client on multithreaded scenario by themselves 
- Integrations  add the retry transport manually to the httpx client
- When there is a cyclic dependency between entities without createMissingRelation flag an unclear error is shown
- When sending a request to port the token might just get expired 

Why:
- When using the httpx async client and creating a new thread there is a need to create a new async client because the client is bounded to the current event loop
- We added a retry transport but every integration needs to add the transport to their client because each of them holds their own httpx async client
- We just throw the error that is thrown from the graphlib package
- The request itself might take more then the ttl left for the token

How:
- Created an httpx async client accessible from the ocean core to the integration wrapped with local proxy to handle recreation for multithreaded integrations
- To the same httpx async client we attched the retry transport to automatically deliver the retries to any integration using the ocean httpx client
- Added custom ocean exception for the cyclic error with a more clear message directing to use the right flags
- To the internal httpx client used by the port client we added ability for the transport to refresh the token when receiving status code 401 from port


port_ocean/utils.py
```diff
@@ -3,17 +3,40 @@
...
+                httpx.AsyncHTTPTransport(),
+                logger=logger,
+            )
+        )
+        _http_client.push(client)
+
+    return client
+
+
+http_async_client: httpx.AsyncClient = LocalProxy(lambda: _get_http_client_context())  # type: ignore
```
Tankilevitch, line 39: Lets add comment describing that the purpose of this http_async_client is for integration use

Tankilevitch (approved):
very nice, super clean 🌊 

minor description suggestion

PR #254, opened 2023-12-03
PORT-5378 normalizing identifier to lowercase

yairsimantov20:
# Description

What - Bug Fixes

- Fixed an issue causing the the integration to update when running the initialization of the integration when integration already exists and there are not blueprints (PORT-5378)
- Fixed an issue where setting integration type or identifier that contains a capital leter will not show the integration in the UI (PORT-5399)

Why -

- The initialization used the `initialize_integration` function that caused it to update integration that already exists instead of failing
- Port-api normalize the incoming user agent by changing it to lowercase what may cause the UI to search for the integration with the origin identifier while port knows it with the lowercase identifier

How -

- Switched the usage to `create_integration` in order to fail for existing integration
- Forcing the integration identifier & type to be lowercase by using pydantic validations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
-## 0.4.5 (2023-12-04)
+## 0.4.6 (2023-12-04)
 
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Bug Fixes
 
-- Fixed an issue causing the the integration to update when running the initialization of the integration when integration already exists and there are not blueprints (PORT-5378)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+## 0.4.5 (2023-12-04)
+
+
+### Bug Fixes
+
+- Fixed an issue causing the the integration to update when running the initialization of the integration when integration already exists and there are not blueprints (PORT-5378)
```
Tankilevitch, line 15: unclear

Tankilevitch: and also double `the the`

port_ocean/config/settings.py
```diff
@@ -33,6 +33,18 @@ class IntegrationSettings(BaseModel, extra=Extra.allow):
...
+    )
+    def validate_identifier(cls, v: str) -> str:
+        return v.lower()
+
+    @validator(
+        "type",
+    )
+    def validate_type(cls, v: str) -> str:
+        return v.lower()
+
```
Tankilevitch, line 47:
```suggestion
    @validator(
        "identifier", "type"
    )
    def validate_lower(cls, v: str) -> str:
        return v.lower()
```

Tankilevitch (approved):

PR #263, opened 2023-12-12
[Integration][Gitlab] Add support for gitlab system hooks

Tankilevitch:
# Description

What - Support handling gitlab system hooks 
Why - When user has a lot of groups and projects, it would be more efficient to just create one system hook for all the system  
How -  Support both handling system events and creating 

## Type of change

### DIFF
--- a/integrations/gitlab/CHANGELOG.md
+++ b/integrations/gitlab/CHANGELOG.md
@@ -11,7 +11,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Improvements
 
-- Added support for system hooks (#5220)
+- Added support for system hooks, this capability can be enabled using the useSystemHook flag. Enabling this capability will create system hooks instead of group webhooks (PORT-5220)
 
 
 0.1.33 (2023-12-05)
--- a/integrations/gitlab/CHANGELOG.md
+++ b/integrations/gitlab/CHANGELOG.md
@@ -7,12 +7,19 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
-0.1.33 (2023-12-12)
-===================
+0.1.34 (2023-12-12)
 
 ### Improvements
 
 - Added support for system hooks (#5220)
+
+
+0.1.33 (2023-12-05)
+===================
+
+### Improvements
+
+- Bumped ocean version to ^0.4.7 (#1)
 
 
 0.1.32 (2023-12-04)
--- a/integrations/gitlab/CHANGELOG.md
+++ b/integrations/gitlab/CHANGELOG.md
@@ -7,6 +7,14 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+0.1.32 (2023-12-04)
+===================
+
+### Improvements
+
+- Bumped ocean version to ^0.4.6 (#1)
+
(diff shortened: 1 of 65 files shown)


integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -186,6 +186,35 @@ def create_webhooks(self) -> list[int | str]:
...
+    def create_system_hook(self) -> None:
+        logger.debug("Checking if system hook already exists")
+        try:
+            for hook in self.gitlab_client.hooks.list(iterator=True):
+                if hook.url == f"{self.app_host}/integration/system/hook":
+                    logger.debug("System hook already exists, no need to create")
+                    return
+        except Exception:
+            logger.error(
+                "Failed to check if system hook exists, skipping trying to create, to avoid duplicates"
```
yairsimantov20, line 198: dont forget to log the exception
maybe you should use logger.exception

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -186,6 +186,35 @@ def create_webhooks(self) -> list[int | str]:
...
+                    "push_events": True,
+                    "merge_requests_events": True,
+                    "issues_events": True,
+                    "job_events": True,
+                    "tag_push_events": True,
+                }
+            )
+            logger.debug(f"Created system hook with id {resp.get_id()}")
+        except Exception:
+            logger.error("Failed to create system hook")
```
yairsimantov20, line 216: in case of exception please have the log print the exception message

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -21,7 +21,16 @@
...
+    await event_handler.notify(event_id, body)
+    return {"ok": True}
+
+
+@ocean.router.post("/system/hook")
+async def handle_system_webhook(request: Request) -> dict[str, Any]:
+    body = await request.json()
+    # some system hooks have event_type instead of event_name in the body, such as merge_request events
+    event_name = body.get("event_name") or body.get("event_type")
+    await system_event_handler.notify(event_name, body)
```
yairsimantov20, line 33: Because it is now possible not to setup those listeners you might have an error if a request will come to the system hook route when the system hook setup_listener was not called

When notifying the handler looks for the event in a dict (handlers[event]) and in python if you try to access a values that does not exists in the list there is an exception

Maybe we should use `.get` for the handlers or disable the hooks route

the same applied to the regular hooks

integrations/gitlab/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+0.1.34 (2023-12-12)
+
+### Improvements
+
+- Added support for system hooks (#5220)
```
yairsimantov20, line 14: Can you please be a bit more verbose?
When using a port issue please prefix it with PORT-

yairsimantov20 (changes requested):
some comments

integrations/gitlab/gitlab_integration/gitlab_service.py
Tankilevitch: logger.error, prints the error out of the box when inside a except :)

integrations/gitlab/gitlab_integration/gitlab_service.py
Tankilevitch: same

PR #265, opened 2023-12-14
terraform integration

mk-armah:


integrations/terraform/pyproject.toml
```diff
@@ -0,0 +1,102 @@
+[tool.poetry]
+name = "terraform"
+version = "0.1.0"
+description = "Terraform workspace and runs integration into port"
+authors = ["Michael Kofi Armah <michaelarmah@getport.io>"]
+
+[tool.poetry.dependencies]
+python = "^3.11"
+port_ocean = { version = "^0.4.5", extras = ["cli"] }
```
yairsimantov20, line 9: Upgrade ocean pls

integrations/terraform/.port/spec.yaml
```diff
@@ -0,0 +1,24 @@
+type: terraform
```
yairsimantov20, line 1: Pls change every where
This integration is to terraform cloud

yairsimantov20 (changes requested):
🌊

integrations/terraform-cloud/.port/spec.yaml
```diff
@@ -0,0 +1,24 @@
...
+description: terraform cloud integration for Port Ocean
+icon: Terraform # Should be one of the available icons in Port
+features:
+  - type: exporter
+    section: IaC # Should be one of the available sections in Port
+    resources:
+      - kind: workspace
+      - kind: run
+configurations:
+  - name: terraformHost
```
PeyGis, line 11: Add a `default` value of https://app.terraform.io for the terraformHost

mk-armah: added

yairsimantov20:
I left some comments and i have some questions 

- i dont see the outputs in your pr how does the integration support outputs?
- i think we should also have state version exported as its the run result that hold connection to the outputs. Also the workspace should hold its active stateVersion as a relation (also i think you can get the outputs from the state version routes)
- Client might have multiple organizations so i think we should hold information about the organization over the workspace entity
- Clients might have a large amount of workspaces and state version so i think we should allow some kind of api level filtering like the jira integration does

yairsimantov20 (commented):
Left some comments

integrations/terraform-cloud/main.py
```diff
@@ -0,0 +1,95 @@
...
+        logger.info(f"Received {len(workspace)} batch workspaces")
+        yield workspace
+
+
+
+@ocean.on_resync(ObjectKind.RUN)
+async def resync_runs(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    terraform_client = init_terraform_client()
+
+    async for workspaces in terraform_client.get_paginated_workspaces():
```
yairsimantov20, line 80: Make sure to cache information you might need on another handlers over the event attributes

yairsimantov20 (commented):
Caching

integrations/terraform-cloud/client.py
```diff
@@ -0,0 +1,144 @@
...
+                    existing_configs_response = await self.client.get(notification_config_url)
+                    existing_configs_response.raise_for_status()
+                    existing_configs = existing_configs_response.json().get('data', [])
+
+                    for config in existing_configs:
+                        if config['attributes']['url'] == webhook_target_url:
+                            logger.info(f"Webhook already exists for workspace {workspace['id']}")
+                            break
+
+                    else:
```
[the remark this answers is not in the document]
mk-armah: For Loop: Iterates through each existing webhook configuration (config) for a workspace.

Condition Inside the Loop: Checks if the current config's URL matches the target webhook URL (webhook_target_url).

Break Statement: If a matching webhook URL is found, the loop is exited prematurely using a break statement, and the program logs that the webhook already exists for this workspace.

Else Block: If no matching webhook URL is found (meaning the for loop completes without hitting a break), the else block is executed. This block contains code to create a new webhook configuration.

This is because the completion of the for loop without a break indicates that no existing configuration matches the target URL, signifying the need to create a new webhook.

integrations/terraform-cloud/.port/resources/blueprints.json
```diff
@@ -0,0 +1,128 @@
+[
+    {
+        "identifier": "terraformWorkspace",
```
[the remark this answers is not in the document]
mk-armah: e.g terraformCloudWorkspace ?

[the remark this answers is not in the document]
yairsimantov20: Correct

integrations/terraform-cloud/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+      query: "true"
+    port:
+      entity:
+        mappings:
+          identifier: .id
+          title: .attributes.name
+          blueprint: '"terraformWorkspace"'
+          properties:
+            workspaceName: .attributes.name
+            createdAt: .attributes."created-at"
```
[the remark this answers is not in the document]
mk-armah: The dashes between the strings causes issues if not accessed in .”property” format, this only applies to to double-word properties

PR #268, opened 2023-12-15
PORT-5453 Create centralized dockerfile

yairsimantov20:
# Description

What - Currently each integration holds its own dockerfile which means that update all docker files can be pretty hard
Why -  Each integration needs to build an image and dependencies might be different
How - Most of the dependencies are the same so i created a base for all integrations to be built upon. If integration needs custom docker file then it can specify its own dockerfile or init script

## Type of change


MatanHeledPort (approved):
LGTM

.github/workflows/release-integrations.yml
```diff
@@ -40,13 +40,19 @@ jobs:
               echo "Image already exists in $repository: port-ocean-$type:$version"
             else
               # Check if the 'version' variable contains any character other than digits and "."
+
+              dockerfile_path=integrations/_infra/Dockerfile
+              if test -e $folder/../Dockerfile; then
+                dockerfile_path=$folder/../Dockerfile
+              fi
+
```
Tankilevitch, line 48:
```suggestion

              dockerfile_path=integrations/_infra/Dockerfile
              if test -e $folder/../Dockerfile; then
                dockerfile_path=$folder/../Dockerfile
              fi
              # Check if the 'version' variable contains any character other than digits and "."

```

Tankilevitch (approved):
lgtm

PR #269, opened 2023-12-15
Servicenow Integration

PeyGis:
# Description

What - Added Servicenow integrations to Port
Why - 
How - Used REST API implementation to make calls to the Servicenow Table Api

## Type of change


integrations/servicenow/client.py
```diff
@@ -0,0 +1,78 @@
...
+                )
+                response.raise_for_status()
+                records = response.json().get("result", [])
+
+                if not records:
+                    break
+
+                yield records
+
+                total_records = int(response.headers.get("X-Total-Count", 0))
```
yairsimantov20, line 65: Can u get this value before the yield? I want more logs for how much data you queried out of how much there is in total

integrations/servicenow/client.py
```diff
@@ -0,0 +1,78 @@
...
+                if (offset + 1) * PAGE_SIZE >= total_records:
```
yairsimantov20, line 66: If you are using >= you might miss some objects

For example

If there are 200 items and you queried 100 then this if will check if 200 >= 200 and the pagination will stop

integrations/servicenow/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,52 @@
+resources:
+  - kind: group
+    selector:
+      query: 'true'
+      path: 'sys_user_group'
```
yairsimantov20, line 5: I dont like this 
Its 2 generic

Is their api really that generic 
The routes are not known?

And if it does then you can just use the kind instead

integrations/servicenow/main.py
```diff
@@ -0,0 +1,23 @@
...
+async def on_resources_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    logger.info(f"Listing Servicenow resource: {kind}")
+    servicenow_client = ServicenowClient(
+        ocean.integration_config["servicenow_instance"],
+        ocean.integration_config["servicenow_username"],
+        ocean.integration_config["servicenow_password"],
+    )
+
+    async for records in servicenow_client.get_paginated_resource():
+        logger.warning(f"Received records batch with {len(records)} {kind}")
```
yairsimantov20, line 17: Why is it a warning?

integrations/servicenow/main.py
```diff
@@ -0,0 +1,23 @@
...
+        yield records
+
+
+@ocean.on_start()
+async def on_start() -> None:
+    print("Starting Servicenow integration")
```
yairsimantov20, line 23: Nothing to setup?

yairsimantov20 (commented):
Left some comments

integrations/servicenow/client.py
```diff
@@ -0,0 +1,103 @@
...
+    async def get_paginated_resource(
+        self, resource_kind: ObjectKind
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        offset = 0
+
+        while True:
+            params: dict[str, Any] = {
+                "sysparm_offset": offset * PAGE_SIZE,
+                "sysparm_limit": PAGE_SIZE,
+                "sysparm_query": "ORDERBYsys_created_on",
```
Tankilevitch, line 50: `"ORDERBYsys_created_on"` typo? 
also why we need it?

integrations/servicenow/client.py
```diff
@@ -0,0 +1,103 @@
...
+                    url=f"{self.servicenow_host}/api/now/table/{resource_kind}",
+                    params=params,
+                )
+                response.raise_for_status()
+                records = response.json().get("result", [])
+
+                if not records:
+                    break
+
+                total_rows = int(response.headers.get("X-Total-Count", 0))
```
Tankilevitch, line 63: You override it in each query? can it change while you paginate?

integrations/servicenow/client.py
```diff
@@ -0,0 +1,103 @@
...
+                queried_rows = (offset * PAGE_SIZE) + len(records)
+                if queried_rows > total_rows:
```
Tankilevitch, line 65: how is it possible that you queried more than the total rows available? when this case matches?

integrations/servicenow/client.py
```diff
@@ -0,0 +1,103 @@
...
+                raise
+            except httpx.HTTPError as e:
+                logger.error(f"HTTP occurred while fetching Servicenow data: {e}")
+                raise
+
+    async def sanity_check(self) -> None:
+        try:
+            response = await self.http_client.get(
+                f"{self.servicenow_host}/api/now/table/cmdb_ci_app_server?sysparm_limit=1"
+            )
```
Tankilevitch, line 88: add comment what is this table

integrations/servicenow/client.py
```diff
@@ -0,0 +1,103 @@
...
+        except httpx.HTTPStatusError as e:
+            logger.error(
+                f"Servicenow failed connectivity check to the Servicenow instance because of HTTP error: {e.response.status_code} and response text: {e.response.text}"
+            )
+            raise
+        except httpx.HTTPError as e:
+            logger.error(
+                f"Servicenow failed connectivity check to the Servicenow instance because of HTTP error: {e}"
+            )
```
Tankilevitch, line 102: you can use the `logger.exception` which adds the exception information out of the box

https://docs.python.org/3/library/logging.html#logging.exception

integrations/servicenow/pyproject.toml
```diff
@@ -0,0 +1,102 @@
+[tool.poetry]
+name = "Servicenow"
```
Tankilevitch, line 2:
```suggestion
name = "servicenow"
```

Tankilevitch (changes requested):
Nice Job, added some comments :) 🌊

integrations/servicenow/client.py
PeyGis: This is the total number of records in the kind so it is static

integrations/servicenow/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+            createdOn: '.sys_created_on | (strptime("%Y-%m-%d %H:%M:%S") | strftime("%Y-%m-%dT%H:%M:%SZ"))'
+            createdBy: .sys_created_by
+  - kind: incident
+    selector:
+      query: 'true'
+    port:
+      entity:
+        mappings:
+          identifier: .number
+          title: .short_description
```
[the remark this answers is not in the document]
PeyGis: This is a string. I can also safely covert it to a string

integrations/servicenow/.port/resources/blueprints.json
```diff
@@ -0,0 +1,123 @@
...
+          "properties": {
+            "description": {
+              "title": "Description",
+              "type": "string"
+            },
+            "isActive": {
+              "title": "Is active",
+              "type": "boolean",
+              "default": true
+            },
```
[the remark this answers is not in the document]
PeyGis: Gotcha

integrations/servicenow/.port/spec.yaml
```diff
@@ -0,0 +1,22 @@
...
+features:
+  - type: exporter
+    section: Incident Management # Should be one of the available sections in Port
+    resources:
+      - kind: sys_user_group
+      - kind: sc_catalog
+      - kind: incident
+configurations:
+  - name: servicenowInstance
+    required: true
```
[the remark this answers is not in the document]
PeyGis: No there isn't. Each Servicenow app comes with its unique instance

integrations/servicenow/client.py
PeyGis: The results returned by Servicenow is not ordered. And since we are doing pagination, we will need to order the results so the offset and limit can be applied correctly

Tankilevitch: sounds good, my point was that it looks like the `ORDERBY` got joined with `sys_created_on` without any spacing. is that their query language?

integrations/servicenow/client.py
Tankilevitch: I am afraid from raise condition, where the total count change while we query and therefor can cause infinite loop. Most of the times pagination api's return some link or id that you use to query the next page so that the API understands that you keep paginating as part of the first request context. 

Let's make sure that we don't miss something.

Tankilevitch: Looking in their [docs](https://developer.servicenow.com/dev.do#!/reference/api/sandiego/rest/c_TableAPI%23table-GET?navFilter=table).

Seems like as part of the headers response you also get a link header in which it points you to the next request you need to perform.  I think we need to use this native approach, that way you just keep looping until there is no next link to go over or something like that.

Screen Shot 2023-12-20 at 14 09 10

integrations/servicenow/client.py
```diff
@@ -0,0 +1,97 @@
...
+                f"Servicenow instance name: {response.json().get('result', [])[0].get('instance_name')}"
+            )
+        except httpx.HTTPStatusError as e:
+            logger.error(
+                f"Servicenow failed connectivity check to the Servicenow instance because of HTTP error: {e.response.status_code} and response text: {e.response.text}"
+            )
+            raise
+        except httpx.HTTPError as e:
+            logger.exception(
+                f"Servicenow failed connectivity check to the Servicenow instance because of HTTP error: {e}"
```
Tankilevitch, line 95:
```suggestion
                "Integration failed to connect to Servicenow instance as part of sanity check due to HTTP error"
```

Note that the `logger.exception` will already add the error and traceback, so no need to add it to the string

integrations/servicenow/client.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+            response.raise_for_status()
+            logger.info("Servicenow sanity check passed")
+            logger.info(
+                f"Servicenow instance name: {response.json().get('result', [])[0].get('instance_name')}"
+            )
+        except httpx.HTTPStatusError as e:
+            logger.error(
+                f"Servicenow failed connectivity check to the Servicenow instance because of HTTP error: {e.response.status_code} and response text: {e.response.text}"
+            )
```
Tankilevitch, line 91:
```suggestion
            logger.error(
                f"Integration failed to connect to Servicenow instance as part of sanity check due to HTTP error: {e.response.status_code} and response text: {e.response.text}"
            )
```

integrations/servicenow/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- towncrier release notes start -->
+
+# Port_Ocean 0.4.8 (2023-12-15)
```
Tankilevitch, line 10:
```suggestion
# 0.1.0 (2023-12-15)
```

Tankilevitch (changes requested):

integrations/servicenow/client.py
PeyGis: The sysparm_query parameters for `ORDER_BYsys_created_on` seems correct. As seen [here](https://www.servicenow.com/community/developer-forum/how-to-sort-sys-updated-on-sort-by-datetime/m-p/2174358), and [here](https://www.servicenow.com/community/developer-forum/using-rest-api-with-servicenow/m-p/1648726), there is no space between the ORDERBY clause and the field

integrations/servicenow/.port/resources/blueprints.json
```diff
@@ -0,0 +1,120 @@
+[
+    {
+        "identifier": "servicenowGroup",
+        "title": "Servicenow Group",
+        "icon": "Service",
+        "schema": {
+          "properties": {
+            "description": {
+              "title": "Description",
```
Tankilevitch, line 9: indentation of these blueprint seems a bit off compared to others

integrations/servicenow/.port/resources/blueprints.json
```diff
@@ -0,0 +1,120 @@
...
+          "type": "string"
+        }
+      },
+      "required": []
+    },
+    "mirrorProperties": {},
+    "calculationProperties": {},
+    "aggregationProperties": {},
+    "relations": {}
+  },
```
Tankilevitch, line 63: no relation?

integrations/servicenow/.port/resources/blueprints.json
```diff
@@ -0,0 +1,120 @@
...
+            "type": "string"
+        }
+        },
+        "required": []
+    },
+    "mirrorProperties": {},
+    "calculationProperties": {},
+    "aggregationProperties": {},
+    "relations": {}
+    }
```
Tankilevitch, line 119: same

Tankilevitch (changes requested):
Nicely done with the pagination !! much cleaner 🚀 

Why we don't have any relation? is that on purpose? what is the usecase?

Tankilevitch (approved):

PR #273, opened 2023-12-18
[Integration][Azure] Add "Best Effort" delete resources on delete event 

Tankilevitch:
# Description

What - Added azure delete resources capability when receiving delete event
Why - Until now we went with the thought that we can't know what is the identifier that the user chose for its resources ( e.g if the user change the way the default identifier in the template is configured ) so we chose not to try to delete the resource
How - In this feature I've added best effort to try to delete the resource with the default identifier

## Type of change


yairsimantov20 (approved):
🌊

PR #277, opened 2023-12-19
Add Pagerduty Incident Analytics

PeyGis:
# Description

What -  Improved on the incident object by adding analytics data 
Why - Customers needed to see their pagerduty incident metrics
How - For each incident, we query the [PagerDuty Analytics API](https://developer.pagerduty.com/api-reference/3b465fd63d888-analytics-incident-metrics) and fetch data

## Type of change


integrations/pagerduty/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- towncrier release notes start -->
+# Port_Ocean 0.1.26 (2023-12-19)
+
+### Improvements
+
+- Improved on the PagerDuty incident object by adding analytics data (#26)
+
```
yairsimantov20, line 14: thats a feature

integrations/pagerduty/main.py
```diff
@@ -31,7 +47,13 @@ async def on_incidents_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
         data_key=ObjectKind.INCIDENTS,
         params=query_params.generate_request_params() if query_params else None,
     ):
-        yield incidents
+        logger.info(f"Received batch with {len(incidents)} incidents")
+        tasks = [
+            enrich_incident_with_analytics_data(pager_duty_client, semaphore, incident)
+            for incident in incidents
+        ]
```
yairsimantov20, line 54: because you are using semaphore it might take alot of time

can you somehow start yielding the incidents?

integrations/pagerduty/main.py
```diff
@@ -77,7 +99,11 @@ async def upsert_incident_webhook_handler(data: dict[str, Any]) -> None:
         response = await pager_duty_client.get_singular_from_pager_duty(
             object_type=ObjectKind.INCIDENTS, identifier=incident_id
         )
-        await ocean.register_raw(ObjectKind.INCIDENTS, [response["incident"]])
+        analytics_data = await pager_duty_client.get_incident_analytics(
+            incident_id=incident_id
+        )
+        incident_data = {**response["incident"], "__analytics": analytics_data}
```
yairsimantov20: reuse the enrichment function you created

yairsimantov20 (changes requested):
left some comments

integrations/pagerduty/main.py
PeyGis: Can you explain more on this?

integrations/pagerduty/clients/pagerduty.py
```diff
@@ -191,3 +191,27 @@ async def update_oncall_users(
...
+            response = await self.http_client.get(url)
+            response.raise_for_status()
+            data = response.json()
+            return data
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == 404:
+                logger.error(
+                    f"Incident {incident_id} analytics data was not found, skipping..."
+                )
+                return {}
```
Tankilevitch, line 209: if we skip, is this actual an error log? we want to log an error if there is an actual problem so it won't be too noisy and only important error that needs to be checked will be logged as error

integrations/pagerduty/clients/pagerduty.py
```diff
@@ -191,3 +191,27 @@ async def update_oncall_users(
...
+
+            logger.error(
+                f"HTTP error with status code: {e.response.status_code} and response text: {e.response.text}"
+            )
+            raise
+        except httpx.HTTPError as e:
+            logger.error(f"HTTP occurred while fetching incident analytics data: {e}")
+            raise
```
Tankilevitch, line 217: are we sure that we won't to explode if we were not able to get analytics? it seems like an optional property 🤔

Tankilevitch (changes requested):

integrations/pagerduty/main.py
yairsimantov20: you might have 100 incidents
you are blocking the semaphore to 5 analytics requests
which means all the 100 will wait for all the analytics

can you batch request the analytics and yield a batch every time?

yairsimantov20 (approved):
🌊 approved

Tankilevitch (approved):
good job

integrations/pagerduty/integration.py
```diff
@@ -88,6 +88,11 @@ class PagerdutySelector(Selector):
         api_query_params: PagerdutyIncidentAPIQueryParams | None = Field(
             alias="apiQueryParams"
         )
+        enable_incident_analytics: bool = Field(
+            default=False,
+            description="If set to true, will ingest incident analytics data to Port. Default value is false",
+            alias="enableIncidentAnalytics",
+        )
```
Tankilevitch, line 95: Its under selector so I feel like there is no need to add the "enable" verb as we are saying "what to select" under the selector 🤔 
Let me know if that makes sense for you
```suggestion
        incident_analytics: bool = Field(
            default=False,
            description="If set to true, will ingest incident analytics data to Port. Default value is false",
            alias="IncidentAnalytics",
        )
```

Tankilevitch (approved):

PR #279, opened 2023-12-20
Fix delete entities to be throttled 

Tankilevitch:
# Description

What - Change the way we delete entities from port to be done in batch 
Why - Until now we were sending all delete request at once which in case of thousands of entities to delete was causing a raise condition where entities were timing out due to pool timeout in the client as we limit the amount of max connections to 200 
How - changed to batch

## Type of change


yairsimantov20 (approved):
🌊

MPTG94 (commented):
See my comments

PR #280, opened 2023-12-20
Move Opsgenie calculations to blueprint level

PeyGis:
# Description
The Opsgenie service blueprint has an entity property for calculating the number of open incidents. To make this possible, one has to make extra API calls to fetch incidents related to the service. One has to now use jq to calculate the number of open incidents in the port-app-config.yaml.

This PR utilizes the `aggregationProperties` capabilities in Port by moving the number of open incidents calculations from the jq mapper to benefit from this feature.

What -
Why -
How -

## Type of change


yairsimantov20 (approved):
🌊

PR #282, opened 2023-12-20
PORT-5711-replace-poetry-lock-check-with-the-new-poetry-check-lock

yairsimantov20:
# Description

What - Swapped `poetry lock --check` with `poetry check --lock`
Why - `poetry lock --check` is deprecated

## Type of change


MPTG94 (approved):
LGTM

PR #283, opened 2023-12-20
Port 5712 ocean validate integration metadata files

yairsimantov20:
# Description

What -
Why -
How -

## Type of change

### DIFF
--- a/.github/workflows/validate-integration-files.yml
+++ b/.github/workflows/validate-integration-files.yml
@@ -13,22 +13,19 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
-    - name: Get changed directories
+    - name: Check Ocean version 🌊
       run: |
-        changed_dirs=$(git diff --name-only ${{ github.base_ref }}...${{ github.sha }} | grep '/' | cut -d'/' -f 1 | sort -u)
-        echo "Changed directories: $changed_dirs"
-
-    - name: Extract ocean package and version
-      run: |
+        changed_dirs=$(git diff --name-only ${{ github.base_ref }}...${{ github.sha }} | grep 'integrations/' | cut -d'/' -f 1,2 | sort -u)
+        package_version=$(pip show port-ocean | awk '/Version:/ {print $2}')
         for dir in $changed_dirs; do
-          pyproject_file=$(find $dir -name 'pyproject.toml')
+          pyproject_file=$(find $dir -name 'pyproject.toml' -not -path "**/.venv/*")
           if [ -n "$pyproject_file" ]; then
-            ocean_package=$(awk -F= '/^\[tool\.poetry\.dependencies\]/{flag=1;next}/^\[.*\]/{flag=0}flag' $pyproject_file | grep '^ocean' | cut -d'=' -f 2 | tr -d ' ' | tr -d '"')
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-integration-files.yml
```diff
@@ -0,0 +1,71 @@
...
+    paths:
+      - 'integrations/**'
+
+jobs:
+  validate-files:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
```
MPTG94, line 14:
```suggestion
      uses: actions/checkout@v4
```

integrations/argocd/pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = ["Isaac Coffie <isaac@getport.io>"]
 
 [tool.poetry.dependencies]
 python = "^3.11"
-port_ocean = {version = "^0.4.8", extras = ["cli"]}
+port_ocean = {version = "^0.4.9", extras = ["cli"]}
```
MPTG94, line 9: Remove when done

integrations/argocd/README.md
```diff
@@ -5,3 +5,4 @@ An integration used to import ArgoCD resources into Port.
 #### Install & use the integration - [Integration documentation](https://docs.getport.io/build-your-software-catalog/sync-data-to-catalog/kubernetes/argocd)
 
 #### Develop & improve the integration - [Ocean integration development documentation](https://ocean.getport.io/develop-an-integration/)
+test
```
MPTG94, line 8: Remove

.github/workflows/validate-integration-files.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: |
+        pip install toml-cli yq
+
+    - name: Check Ocean version 🌊
+      run: |
+        changed_dirs=$(git diff --name-only origin/main origin/${{ github.head_ref }} | grep 'integrations/' | cut -d'/' -f 1,2 | sort -u)
+        package_version=$(curl -s https://pypi.org/pypi/port-ocean/json | jq -r '.info.version')
+        for dir in $changed_dirs; do
+          pyproject_file=$(find $dir -name 'pyproject.toml' -not -path "**/.venv/*")
+          if [ -n "$pyproject_file" ]; then
```
MPTG94, line 28: Probably better to check the exit code of the find command (it will return 0 if something was found, and 1 if there were no results)

.github/workflows/validate-integration-files.yml
```diff
@@ -0,0 +1,71 @@
...
+            installed_version=$(toml get  tool.poetry.dependencies.port_ocean.version --toml-path $pyproject_file)
+            is_version_updated=$(python -c "from packaging import version;print(version.parse('$installed_version'.lstrip('^')) >= version.parse('$package_version'))")
```
MPTG94, line 30: This might be a bit risky because we are relying on the pre-installed python in the github runner, while I don't expect it to change into an incompatible version very soon, it still might happen, is there a way we can achieve this is bash?

If not, we can also just use the `setup-python` action, it will add a few more seconds to the CI but it will be more reproducible (we also don't need any packages so it shouldn't be too bad)

MPTG94 (approved):
Nice work, look at my comments

.github/workflows/validate-integration-files.yml
yairsimantov20: i already save the result to a var so i dont think it really matters

PR #285, opened 2023-12-21
Ocean async client

yairsimantov20:
# Description

What - server proxy environment variables wont work with the current client
Why - There are some features that are ignored when passing a custom transport to the httpx.async_client
How - Wrapped the httpx async client with implementation that overrides the default transport class with custom tranport in order to apply all defaults httpx features that are ignored when passing a custom tranport instance

## Type of change


Tankilevitch (approved):
🌊

PR #287, opened 2023-12-21
Support creation of blueprints with aggregation properties & Change entity from optional to required in port app config

Tankilevitch:
# Description

What - 
- Added handling for aggregation properties when initializing the integration, so it will patch the aggregation properties after creating the relations (PORT-5717)
- Changed entity property in the `portResourceConfig` to be required instead of optional, as we don't support creation of blueprints as part of the app config (PORT-4549) 
- Fixed port merge cr comments in #279  

Why -
- It was failing to create blueprints with aggregation properties due to the order in which we create the blueprints
- There is no need for it any more, and it was causing integration to use `# type ignore` as it was relying on the entity to exist in the port app config

How -
- added the aggregation properties to be patched after relations are being created
- removed optional

## Type of change


yairsimantov20 (approved):
🌊 LGTM

PR #308, opened 2023-12-30
[Core] Add support to create pages as part of integration setup

Tankilevitch:
# Description

What - 
- Add support to create pages
- Add blueprints and integration validation existence before defaults creation
- Add more verbosity on the delete entities resync process

Why -
- To be able to create pages
- To reduce the noisiness  when restarting the integration in the audit log
- To help with understanding the progress in entities deletion

## Type of change

### DIFF
--- a/port_ocean/clients/port/mixins/blueprints.py
+++ b/port_ocean/clients/port/mixins/blueprints.py
@@ -105,3 +105,16 @@ class BlueprintClientMixin:
         )
 
         handle_status_code(response)
+
+    async def create_page(
+        self,
+        page: dict[str, Any],
+    ) -> None:
+        logger.info(f"Creating page: {page}")
+        response = await self.client.post(
+            f"{self.auth.api_url}/pages",
+            json=page,
+            headers=await self.auth.headers(),
+        )
+
+        handle_status_code(response)
--- a/port_ocean/clients/port/mixins/blueprints.py
+++ b/port_ocean/clients/port/mixins/blueprints.py
@@ -14,13 +14,15 @@ class BlueprintClientMixin:
         self.auth = auth
         self.client = client
 
-    async def get_blueprint(self, identifier: str) -> Blueprint:
+    async def get_blueprint(
+        self, identifier: str, should_log: bool = True
+    ) -> Blueprint:
         logger.info(f"Fetching blueprint with id: {identifier}")
         response = await self.client.get(
             f"{self.auth.api_url}/blueprints/{identifier}",
             headers=await self.auth.headers(),
         )
-        handle_status_code(response)
+        handle_status_code(response, should_log=should_log)
         return Blueprint.parse_obj(response.json()["blueprint"])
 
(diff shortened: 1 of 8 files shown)


port_ocean/clients/port/mixins/blueprints.py
```diff
@@ -14,13 +14,15 @@ def __init__(self, auth: PortAuthentication, client: httpx.AsyncClient):
...
+    async def get_blueprint(
+        self, identifier: str, should_log: bool = True
+    ) -> Blueprint:
         logger.info(f"Fetching blueprint with id: {identifier}")
         response = await self.client.get(
             f"{self.auth.api_url}/blueprints/{identifier}",
             headers=await self.auth.headers(),
         )
-        handle_status_code(response)
+        handle_status_code(response, should_log=should_log)
```
MPTG94, line 25: Why is `should log` added only here? should it not be in other places? is it necessary for it to be on get_blueprint specifically?

MPTG94 (commented):
Take a look at my comments

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -368,6 +369,9 @@ async def sync_raw_all(
 
                     logger.error(message, exc_info=error_group)
                 else:
+                    logger.info(
+                        f"Starting resync diff calculation, entities at port before resync: {len(entities_at_port)}, entities created: {len(flat_created_entities)}"
```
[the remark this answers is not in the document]
Tankilevitch: its after the integration finished upserting, note the code above where it loops over each mapping and performs a sync

port_ocean/clients/port/mixins/blueprints.py
Tankilevitch: Note line https://github.com/port-labs/ocean/pull/308/files/ffaf3b7ef7fa079d5ff99707d45275626d18e1e9#diff-805c13724e2962cfe2a15019fe6cbfb8dcd894d15cda5d0edb414dbbf84fd64cR70 which eventually goes over all blueprints which we need to create and checks if its exists, we want the resource to not exists which in this case will raise an exception, but the exception is actually a good thing so I don't want to log an error in this case.

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -368,6 +369,9 @@ async def sync_raw_all(
 
                     logger.error(message, exc_info=error_group)
                 else:
+                    logger.info(
+                        f"Starting resync diff calculation, number of entities at Port before resync: {len(entities_at_port)}, number of entities created: {len(flat_created_entities)}"
```
MPTG94, line 373:
```suggestion
                        f"Running resync diff calculation, number of entities at Port before resync: {len(entities_at_port)}, number of entities created during sync: {len(flat_created_entities)}"
```

MPTG94 (approved):
LGTM

PR #323, opened 2024-01-07
Port 5938 ocean integration exclude unset mapping

yairsimantov20:
# Description

What - Integrations with no team mapping fails to create entitites
Why - the core use none for teams with no mappings
How - using pydantic excluded unset fields from the mapping

## Type of change


Tankilevitch (approved):
lgtm

PR #326, opened 2024-01-08
chore: fix workflow dependency on release integrations and always syn…

danielsinai:
# Description

What - Alwasys syncing S3
Why - Sometimes, you might want to update the spec.yaml but not to actually require a new integration release
How - Change the depends_on key in the gh workflow

## Type of change


danielsinai:
After a bit of thinking, it's better to keep it this way in order to validate integrations are synced before deploying changes to the registry it self

PR #356, opened 2024-01-19
[Integration][Sonarqube] Add analysis data for onpremise

PeyGis:
# Description

What - The integration currently does not sync `analysis` kind data to Port when the SonarUrl is on-premise
Why - The Analysis kind is only available for Sonarcloud (the SaaS)
How - This PR adds support for ingesting analysis data to Port using the PR and Measures API

Also improved on the way we upsert entities in Port using the bulk upsert rather than single upsert

## Type of change


integrations/sonarqube/main.py
```diff
@@ -36,26 +36,19 @@ async def on_project_resync(kind: str) -> list[dict[str, Any]]:
...
     logger.info(f"Listing Sonarqube resource: {kind}")
-    if is_onpremise_deployment():
-        logger.debug(
-            "Skipping resync because the integration does not support on-premise Sonarqube deployment"
-        )
-        return
 
     sonar_client = init_sonar_client()
-
     async for analyses_list in sonar_client.get_all_analyses():
```
Tankilevitch, line 50: is there a specific error we are expecting when running against on-premise?

Tankilevitch (changes requested):
Looks good, minor question

integrations/sonarqube/.port/resources/port-app-config.yaml
```diff
@@ -30,15 +30,15 @@ resources:
...
+            branch: .branch // .__branchName
             fixedIssues: .measures.violations_fixed
-            newIssues: .measures.violations_added
-            coverage: .measures.coverage_change
-            duplications: .measures.duplicated_lines_density_change
-            createdAt: .__analysisDate
+            newIssues: .measures.violations_added // .__measures[]? | select(.metric == "new_violations") | .period.value
+            coverage: .measures.coverage_change // .__measures[]? | select(.metric == "new_coverage") | .period.value
+            duplications: .measures.duplicated_lines_density_change // .__measures[]? | select(.metric == "new_duplicated_lines_density") | .period.value
+            createdAt: .analysisDate // .__analysisDate
```
Tankilevitch, line 41: I am unable to follow the convention you used, for some properties you added some mapping to be first and some as last.

1. Is that on purpose, if so we need to document it somehow.
2. seems that most of the properties have different mapping resources, so maybe we should just define a new "mapping" resource and we will use the relevant one per whether its onpremise or not. 
@yairsimantov20  would like to get your input on it.

PeyGis: Yes, the reason why we're having different resource mapping is because of the enhancements: One API route is unable to contain all the needed information so I have to call other APIs to get the full data required.

I like the idea of having another mapping for the same `analysis` kind but filter when it's on-premise. Challenge will be the best way to determine whether it is on-premise or not

Tankilevitch: sure understood regarding the different APIs to get the same data, what I referred to is the order, which makes it harder after committed to understand whether which "mapping" is on-premise and which is saas.

For example, in branch key, you added the mapping as the first executed while in newIssues you added as the second option.

I think the way to understand whether its on-premise is using the `sonarIsOnPremise` parameter that you just added to the config. 
And have it as two different kinds instead of the same one, that way when you in on-premise you ignore the saas one and the same for the opposite

integrations/sonarqube/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "sonarqube"
-version = "0.1.27"
+version = "0.1.28"
```
Tankilevitch, line 3:
```suggestion
version = "0.1.32"
```

Tankilevitch (approved):

PR #372, opened 2024-01-29
PORT-5605 | Dynatrace Ocean Integration

lordsarcastic:
# Description

What - Implemented Ocean integration for Dynatrace
Why - Allows users to bring in Problems, SLOs, Entities and Problems into Port
How - Added an integration following the integration template

## Type of change


integrations/dynatrace/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- towncrier release notes start -->
```
PeyGis, line 8: The changelog content for this feature is missing. You can use this steps to generate a changelog

towncrier create --content "Implemented Dynatrace integration to bring slos, entities and problems data to Port" 0.1.0.feature.md

commit the change log files
towncrier build --yes --version 0.1.16


repeat for each new feature/PR

integrations/dynatrace/changelog/1.0.0.feature.md
```diff
@@ -0,0 +1 @@
+Add Dynatrace integration
```
PeyGis, line 1: This should be deleted after you build the changelog

integrations/dynatrace/config.yaml
```diff
@@ -0,0 +1,17 @@
...
+  clientId: "{{ from env PORT_CLIENT_ID }}" # Can be loaded via environment variable: PORT_CLIENT_ID
+  clientSecret: "{{ from env PORT_CLIENT_SECRET }}" # Can be loaded via environment variable: PORT_CLIENT_SECRET
+# The event listener to use for the integration service.
+eventListener:
+  type: KAFKA
+integration:
+  # The identifier of this integration instance.
+  identifier: "{{ from env INTEGRATION_IDENTIFIER }}"
+  # The type of the integration.
+  type: "My Integration type (Gitlab, Jira, etc.)"
```
PeyGis, line 14: the value for type should be `dynatrace`... it should always be the same as the one in the pyproject.toml

[tool.poetry]
name = "dynatrace"

integrations/dynatrace/config.yaml
```diff
@@ -0,0 +1,17 @@
...
+  config:
+    myGitToken: "{{ from env MY_GIT_TOKEN }}"
```
PeyGis, line 16: The `config` object should match the ones you specified in the spec.yaml

you should change them to `dynatraceApiKey`, `dynatraceHostUrl`

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,58 @@
...
+from port_ocean.context.ocean import ocean
+
+from client import DynatraceClient
+from port_ocean.core.ocean_types import ASYNC_GENERATOR_RESYNC_TYPE
+
+
+class ObjectKind(StrEnum):
+    PROBLEM = "problem"
+    SLO = "slo"
+    ENTITY_TYPE = "entity_type"
```
PeyGis, line 14: Remove entity type enum since we don't need it

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,58 @@
...
+    ENTITY = "entity"
+
+
+def initialize_client() -> DynatraceClient:
+    return DynatraceClient(
+        ocean.integration_config["dynatrace_host"],
+        ocean.integration_config["dynatrace_api_key"],
```
PeyGis, line 21: In the spec.yaml and config.yaml, we have `dynatraceApiKey`, `dynatraceHostUrl`

the integration_config uses the same variables but in underscore. this means you need to change these to

`dynatrace_host_url` and `dynatrace_api_key` (which you got right here)

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+
+@ocean.on_resync(ObjectKind.PROBLEM)
+async def on_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
```
PeyGis, line 26: The async function should be have a descriptive name such as

async def on_problems_resync(kind: str) ->

Please rename the other functions as well where you have the @ocean.on_resync() decorator

PeyGis (changes requested):
Looks good so far. I have left some comments. Please go over it and update. I am re-reviewing and will add more comments

integrations/dynatrace/.port/resources/blueprints.json
```diff
@@ -0,0 +1,308 @@
+[
+    {
+        "identifier": "dynatraceProblem",
+        "description": "This blueprint creates a Dynatrace Problem",
```
PeyGis, line 4: The blueprint represents a Dynatrace problem

integrations/dynatrace/.port/resources/blueprints.json
```diff
@@ -0,0 +1,308 @@
...
+                    "title": "Root Cause Entity",
+                    "target": "dynatraceEntity",
+                    "required": false,
+                    "many": false
+                }
+            }
+        }
+    },
+    {
+        "identifier": "dynatraceEntityType",
```
PeyGis, line 101: Remove this blueprint

integrations/dynatrace/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,88 @@
...
+            problemFilters: .problemFilters[].name
+            severityLevel: .severityLevel
+            status: .status
+            startTime: '.startTime / 1000 | todate'
+            endTime: '.endTime | if . == -1 then null else (./1000 | todate) end'
+          relations:
+            impactedEntities: .impactedEntities[].name
+            linkedProblemInfo: .linkedProblemInfo.problemId
+            rootCauseEntity: .rootCauseEntity.name
+  - kind: entity_type
```
PeyGis, line 26: let's remove this kind

integrations/dynatrace/.port/spec.yaml
```diff
@@ -0,0 +1,19 @@
...
+description: Dynatrace integration for Port Ocean
+icon: Dynatrace
+features:
+  - type: exporter
+    section: Incident Management
+    resources:
+      - kind: problem
+      - kind: slo
+      - kind: entity
+      - kind: entity_type
```
PeyGis, line 11: let's remove this

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,104 @@
...
+import httpx
+from loguru import logger
+from port_ocean.context.event import event
+from port_ocean.utils import http_async_client
+
+
+class ResourceKey(StrEnum):
+    PROBLEM = "problem"
+    SLO = "slo"
+    ENTITY_TYPE = "entity_type"
```
PeyGis, line 13: let's remove this

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,104 @@
...
+    ENTITY = "entity"
+
+
+class DynatraceClient:
+    def __init__(self, host_url: str, api_key: str) -> None:
+        self.host_url = host_url
+        self.client = http_async_client
+        self.client.headers = httpx.Headers({"Authorization": f"Api-Token {api_key}"})
```
PeyGis, line 21: Our http_async_client has custom headers such as managing retry transport etc. So instead of overwriting it, you could instead update it using an example code copied from the Argocd integration

`        self.http_client.headers.update({"Authorization": f"Api-Token {api_key}"})`

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,58 @@
...
+
+@ocean.on_resync(ObjectKind.ENTITY_TYPE)
+async def on_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    dynatrace_client = initialize_client()
+
+    async for entities in dynatrace_client.get_resource(kind):
+        logger.info(f"Received batch with {len(entities)} entities")
+        yield entities
+
+
```
PeyGis, line 51: Let's all add the webhook route to bring real time data from Dynatrace

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,104 @@
...
+            next_page_key = response_data.get("nextPageKey")
+            if not next_page_key:
+                break
+            url = f'{self.host_url}/api/v2/?entitySelector=type("{type}")&nextPageKey={next_page_key}'
+
+    async def _get_entities(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        async for entity_types in self._get_entity_types():
+            for entity_type in entity_types:
+                async for entities in self._get_entities_from_type(entity_type["type"]):
+                    yield entities
```
PeyGis, line 85: Dynatrace has about 300 plus entity types. So instead of going through iteratively, I will suggest you run this in parallel (using asyncio) to speed up the process.

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        if cache := event.attributes.get(resource):
+            logger.info(f"picking {resource} from cache")
+            yield cache
+            return
+
+        async for resources in RESOURCE_MAPPING[resource]():
+            event.attributes.setdefault(resource, []).extend(resources)
+            yield resources
```
PeyGis, line 104: Instead of having this function, what I will suggest is to have a global async paginated function that can fetch and return paginated resources. You the call this function in the individual specific resources.

Have a look at how we've done it in the [terraform cloud integration](https://github.com/port-labs/ocean/blob/main/integrations/terraform-cloud/client.py).

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,104 @@
...
+class DynatraceClient:
+    def __init__(self, host_url: str, api_key: str) -> None:
+        self.host_url = host_url
+        self.client = http_async_client
+        self.client.headers = httpx.Headers({"Authorization": f"Api-Token {api_key}"})
+
+    async def _get_problems(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        url = f"{self.host_url}/api/v2/problems"
+        logger.info(f"Fetching problems from {url}")
+        while True:
```
PeyGis, line 26: This block is missing try catch and logger for potential errors. We're interested in httpx.HTTPStatusError and httpx.HTTPError, in addition to other errors that may be relevant to the code.

Please update it in other functions too. If you implement the global paginated resource function, you will do it once, and apply it to all resources.

        try:
            url = f"{self.api_url}/{endpoint}"
            logger.info(
                f"URL: {url}, Method: {method}, Params: {query_params}, Body: {json_data}"
            )
            response = await self.client.request(
                method=method,
                url=url,
                params=query_params,
                json=json_data,
            )
            response.raise_for_status()

            logger.info(f"Successfully retrieved data for endpoint: {endpoint}")

            return response.json()

            logger.error(
                f"HTTP error on {endpoint}: {e.response.status_code} - {e.response.text}"
            )
            raise
        except httpx.HTTPError as e:
            logger.error(f"HTTP error on {endpoint}: {str(e)}")
            raise

PeyGis (changes requested):
I've added more comments. This time, I focus on the client.py and main.py

integrations/dynatrace/CHANGELOG.md
```diff
@@ -0,0 +1,33 @@
...
+### Bug Fixes
+
+- Fixed bugs in queries (1.0.1)
+
+
+# Port_Ocean 1.0.0 (2024-01-29)
+
+### Features
+
+- Add Dynatrace integration (1.0.0)
```
PeyGis, line 33: Since the integration has not been published yet, all these changes is still in version 0.1.0. So create just 1 Features changelog

`Implemented Dynatrace integration to bring slos, entities and problems data to Port`

PeyGis (changes requested):

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        if cache := event.attributes.get(resource):
+            logger.info(f"picking {resource} from cache")
+            yield cache
+            return
+
+        async for resources in RESOURCE_MAPPING[resource]():
+            event.attributes.setdefault(resource, []).extend(resources)
+            yield resources
```
PeyGis, line 97: This function is over complicating the application. Let's remove it and instead call the individual methods in the main.py

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,97 @@
...
+                f"HTTP error on {url}: {e.response.status_code} - {e.response.text}"
+            )
+            raise
+        except httpx.HTTPError as e:
+            logger.error(f"HTTP error on {url}: {e}")
+            raise
+
+    async def _get_problems(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        async for problems in self._get_paginated_resources(
+            f"{self.host_url}/api/v2/problems", "problems"
```
PeyGis, line 49: Since all the APIs are referencing the v2, you can move this to the constructor

    def __init__(self, host_url: str, api_key: str) -> None:
        self.host_url = f"{host_url}/api/v2"
        self.client = http_async_client
        self.client.headers.update({"Authorization": f"Api-Token {api_key}"})

then in the individual methods, you can use it as

      async for slos in self._get_paginated_resources(
            f"{self.host_url}/slo", "slo"
        ):
            yield slos

integrations/dynatrace/client.py
lordsarcastic: I think that method simplifies things better compared to having individual methods. Just like we abstracted away the pagination into a different method, I'm abstracting away the caching instead of duplicating both blocks of code into all the methods.

PeyGis: The reason why this function may not be needed is because the application can still run without it. What I see it doing is creating a mapping between the resource keys and the functions that implement each kind. In the event we have say then, this mapping will grow and it wouldn't be sustainable.

Another reason I think we may not need it is on the issue of caching. At the moment, I don't see any kind resync or event that will be triggered twice to benefit from the cached resoures. the only resource that will benefit from caching is the _get_entity_type. But that's also used just once in the _get_entity method

PeyGis: What could work even better is finding a way to call the _get_paginated_resources(resource_kind) directly in the main.py. You're free to decide on this

lordsarcastic: So, no need to cache, except for retrieving entity types, yes?

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,61 @@
...
+@ocean.on_resync(ObjectKind.ENTITY)
+async def on_resync_entities(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    dynatrace_client = initialize_client()
+
+    async for entities in dynatrace_client.get_entities():
+        logger.info(f"Received batch with {len(entities)} entities")
+        yield entities
+
+
+@ocean.router.post("/events")
```
PeyGis, line 51: Let's rename the route to `webhook` instead of `events`

integrations/dynatrace/client.py
lordsarcastic: I'm not sure we can do that easily. We won't get much performance gains. First, we can't use asyncio while running within the context of Ocean.

If I have to do this, I'll have to yield each type from the `_get_entity_types` method. This involves looping through each result in each API call. We may not have that much gain.

What do you think?

integrations/dynatrace/.port/resources/blueprints.json
```diff
@@ -0,0 +1,264 @@
...
+        "type": {
+          "type": "string",
+          "title": "Type",
+          "description": "The type of the entity."
+        },
+        "tags": {
+          "type": "array",
+          "title": "Tags",
+          "description": "A list of tags of the entity."
+        },
```
Tankilevitch, line 128: make sure you define the array type in each property
e.g. 
```
          "items": {
            "type": "string"
          },
```

integrations/dynatrace/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,72 @@
...
+            errorBudgetMetricKey: .errorBudgetMetricKey
+            evaluatedPercentage: .evaluatedPercentage
+            evaluationType: .evaluationType
+            filter: .filter
+            metricExpression: .metricExpression
+            metricKey: .metricKey
+            normalizedErrorBudgetMetricKey: .normalizedErrorBudgetMetricKey
+            relatedOpenProblems: .relatedOpenProblems
+            relatedTotalProblems: .relatedTotalProblems
+            timeframe: .timeframe
```
Tankilevitch, line 72: you are exporting almost the whole rest api response - https://docs.dynatrace.com/docs/dynatrace-api/environment-api/service-level-objectives/get-slo#response-body-json-model

I think this needs to be reduced. not a blocker in any way, remember that this is only the template, the users will be able to extend it and if they will want they will add more properties.

lets try to reduce the number of properties that we return to only the most relevant ones. If you are not sure which one, we can discuss it.

integrations/dynatrace/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- towncrier release notes start -->
+
+# Port_Ocean 1.0.0 (2024-01-29)
```
Tankilevitch, line 10: Make sure you specify the same version here as in the pyproject.toml
```suggestion
# Port_Ocean 0.1.0 (2024-01-29)
```

integrations/dynatrace/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+
+### Features
+
+- Implemented Dynatrace integration to bring slos, entities and problems data to Port (1.0.0)
```
Tankilevitch, line 14: specify the jira ticket `PORT-x` or set to `#1`
```suggestion
- Implemented Dynatrace integration to bring slos, entities and problems data to Port (PORT-x)
```

integrations/dynatrace/.port/spec.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - kind: slo
+      - kind: entity
+configurations:
+  - name: dynatraceApiKey
+    required: true
+    type: string
+    sensitive: true
+  - name: dynatraceHostUrl
+    type: url
+    required: true
```
Tankilevitch, line 18: is there a default url that can be set?

integrations/dynatrace/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,72 @@
...
+            firstSeen: ".firstSeenTms / 1000 | todate"
+            lastSeen: ".lastSeenTms / 1000 | todate"
+            type: .type
+            tags: .tags[].stringRepresentation
+            managementZones: .managementZones[].name
+            properties: .properties
+            fromRelationships: .fromRelationships
+            toRelationships: .toRelationships
+  - kind: slo
+    selector:
```
Tankilevitch, line 46: same here as with the `blueprints.json` make sure you define the order of the kinds in a way where the first ones will be without relations, the order here is the same order in which the resync will run, meaning that the `problem` kind will be run first, and will define relations to entities that yet to get created ( It won't fail because we have default `createMissingRelatedEntities` as default which will create related entities if they are not exists)

integrations/dynatrace/.port/resources/blueprints.json
```diff
@@ -0,0 +1,264 @@
...
+          "type": "object",
+          "title": "From Relationships",
+          "description": "A list of all relationships where the entity is the source."
+        },
+        "toRelationships": {
+          "type": "object",
+          "title": "To Relationships",
+          "description": "A list of all relationships where the entity is the target."
+        }
+      }
```
Tankilevitch, line 149: shouldn't this be a relation?

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,61 @@
...
+async def on_problem_event(event: dict[str, str | Any]) -> dict[str, bool]:
+    dynatrace_client = initialize_client()
+    logger.info(f"Received problem event: {event}")
+
+    if problem_id := event.get("ProblemID"):
+        problem = await dynatrace_client.get_single_problem(problem_id)
+        await ocean.register_raw(ObjectKind.PROBLEM, [problem])
+
+    logger.info("Webhook event processed")
+    return {"ok": True}
```
Tankilevitch, line 61: you are missing `on_startup` webhook initialization. 
its ok if its not possible as it requires specific configuration. we will need to make sure we have clear definition on how to create the webhook.
Either way I would suggest making the url more type specific to `/webhook/event` as there might be more than event webhook, if this is not the case, feel free to correct me.

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,91 @@
...
+            "entities",
+            params={"entitySelector": f'type("{type_}")'},
+        ):
+            yield entities
+
+    async def get_entities(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        async for entity_types in self._get_entity_types():
+            for entity_type in entity_types:
+                async for entities in self._get_entities_from_type(entity_type["type"]):
+                    yield entities
```
Tankilevitch, line 76: I think it would be a useful comment to add a reference to the API here and state that it is required to pass types in the entitySelector.  as it wasn't straight forward why only for the entities we perform extra request to the types.

https://docs.dynatrace.com/docs/dynatrace-api/environment-api/entity-v2/get-entities-list#parameters

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,91 @@
...
+        self.client = http_async_client
+        self.client.headers.update({"Authorization": f"Api-Token {api_key}"})
+
+    async def _get_paginated_resources(
+        self, url: str, resource_key: str, params: dict[str, Any] = {}
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        logger.info(f"Fetching {resource_key} from {url} with params {params}")
+        try:
+            while True:
+                response = await self.client.get(url, params=params)
```
Tankilevitch, line 28: you can pass pageSize in some of the api, it will be useful to reduce the possibility to get rate-limit, as we increase the number of entities that we query in each request will require less requests to the API. 

from the docs 
```
pageSize | integer | The amount of entities.If not set, 50 is used.
-- | -- | --

```
https://docs.dynatrace.com/docs/dynatrace-api/environment-api/entity-v2/get-entities-list#parameters

Change to 100

integrations/dynatrace/client.py
Tankilevitch: Another thought, bringing all the monitored entities in a user environment can be so much data, maybe we should allow to pass `types` as part of the entity kind in the selector, where by default we will only export two types (Application / Service) and is open to user changes.

Tankilevitch (changes requested):
Nice Job @lordsarcastic 
Commented on the required changes and adjustments.

integrations/dynatrace/.port/resources/blueprints.json
```diff
@@ -0,0 +1,264 @@
...
+    "mirrorProperties": {},
+    "calculationProperties": {},
+    "relations": {}
+  },
+  {
+    "identifier": "dynatraceSlo",
+    "description": "This blueprint represents a Dynatrace SLO",
+    "title": "Dynatrace SLO",
+    "icon": "Dynatrace",
+    "schema": {
```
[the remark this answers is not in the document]
lordsarcastic: There isn't any relation between the two. It's just a standalone resource.

integrations/dynatrace/.port/resources/blueprints.json
lordsarcastic: Not exactly, both are ambiguous objects with little to not information given on the contents. Since the schema isn't standardized, we can't reliably pull a relationship out of it. I'll rather remove it.

integrations/dynatrace/client.py
lordsarcastic: What do you think about using the maximum for each of the API calls?
500: Problems
10,000: SLOs
100: Entities

lordsarcastic: That's the maximum we can use as page size. I used that.

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,118 @@
...
+
+    async def get_problems(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        async for problems in self._get_paginated_resources(
+            f"{self.host_url}/problems", "problems", {"pageSize": 500}
+        ):
+            yield problems
+
+    async def get_slos(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        async for slos in self._get_paginated_resources(
+            f"{self.host_url}/slo", "slo", {"pageSize": 10_000}
```
PeyGis, line 59: is 10_000 a typo or that's how Dynatrace expects the pageSize value to be passed?

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,118 @@
...
+        ):
+            yield slos
+
+    async def _get_entity_types(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        async for entity_types in self._get_paginated_resources(
+            f"{self.host_url}/entityTypes", "types", {"pageSize": 100}
+        ):
+            yield entity_types
```
PeyGis, line 67: Since we are using the selector for the entity types, I guess we don't need this then

integrations/dynatrace/client.py
```diff
@@ -0,0 +1,118 @@
...
+from port_ocean.context.event import event
+from port_ocean.utils import http_async_client
+
+from integration import DynatraceResourceConfig
+
+
+class ResourceKey(StrEnum):
+    PROBLEM = "problem"
+    SLO = "slo"
+    ENTITY_TYPE = "entity_type"
```
PeyGis, line 16: Since we're relying on the selector for the entity types, I guess we don't need this then

integrations/dynatrace/integration.py
```diff
@@ -0,0 +1,28 @@
...
+    Selector,
+)
+from port_ocean.core.integrations.base import BaseIntegration
+from pydantic.fields import Field
+
+
+class DynatraceEntitySelector(Selector):
+    entity_types: list[str] = Field(
+        default=["APPLICATION", "SERVICE"],
+        description="List of entity types to be fetched",
```
PeyGis, line 14: Add and `alias = "entityTypes"`. this is what will be passed in the port-app-config when the user is specifying their interested entity types

PeyGis (changes requested):
Some minor comments

integrations/dynatrace/client.py
lordsarcastic: Neither. I wanted the 10k to be easier to read. Python will treat it as 10k

PeyGis: Okay. What happens when you do just 10000?

integrations/dynatrace/.port/spec.yaml
lordsarcastic: No, unfortunately, there isn't a default. URLs are based on instances

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@ocean.on_start()
+async def on_start() -> None:
+    logger.info("Starting Port Ocean Dynatrace integration")
+    logger.info("Performing healthcheck")
+    dynatrace_client = initialize_client()
```
Tankilevitch, line 68:
```suggestion
    dynatrace_client = initialize_client()
    logger.info("Performing healthcheck")
```

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,70 @@
...
+        logger.info(f"Received batch with {len(problems)} problems")
+        yield problems
+
+
+@ocean.on_resync(ObjectKind.SLO)
+async def on_resync_slos(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    dynatrace_client = initialize_client()
+
+    async for slos in dynatrace_client.get_slos():
+        logger.info(f"Received batch with {len(slos)} SLOs")
```
Tankilevitch, line 38: you are repeating this log line in each resource in the `_get_paginated_resources`  with including the `resource_key` in the log to indicate to what resource it is related to.

integrations/dynatrace/main.py
```diff
@@ -0,0 +1,70 @@
...
+
+    async for entities in dynatrace_client.get_entities():
+        logger.info(f"Received batch with {len(entities)} entities")
+        yield entities
+
+
+@ocean.router.post("/webhook/problem")
+async def on_problem_event(event: dict[str, str | Any]) -> dict[str, bool]:
+    dynatrace_client = initialize_client()
+    logger.info(f"Received problem event: {event}")
```
Tankilevitch, line 54: please add a comment with a link to the webhook event that you relate to, so it will be easy to access

Tankilevitch (changes requested):
well done ! 
super minor comments

integrations/dynatrace/main.py
lordsarcastic: I'm sorry, I don't understand what you mean

PeyGis: In the     ```
async def _get_paginated_resources(
        self, url: str, resource_key: str, params: dict[str, Any] = {}
    ) -> AsyncGenerator[list[dict[str, Any]], None]:
```
       
        
        You are already logging the resources key
         `logger.info(f"Fetching {resource_key} from {url} with params {params}")`
         
         So I believe the comment is to remove the log line in the 
        ` logger.info(f"Received batch with {len(slos)} SLOs")`

File: integrations/dynatrace/integration.py [hunk 40]
```diff
@@ -0,0 +1,29 @@
...
+    ResourceConfig,
+    Selector,
+)
+from port_ocean.core.integrations.base import BaseIntegration
+from pydantic.fields import Field
+
+
+class DynatraceEntitySelector(Selector):
+    entity_types: list[str] = Field(
+        default=["APPLICATION", "SERVICE"],
```
Tankilevitch, line 13: I don't see that you are using it in any of the default mappings examples, I understand that its default but I think it will be more readable for users to understand how it works.

integrations/dynatrace/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+            target: .target
+            enabled: .enabled
+            warning: .warning
+            error: .error
+            errorBudget: .errorBudget
+            evaluatedPercentage: .evaluatedPercentage
+            evaluationType: .evaluationType
+            filter: .filter
+            relatedOpenProblems: .relatedOpenProblems
+            timeframe: .timeframe
```
Tankilevitch, line 62: what data does it show? lets remove the relatedOpenProblems

lordsarcastic: From the description, the `timeframe` is "The timeframe for the SLO evaluation.", the time it takes to evaluate the SLO.

integrations/dynatrace/integration.py
lordsarcastic: I am using this in the client. There are over 300 entity types and making nested calls 300 times when retrieving entities in `client.DynatraceClient::get_entities` would have performance hits. Instead we provide a default and the user gets to specify more if they want entities from additional types.

integrations/dynatrace/.port/resources/port-app-config.yaml
Tankilevitch: lets remove it as well

Tankilevitch: from the blueprint as well

integrations/dynatrace/integration.py
Tankilevitch: I couldn't find that you set it in the `port-app-config.yaml`

lordsarcastic: It isn't set since it is the default value. Which means that when it isn't set on `port-app-config.yaml` it holds `["APPLICATION", "SERVICE"]`

Tankilevitch: I understand, but I think it will give more context to the user, once he sees that the default mapping that he has contains filter on `APPLICATION` | `SERVICE`.

Tankilevitch (approved):
LGTM 🌊  well done @lordsarcastic

PR #394, opened 2024-02-21
(Bugfix) - Integrations crashing with regex error after 0.5.0

yairsimantov20:
# Description

What - Integrations were failing with regex error over the sensitive fields
Why - We are compiling the sensitive values to redact later from the logs and we didnt use escaping over those values
How - Used re.escape()

## Type of change


Tankilevitch (approved):
LGTM 🌊

PR #400, opened 2024-02-22
Update pr-labeler.yml

yairsimantov20:
Updating the labler event from `pull_request` to `pull_request_target` as it needs permissions to set tags on the pull request

https://stackoverflow.com/a/74959635


MPTG94 (approved):
LGTM

PR #403, opened 2024-03-03
Port 6993 ocean allow async calls in the jq entity processor

yairsimantov20:
# Description

What - Right now we cant run async logic from within the jq entities processor
Why - The functions are sync
How - changed the functions to be async & applied an async wrap for the jq

## Type of change


danielsinai (approved):
Nice one! :)

PR #491, opened 2024-04-09
[Integration][Gitlab] Add logs for event investigation

talru1123:
# Description

What - added logs for a more thorough indication of what's happening in event hook, and webhook creation.
Why - couldn't understand whats happening when an event in gitlab integration did not start it's handling properly.
How - added logs in various weak points in the process. 

## Type of change


integrations/gitlab/gitlab_integration/events/hooks/push.py
```diff
@@ -49,3 +49,9 @@ async def _on_hook(self, body: dict[str, Any], gitlab_project: Project) -> None:
                 f"Updating project information after push hook for project {gitlab_project.path_with_namespace}"
             )
             await ocean.register_raw(ObjectKind.PROJECT, [gitlab_project.asdict()])
+        else:
+            logger.info(
+                "The event was not sent on a branch that matches the configured branch, "
+                "therefore the changes will not be imported to port. The event's ref: "
+                f"{ref}, the configured branch as ref: {generate_ref(branch)}"
+            )
```
Tankilevitch, line 57: this is too spammy and long, I wouldn't want to add this log as I suspect most of the cases there are more commits to other branch ( while developing ) than the main branch, so it means the user will this message too much

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -27,6 +27,7 @@
 async def handle_webhook(group_id: str, request: Request) -> dict[str, Any]:
     event_id = f'{request.headers.get("X-Gitlab-Event")}:{group_id}'
     with logger.contextualize(event_id=event_id):
+        logger.info(f"Got an event from webhook. event_type and group_id: {event_id}")
```
Tankilevitch, line 30: you are missing the webhook type

Tankilevitch: I wonder if maybe this should be logged only when missed observer handler, wdyt?

Tankilevitch (changes requested):

integrations/gitlab/gitlab_integration/events/hooks/push.py
talru1123: you are correct, though when you get an event like that, you will log all of the info logs of the handling anyway, and wouldn't be able to realize if there is a bug or the branch is irrelevant.

integrations/gitlab/gitlab_integration/ocean.py
talru1123: what do you mean by webhook type? event_id holds template of: <EVENT_TYPE>:<GROUP_ID>
webhook_id initialized with group_id. for example: Push Hook:834848

talru1123: for the second comment: are we sure there is no other case where an event will not print logs because it failed before reaching the handler? if the only thing we are sure that could happen is unsupported observer then I agree, but I believe this log should be there for any case whatsoever.

Tankilevitch (approved):

PR #528, opened 2024-04-14
[Framework] Port 7567 bug ocean integration is being terminated due to OOM

omby8888:
# Description

What - ocean integration is being terminated due to OOM whene dealing with large amount of entities to process
Why - With asyncio.gather, all items are gathered into a single list before being processed, potentially causing memory issues if the list is too large, while asyncio.queue doesn't require holding all items in memory at once. 

## Type of change

## Screenshots

## API Documentation


Tankilevitch (changes requested):
bump version and add changelog

port_ocean/core/utils.py
```diff
@@ -40,26 +40,30 @@ def get_port_diff(
...
+    for entity in after:
+        key = (entity.identifier, entity.blueprint)
+        after_dict[key] = entity
+
+    # Find created, modified, and deleted objects
+    for key, obj in after_dict.items():
+        if key not in before_dict:
+            created.append(obj)
+        elif obj != before_dict[key]:
+            modified.append(obj)
```
Tankilevitch, line 63: why not using `is_same_entity`? 
using trying to evaluate the full object is much more expensive than just evaluating `identifier` and `blueprint`

port_ocean/utils/queue_utils.py
```diff
@@ -0,0 +1,52 @@
...
+        finally:
+            queue.task_done()
+
+
+async def process_in_queue(
+    objects_to_process: list[Any],
+    func: Callable[..., Coroutine[Any, Any, T]],
+    *args: Any,
+    concurrency: int = 50,
+) -> list[T]:
```
Tankilevitch, line 31: this function will be available for integrations as well ( as its in the `port_ocean/utils` ), lets make sure we add documentation to the function and an example, super important to add the **why & when** to use the function.

port_ocean/utils/queue_utils.py
```diff
@@ -0,0 +1,52 @@
...
+            asyncio.create_task(
+                _start_processor_worker(queue, func, processing_results)
+            )
+        )
+
+    for i in range(len(objects_to_process)):
+        await queue.put((objects_to_process[i], *args))
+
+    for i in range(concurrency):
+        await queue.put(None)
```
Tankilevitch, line 47: why is that?

Tankilevitch (changes requested):

port_ocean/core/utils.py
omby8888: saw that no need for the elif, could be replace simply with else

port_ocean/utils/queue_utils.py
omby8888: so the worker thread will stop working, if we wont do that then `raw_params = await queue.get()` will forever wait for more tasks and leave 50 workers threads open

Tankilevitch: then add comment as it is not straight forward

port_ocean/utils/queue_utils.py
```diff
@@ -0,0 +1,78 @@
...
+async def process_in_queue(
+    objects_to_process: list[Any],
+    func: Callable[..., Coroutine[Any, Any, T]],
+    *args: Any,
+    concurrency: int = 50,
+) -> list[T]:
+    """
+    This function takes a list of objects and process the within a queue buffer
+    allowing for smoother handling of bursts of incoming objects
+    This can help prevent overload and improve memory issues when dealing with large sset of data
```
Tankilevitch, line 35:
```suggestion
    This function executes multiple asynchronous tasks in a bounded way (e.g. having 200 tasks to execute, while running only 20 concurrently), to prevent overload and improve memory issues when dealing with large sets of data and tasks.
```

Tankilevitch (changes requested):
minor

Tankilevitch (approved):

PR #581, opened 2024-04-23
PORT-7618 | Dynatrace Ocean Integration with Invalid Entity Fields

lordsarcastic:
# Description

What - Fixed bug affecting Entity blueprints where fields: firstSeenTms, lastSeenTms, and tags; are missing
Why - So bugs won't be in the integration
How - Added query parameters to explicitly fetch these missing fields

## Type of change

## Screenshots
Screenshot 2024-04-17 at 18 40 16

## API Documentation
[Monitored Entity API](https://docs.dynatrace.com/docs/dynatrace-api/environment-api/entity-v2/get-entities-list#parameters)


PeyGis (changes requested):
@lordsarcastic for every new updates to the integration, we need to bump the integration version as well as creating changelog for it. This is missing from this PR. Please include them

yairsimantov20 (approved):
🌊

integrations/dynatrace/client.py
```diff
@@ -63,10 +63,15 @@ async def get_slos(self) -> AsyncGenerator[list[dict[str, Any]], None]:
     async def _get_entities_from_type(
         self, type_: str
     ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        ENTITY_FIELDS = "firstSeenTms,lastSeenTms,tags"
```
yairsimantov20, line 66: why not enabling it as part of the resource selector?

lordsarcastic: They are fields defined in the blueprint itself. If they are put in the resource selector, there is the possibility of users overriding it. By putting it here, we ensure they are included automatically. They are not meant to be customized.

yairsimantov20: Wdym? The blueprint is only a default option for modeling the data
The customer can and will change the blueprint to his needs
Therefore we need to allow him to pick which data he wants

lordsarcastic: If that's the case, which do you suggest:
- Take this out and make it as a default option on the selector
- Take this out, take the corresponding fields out in the default blueprint and add it to the selector.
- Any other option...

yairsimantov20: First option should be good

integrations/opencost/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
+# Port_Ocean 0.1.9 (2024-04-26)
```
yairsimantov20, line 10: 0.1.9?

yairsimantov20: why is in the opencost?

integrations/dynatrace/CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
+# Port_Ocean 0.1.14 (2024-05-24)
```
yairsimantov20, line 10: update changelog date

integrations/dynatrace/client.py
```diff
@@ -61,20 +61,26 @@ async def get_slos(self) -> AsyncGenerator[list[dict[str, Any]], None]:
...
+        self, type_: str, entity_fields: EntityFieldsType | None
     ) -> AsyncGenerator[list[dict[str, Any]], None]:
         async for entities in self._get_paginated_resources(
             f"{self.host_url}/entities",
             "entities",
-            params={"entitySelector": f'type("{type_}")', "pageSize": 100},
+            params={
+                "entitySelector": f'type("{type_}")',
+                "pageSize": 100,
+                "fields": entity_fields or "",
```
yairsimantov20, line 72: add the fields property only when u have entity fields

yairsimantov20 (changes requested):

lordsarcastic:
PR moved to [non fork](https://github.com/port-labs/ocean/pull/742)

Issue #577, opened 2024-04-21
[Bug] [Integration] [argoCD] : failing to synchronize all apps


sgandon:
### Ocean Framework Version

🌊 Ocean version: 0.5.12

### Integration Version

🚢 Integration version: 0.1.38

### Steps to reproduce

launched the "docker run command"
may entities where ingested properly but one is failing

### What did you expect to see?

everything worked

### What did you see instead?

the following error
```
2024-04-21 09:30:22.551 | INFO     | Sending request to ArgoCD API: GET https://readacted.domain.com/api/v1/applications/messaging/managed-resources
2024-04-21 09:30:22.594 | ERROR    | Failed to execute resync function, error: 'items'
Traceback (most recent call last):
  File "/usr/local/bin/ocean", line 8, in <module>
    sys.exit(cli_start())
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.11/site-packages/port_ocean/cli/commands/sail.py", line 74, in sail
    run(path, log_level, port, initialize_port_resources, override)
  File "/usr/local/lib/python3.11/site-packages/port_ocean/run.py", line 57, in run
    uvicorn.run(app, host="0.0.0.0", port=application_settings.port)
  File "/usr/local/lib/python3.11/site-packages/uvicorn/main.py", line 578, in run
    server.run()
  File "/usr/local/lib/python3.11/site-packages/uvicorn/server.py", line 61, in run
    return asyncio.run(self.serve(sockets=sockets))
  File "/usr/local/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
  File "/usr/local/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
[...]

PR #588, opened 2024-04-25
[Integration][ArgoCD] - Fix managed resources KeyError

PeyGis:
# Description

What - A PR to fix the issue reported [here](https://github.com/port-labs/ocean/issues/577) when trying to get item manage resources items 
Why -
How - it is understood that some apps may not be managing any k8s resources. This causes the ArgoCD to return null, or without the items keyword. So instead of using a square bracket to fetch the response, we now use .get("items", [])

## Type of change

## Screenshots

## API Documentation

### DIFF
--- a/integrations/argocd/.port/resources/port-app-config.yaml
+++ b/integrations/argocd/.port/resources/port-app-config.yaml
@@ -69,7 +69,7 @@ resources:
     selector:
       query: "true"
     port:
-      itemsToParse: .status.history
+      itemsToParse: .status.history // []
       entity:
         mappings:
           identifier: .metadata.uid + "-" + (.item.id | tostring)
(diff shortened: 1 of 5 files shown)


yairsimantov20 (approved):
🌊

integrations/argocd/.port/resources/port-app-config.yaml
```diff
@@ -69,7 +69,7 @@ resources:
     selector:
       query: "true"
     port:
-      itemsToParse: .status.history
+      itemsToParse: .status.history // []
```
sgandon, line 72: strange comment here ?

PeyGis: Screenshot 2024-04-26 at 1 24 26 PM

Add the `// []` takes care of these warnings that spams the log

PR #591, opened 2024-04-26
[Integration][PagerDuty] - Map oncall user with highest escalation level

PeyGis:
# Description

What - Users reported that the oncall property in their PagerDuty service blueprint contains array of users, while in fact, it should just be one user
Why - Their PagerDuty service dashboard also correctly lists only one user as oncall
How - After investigation, it was understood that an escalation policy can have many levels. On the PagerDuty dashboard, the user in the escalation policy with level 1 get's assigned as the primary oncall user. Here is the current service discovery dashboard in our PD account:

Screenshot 2024-04-26 at 12 50 57 PM

## Type of change

## Screenshots
PD Service in Port
image

## API Documentation

[Who is on call](https://developer.pagerduty.com/api-reference/3a6b910f11050-list-all-of-the-on-calls)

### DIFF
--- a/integrations/pagerduty/.port/resources/blueprints.json
+++ b/integrations/pagerduty/.port/resources/blueprints.json
@@ -31,11 +31,8 @@
         },
         "oncall": {
           "title": "On Call",
-          "type": "array",
-          "items": {
-            "type": "string",
-            "format": "user"
-          }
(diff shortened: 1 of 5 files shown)


yairsimantov20 (approved):
🌊

PR #652, opened 2024-05-26
improvement(Port 8335) ocean make config yml not required

yairsimantov20:
# Description

What - There are a lot of unused config.yml files
Why - Most of the time we prefer to control it with the env vars
How - made it unrequired and only using it for defaults

## Type of change


port_ocean/cli/cookiecutter/{{cookiecutter.integration_slug}}/.port/spec.yaml
```diff
@@ -1,4 +1,3 @@
-type: {{cookiecutter.integration_slug}}
```
omby8888, line 1: ?

omby8888 (approved):

PR #682, opened 2024-06-02
linter matrix

yairsimantov20:
# Description

What - Lint takes a lot of time
Why - Linting all integrations every time
How - Running only the changed scopes lint with matrix

## Type of change

image


.github/workflows/lint.yml
```diff
@@ -17,5 +52,10 @@ jobs:
         uses: actions/checkout@v4
       - name: Lint
         run: |
-          make install/all
-          make lint/all
+          if [ "${{ matrix.folder }}" == "." ]; then
+            cd .
```
MPTG94, line 56: The if can be replaced with an inequality check to achieve the same result

MPTG94 (approved):
LGTM

PR #699, opened 2024-06-07
[Integration][PagerDuty] Show also secondaries on call on PagerDuty Service

lordsarcastic:
# Description

What:
- Added secondary on call user
- Changed calculation for oncall user to ordering by escalation level instead of filtering by the highest level.

Why:
- By ordering, it is guaranteed that both secondary and primary oncall are chosen correctly and are not the same user

How: Ordered on call by escalation level

## Type of change

## Screenshots
Screenshot 2024-06-07 at 17 26 02

## API Documentation
[List of services](https://developer.pagerduty.com/api-reference/e960cca205c0f-list-services)


matan84 (commented):
missing version bump, everything else lgtm

lordsarcastic:
Done @matan84

Tankilevitch (approved):

PR #750, opened 2024-06-26
[Integration][sonarQube] - Overcome 10k issue limit imposed by SonarQube

PeyGis:
# Description

What - SonarQube imposes a [hard limit of 10k records](https://community.sonarsource.com/t/cannot-get-more-than-10000-results-through-web-api/3662/2) on their [issues/search](https://next.sonarqube.com/sonarqube/web_api/api/issues/search) endpoint. This makes the integration not useful for users who would want to understand their sonar issues
Why -
How - The recommended approach is to use the [/search/list endpoint](https://next.sonarqube.com/sonarqube/web_api/api/issues/list?internal=true), which doesn't have the 10k limit. The good news is that both /search and /list return the same issue response so no blueprint property is lost. The only difference is in the paging response. The list endpoint does not return the total records so the get_paginated_resource() was updated accordingly to handle this case

## Type of change

## Screenshots

## API Documentation

[Issues List](https://next.sonarqube.com/sonarqube/web_api/api/issues/list?internal=true)


Tankilevitch (approved):
LGTM

PR #753, opened 2024-06-27
[Integration][Azure] Add support for multiple installation methods

Tankilevitch:
# Description

What - Added support for multiple installation methods
Why - 
How - Added support for passing azure app registration credentials as integration configuration.

## Type of change

## Screenshots

## API Documentation


integrations/azure/CHANGELOG.md
matan84, line 1: Also pyproject.toml

integrations/azure/azure_integration/ocean.py
```diff
@@ -279,3 +280,41 @@ async def cloud_event_validation_middleware_handler(
...
+
+@ocean.on_start()
+async def on_start() -> None:
+
+    logger.info("Setting up credentials for Azure client")
+    azure_client_id = ocean.integration_config.get("azure_client_id")
+    azure_client_secret = ocean.integration_config.get("azure_client_secret")
+    azure_tenant_id = ocean.integration_config.get("azure_tenant_id")
+    azure_subscription_id = ocean.integration_config.get("azure_subscription_id")
+    if not azure_client_id and not azure_client_secret and not azure_tenant_id:
```
matan84, line 293: Is this correct? Can the integration run without one of them?

matan84 (changes requested):
Small changes required

integrations/azure/azure_integration/ocean.py
Tankilevitch: if it wasn't specified, it will try to use the service account that might be running the integration ( e.g. in terraform we run the container with service account )

matan84: But can i just provide the integration with `azure_client_id` for example? Shouldn't this if make sure that all the optional variables are provided? If the service account isn't provided, then all variables must be provided for the integration to run, no?

matan84 (approved):
LGTM

PR #761, opened 2024-07-02
[PORT-8828] [AWS] Ensure default region for global resources

erikzaadi:
# Description

What - Ensure that there's a default region for global AWS resources
Why - Creating an instance of the cloudcontrol client would fail if no AWS default region was known
How - Lookup default regions from the credentials

## Type of change

### DIFF
--- a/integrations/aws/utils/aws.py
+++ b/integrations/aws/utils/aws.py
@@ -1,11 +1,13 @@
-from typing import Any, AsyncIterator, Optional
+from typing import Any, AsyncIterator, Optional, Union
+
 import aioboto3
 from loguru import logger
-from aws.session_manager import SessionManager
+from port_ocean.context.event import event
 from port_ocean.context.ocean import ocean
 from starlette.requests import Request
-from port_ocean.context.event import event
 
+from aws.session_manager import SessionManager
+from aws.aws_credentials import AwsCredentials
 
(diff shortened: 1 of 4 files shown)


integrations/aws/utils/aws.py
```diff
@@ -32,6 +34,14 @@ def describe_accessible_accounts() -> list[dict[str, Any]]:
     return _session_manager._aws_accessible_accounts
 
 
+def get_default_region_from_credentials(
+    credentials: AwsCredentials,
+) -> Union[str, None]:
+    return (
+        credentials.default_regions[0] if len(credentials.default_regions) > 0 else None
```
matan84, line 41:
```suggestion
        credentials.default_regions[0] if credentials.default_regions else None
```

integrations/aws/utils/aws.py
```diff
@@ -44,8 +54,9 @@ async def get_sessions(
 
     if custom_account_id:
         credentials = _session_manager.find_credentials_by_account_id(custom_account_id)
+        default_region = get_default_region_from_credentials(credentials)
```
matan84, line 57: Move this to the `if use_default_region == true`

matan84 (changes requested):
Nice job, add a changelog + version bump in the pyproject.toml .

Tankilevitch (approved):

PR #766, opened 2024-07-03
Allow project specific webhook creation for project admins

phalbert:
# Description

The genesis of this change is documented in this [slack thread](https://port-community.slack.com/archives/C06UR5F2WCX/p1717488013161949)

**What** 
- Users encounter a 400 error when attempting to create subscriptions (webhooks) due to insufficient permissions to edit the subscription. 
- Modify the webhook creation process to include the projectId in the payload.

**Why** 
- The current setup assumes an organisation token when creating webhooks. If a project-scoped token is supplied, the craetion fails

**How**
- When the user holds a project administrator role, and the PAT is scoped only to that specific project, it is necessary to explicitly define the projectId when creating a webhook.
```json
{
  "publisherId": "tfs",
  "eventType": "git.pullrequest.updated",
  "consumerId": "webHooks",
  "consumerActionId": "httpRequest",
  "consumerInputs": {
    "url": "http://20.246.193.102/ado-port/integration/webhook"
  },
  "publisherInputs": {
    "projectId": "4901a905-d07b-4027-b7cc-577d2b6a0e31"
  }
}
```

## Type of change

## Screenshots

## API Documentation

### DIFF
--- a/integrations/azure-devops/.port/spec.yaml
+++ b/integrations/azure-devops/.port/spec.yaml
@@ -15,6 +15,10 @@ configurations:
     required: true
     type: string
     sensitive: true
+  - name: isProjectAdmin
+    required: false
+    type: boolean
+    description: If the user is a project admin.
   - name: appHost
     required: false
     type: url
--- a/integrations/azure-devops/main.py
+++ b/integrations/azure-devops/main.py
@@ -24,7 +24,16 @@ async def setup_webhooks() -> None:
         return
 
     azure_devops_client = AzureDevopsClient.create_from_ocean_config()
-    await setup_listeners(ocean.integration_config["app_host"], azure_devops_client)
+    if ocean.integration_config.get("is_project_admin", False):
+        async for projects in azure_devops_client.generate_projects():
+            for project in projects:
+                await setup_listeners(
+                    ocean.integration_config["app_host"], azure_devops_client, project["id"]
+                )
+    else:
+        await setup_listeners(
+            ocean.integration_config["app_host"], azure_devops_client
+        )
 
 
 @ocean.on_resync(Kind.PROJECT)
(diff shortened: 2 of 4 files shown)


integrations/azure-devops/.port/spec.yaml
```diff
@@ -15,6 +15,10 @@ configurations:
     required: true
     type: string
     sensitive: true
+  - name: isProjectAdmin
```
matan84, line 18: I think this name is misleading. Perhaps this should be `isProjectsLimited` or something like this and be set to true by default. 
We shouldn't assume that we have organization-wide permissions.

integrations/azure-devops/main.py
```diff
@@ -24,7 +24,16 @@ async def setup_webhooks() -> None:
         return
 
     azure_devops_client = AzureDevopsClient.create_from_ocean_config()
-    await setup_listeners(ocean.integration_config["app_host"], azure_devops_client)
+    if ocean.integration_config.get("is_project_admin", False):
```
matan84, line 27: This should be true by default.

matan84 (changes requested):
Overall, Really nice and clean job!
Please also update the changelog + pyproject.toml
Additionally, this will require a docs change.

phalbert:
will open a PR for the docs change

integrations/azure-devops/main.py
```diff
@@ -24,7 +24,16 @@ async def setup_webhooks() -> None:
         return
 
     azure_devops_client = AzureDevopsClient.create_from_ocean_config()
-    await setup_listeners(ocean.integration_config["app_host"], azure_devops_client)
+    if ocean.integration_config.get("is_project_admin", True):
```
matan84, line 27: Fix this integration config key

matan84 (changes requested):

matan84 (approved):
Nice Job!

PR #767, opened 2024-07-03
[Integration][Gitlab] Added support for gitlab member ingestion

mk-armah:
# Description

**Added support for GitLab member ingestion.**

This PR introduces two new kinds: `PROJECTWITHMEMBERS` and `GROUPWITHMEMBERS`. These are enhanced versions of the original `PROJECT` and `GROUP` kinds, now including member data.

- **Public Email Visibility:**
  - GitLab group and project members do not include users' emails by default, especially for free plans. To enable users on free plans to view user emails, we have added a flag `includePublicEmail` to the member selector.
  - **Usage:** If set to `true`, members are enriched with the `public_email` field from the `/users` endpoint. This is dependent on whether the user has allowed public email visibility in their GitLab account settings. [Learn more](https://forum.gitlab.com/t/retrieve-user-emails-via-api/33448).
  - **Note:** It was necessary to call the `/users` endpoint because the members API does not include `public_email`.
  - **Default Value:** The default value for `includePublicEmail` is `false`.

- **Bot Members Filtering:**
  - GitLab returns all members, including bots and tokens created as members. To filter out bots from actual members, we have added a flag `includeBotMembers` to the MemberSelector.
  - **Usage:** If set to `false`, bot members are excluded from the synchronization process.
  - **Default Value:** The default value for `includeBotMembers` is `true`, meaning bots are included by default.
  - **Reasoning:** Bot filtering is necessary when syncing  members to ensure only actual user accounts are considered.

- **Inherited Members Inclusion:**
  - GitLab allows inclusion of inherited members from parent groups. To control this behavior, we have added a flag `includeInheritedMembers`.
  - **Usage:** If set to `true`, members inherited from parent groups are included during synchronization.
  - **Default Value:** The default value for `includeInheritedMembers` is `false`, so inherited members are excluded by default.
  - **Note:** Including inherited members can provide a more comprehensive view of all members who have access to a project or group through group hierarchies.

## Type of change

## Screenshots

image

image


integrations/gitlab/.port/resources/blueprints.json
```diff
@@ -53,5 +53,65 @@
...
+          "description": "GitLab primary email address.",
+          "icon": "User",
+          "format": "user"
+        },
+        "publicEmail": {
+          "type": "string",
+          "title": "Public Email",
+          "description": "User's GitLab public email.",
+          "icon": "User",
+          "format": "user"
```
Tankilevitch, line 100: in what case the user will have a public email? I think we might want to remove it by default

integrations/gitlab/.port/resources/port-app-config.yaml
```diff
@@ -15,3 +15,21 @@ resources:
...
+          blueprint: '"member"'
+          properties:
+            state: .state
+            locked: .locked
+            link: .web_url
+            email: .email
+            publicEmail: .__public_email
+          relations:
+            gitlabGroup: '[.__groups[].full_path]'
+            createdBy: .created_by.username
```
Tankilevitch, line 35: in the blueprint relation you only have gitlabGroup, while in your mapping you have both createdBy and gitlabGroup. I am not sure it is of interest for users who created the user, lets remove it. Let me know if you think otherwise

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -533,6 +550,84 @@ async def get_all_issues(self, group: Group) -> typing.AsyncIterator[List[Issue]
...
+        async def check_group_membership(group: Group) -> Group | None:
+            "check if the user is a member of the group"
+            async with semaphore:
+                try:
+                    await AsyncFetcher.fetch_single(group.members.get, member.get_id())
+                    return group
+                except GitlabError as err:
+                    if err.response_code != 404:
+                        raise err
+                    return None
```
Tankilevitch, line 568: this feels very insufficient, eventually for all the if I have 1 member, and 1000 groups, I'll have to query this api 1000 times to find out what groups he related to? isn't there any other way to get that data?

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -533,6 +550,84 @@ async def get_all_issues(self, group: Group) -> typing.AsyncIterator[List[Issue]
...
+    async def enrich_member_with_groups_and_public_email(
+        self, member
+    ) -> dict[str, Any]:
+        user: User = await self.get_user(member.id)
+
+        user_groups: List[dict[str, Any]] = [
+            {"id": group.id, "full_path": group.full_path}
+            async for groups in self.get_member_groups(user)
+            for group in groups
+        ]
```
Tankilevitch, line 611: I am not sure this should be the default behavior when quering members.

Tankilevitch: maybe this should be something that is part of group members kind? that will query the list of members of each group. 
and instead of having a relation of user -> groups, we will have relation between group -> users

Tankilevitch (changes requested):

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -533,6 +550,84 @@ async def get_all_issues(self, group: Group) -> typing.AsyncIterator[List[Issue]
...
+            sort="asc",
+        ):
+            members: List[GroupMemberAll] = typing.cast(
+                List[GroupMemberAll], users_batch
+            )
+            logger.info(
+                f"Queried {len(members)} members {[user.username for user in members]} from {group.name}"
+            )
+            yield members
+
```
Tankilevitch, line 601:
```suggestion
    async def get_all_group_members(
        self, group: Group
    ) -> typing.AsyncIterator[List[GroupMemberAll]]:

        logger.info(f"Fetching all members of group {group.name}")

        async for users_batch in AsyncFetcher.fetch_batch(
            fetch_func=group.members_all.list,
            validation_func=self.should_run_for_member,
            pagination="offset",
            order_by="id",
            sort="asc",
        ):
            members: List[GroupMemberAll] = typing.cast(
                List[GroupMemberAll], users_batch
            )
            logger.info(
                f"Queried {len(members)} members {[user.username for user in members]} from {group.name}"
            )
            members_enriched_with_group  = [ {...member, group: group} for member in members ] 
            yield members

```

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -533,6 +550,84 @@ async def get_all_issues(self, group: Group) -> typing.AsyncIterator[List[Issue]
...
+    async def enrich_member_with_groups_and_public_email(
+        self, member
+    ) -> dict[str, Any]:
+        user: User = await self.get_user(member.id)
```
Tankilevitch, line 605: feature flag in the selector mapping, and defined in docs.

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -207,3 +207,16 @@ async def resync_pipelines(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+async def resync_members(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    for service in get_cached_all_services():
+        for group in service.get_root_groups():
+            async for members_batch in service.get_all_group_members(group):
+                tasks = [
+                    service.enrich_member_with_groups_and_public_email(member)
+                    for member in members_batch
+                ]
+                members = await asyncio.gather(*tasks)
+                yield members
```
Tankilevitch, line 222:
```suggestion
@ocean.on_resync(ObjectKind.GROUP_MEMBERS)
async def resync_members(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
    for service in get_cached_all_services():
        for group in service.get_root_groups():
           group_members = []
            async for members_batch in service.get_all_group_members(group):
                group_memebers.append(members_batch)
            yield { group: group, group_members: group_members } 
```


`stream_async_iterators_tasks`

integrations/gitlab/.port/resources/blueprints.json
```diff
@@ -53,5 +53,65 @@
...
+    "mirrorProperties": {},
+    "calculationProperties": {},
+    "aggregationProperties": {},
+    "relations": {
+      "gitlabGroup": {
+        "title": "Group",
+        "target": "gitlabGroup",
+        "required": false,
+        "many": true
+      }
```
Tankilevitch, line 114: missing group

Tankilevitch (changes requested):

integrations/gitlab/.port/resources/blueprints.json
```diff
@@ -53,5 +53,99 @@
...
+          "type": "string",
+          "icon": "GitLab",
+          "description": "The current state of the GitLab item (e.g., open, closed)."
+        },
+        "locked": {
+          "type": "string",
+          "title": "Locked",
+          "icon": "GitLab",
+          "description": "Indicates if the GitLab item is locked."
+        },
```
Tankilevitch, line 74: how is this locked?

integrations/gitlab/.port/resources/blueprints.json
```diff
@@ -53,5 +53,99 @@
...
+    "title": "Member",
+    "icon": "GitLab",
+    "schema": {
+      "properties": {
+        "state": {
+          "title": "State",
+          "type": "string",
+          "icon": "GitLab",
+          "description": "The current state of the GitLab item (e.g., open, closed)."
+        },
```
Tankilevitch, line 68: we are talking about member not gitlab item, please make sure its readable and straight forward for the users

integrations/gitlab/.port/resources/blueprints.json
```diff
@@ -53,5 +53,99 @@
...
+          "enum": [
+            "public",
+            "internal",
+            "private"
+          ],
+          "enumColors": {
+            "public": "red",
+            "internal": "yellow",
+            "private": "green"
+          }
```
Tankilevitch, line 123: add description

integrations/gitlab/.port/resources/port-app-config.yaml
```diff
@@ -15,3 +15,23 @@ resources:
...
```
Tankilevitch, line 37: what is `createdBy` we don't have that kind of relation please remove

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -108,7 +111,9 @@ async def on_start() -> None:
 async def resync_groups(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
     for service in get_cached_all_services():
         async for groups_batch in service.get_all_groups():
-            yield [group.asdict() for group in groups_batch]
+            tasks = [service.enrich_group_with_members(group) for group in groups_batch]
+            enriched_groups = await asyncio.gather(*tasks)
+            yield enriched_groups
```
Tankilevitch, line 116: this is not needed any more if i recall correctly, we moved the logic to the member

Tankilevitch (changes requested):

integrations/gitlab/.port/resources/port-app-config.yaml
```diff
@@ -15,3 +15,23 @@ resources:
...
+        mappings:
+          identifier: .username
+          title: .name
+          blueprint: '"member"'
+          properties:
+            state: .state
+            locked: .locked
+            link: .web_url
+            email: .email
+            publicEmail: .__public_email
```
Tankilevitch, line 34: lets remove by default

integrations/gitlab/.port/resources/blueprints.json
```diff
@@ -53,5 +53,99 @@
...
+    "mirrorProperties": {},
+    "calculationProperties": {},
+    "aggregationProperties": {},
+    "relations": {
+      "members": {
+        "title": "Members",
+        "target": "member",
+        "required": false,
+        "many": true
+      }
```
Tankilevitch, line 148: shouldn't be here

integrations/gitlab/.port/resources/blueprints.json
```diff
@@ -53,5 +53,99 @@
...
+          "icon": "User",
+          "format": "user"
+        }
+      },
+      "required": []
+    },
+    "mirrorProperties": {},
+    "calculationProperties": {},
+    "aggregationProperties": {},
+    "relations": {}
```
Tankilevitch, line 102: missing relation

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -13,6 +13,8 @@
     Issue,
     Group,
     ProjectPipeline,
+    User,
+    GroupMember,
```
Tankilevitch, line 17: which one?

mk-armah: both, depends on the context

integrations/gitlab/.port/resources/blueprints.json
mk-armah: relation is group -> member

integrations/gitlab/.port/resources/blueprints.json
mk-armah: relationship is group -> member

integrations/gitlab/gitlab_integration/ocean.py
mk-armah: The other way round instead

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -535,6 +549,77 @@ async def get_all_issues(self, group: Group) -> typing.AsyncIterator[List[Issue]
...
+            def should_run_for_member(member: GroupMember) -> bool:
+                return not member.username.__contains__("bot")
+
+            validation_func: Union[
+                Callable[[User], bool], Callable[[GroupMember], bool]
+            ] = (should_run_for_member if filter_bots else skip_validation)
+
+            logger.info(f"Fetching all members of group {group.name}")
+            async for members_batch in AsyncFetcher.fetch_batch(
+                fetch_func=group.members.list,
```
Tankilevitch, line 574: why `group.member.list` and not `group.member_all.list` ?
https://python-gitlab.readthedocs.io/en/stable/gl_objects/groups.html#id10 
what do you think about adding this as an option to query all rather than only in one hierarchy?

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -207,3 +212,21 @@ async def resync_pipelines(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+    for service in get_cached_all_services():
+        for group in service.get_root_groups():
+            async for members in service.get_all_group_members(group):
+                if selector.public_email_visibility:
+                    yield [
+                        await service.enrich_member_with_public_email(member)
+                        for member in members
+                    ]
+                else:
+                    yield [member.asdict() for member in members]
```
Tankilevitch, line 232: what happens when I have the same user in multiple groups? how would that behave? will I have to perform repeated upserts?

maybe in this method ^ we should use `members = group.members_all.list(get_all=True)` which will return all and reduce the amount of extra requests that we will have to perform?

Tankilevitch (changes requested):

integrations/gitlab/gitlab_integration/ocean.py
mk-armah: my reason for using members as opposed to member_all was because the members_all request returns not the the user in that group but also all inherited and invited members.

Thereby resulting in all groups the same members since the members most commonly belong to the parent group - [details](https://docs.gitlab.com/ee/user/group/subgroups/#:~:text=Inherited%20members%20of%20the%20subgroup,the%20subgroup%20or%20its%20ancestors.)

aside this, the behavior and how we will retrieve members does not differ from member and members_all

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -131,7 +143,12 @@ class GitlabPortAppConfig(PortAppConfig):
     project_visibility_filter: str | None = Field(
         alias="projectVisibilityFilter", default=None
     )
-    resources: list[GitlabResourceConfig] = Field(default_factory=list)  # type: ignore
+    filter_bots: bool | None = Field(
+        alias="filterBots",
+        default=False,
+        description="If set to true, bots will be filtered out from the members list. Default value is false",
+    )
```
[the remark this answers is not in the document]
mk-armah: Both Members and Groups depend on this parameter. Removing from top level means I have to include it in integration for groups and members. Please confirm this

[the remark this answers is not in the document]
mk-armah: Also, I placed it at the top level to keep a consistent behavior in this system since groups are related to user. I strictly expect the value of filterBots to be consistent for groups and members. What could go wrong is that a user might specify this parameter as  false for groups kind and true for member kind. Due to the relationship between members and groups, the catalog will be populated with extra inconsistent data.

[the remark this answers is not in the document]
Tankilevitch: I would add a comment above the `filter_bots` so other developers will understand your motivation.
Also I would rename it to `include_member_bots` and default should be `false`

integrations/gitlab/gitlab_integration/ocean.py
Tankilevitch: ok i understand what you are saying, so if we use the `members_all` only for the members kind wouldn't it reduce the amount of requests by a lot? as we will only have to bring the members for the root groups rather than the `subgroups` as well. 

also just making sure that you have tested subgroups as well. please confirm

Tankilevitch (changes requested):
Also missing webhook handling for new members.

integrations/gitlab/.port/resources/port-app-config.yaml
```diff
@@ -1,6 +1,35 @@
...
+        mappings:
+          identifier: .full_path
+          title: .name
+          blueprint: '"gitlabGroup"'
+          properties:
+            url: .web_url
+            visibility: .visibility
+            description: .description
+          relations:
+            members: '[.__members[].username]'
```
Tankilevitch, line 32: I think adding members should be optional for groups, so customers will be able to decide whether they want to have it or no

Tankilevitch: and need to check that creating a new group, will sync the members correctly

integrations/gitlab/gitlab_integration/ocean.py
mk-armah: calling `/members` on root groups returns the same results as `/members/all`, calling `/members` on subgroups comes with less data than `members/all`, due to the inclusion of invited and inherited members, for root groups, concept of inherited members does not apply, all members of the root groups are returned regardless how we choose to call them.

I believe the optimization here was getting members from root groups instead of all groups (including subgroups), which would have taken more time.

integrations/gitlab/.port/resources/port-app-config.yaml
mk-armah: fair, how about project -> group, same ?

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -207,3 +212,35 @@ async def resync_pipelines(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+            ]
+            enriched_members = await asyncio.gather(*enriched_member_tasks)
+            return enriched_members
+        else:
+            member_dicts = [
+                member.asdict()
+                async for members in service.get_all_group_members(group)
+                for member in members
+            ]
+            return member_dicts
```
Tankilevitch, line 239: why not using the `get_all_group_members`  once and then if `enrich_with_public_email` spawn requests for the members

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -207,3 +212,35 @@ async def resync_pipelines(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+
+    for service in get_cached_all_services():
+        async for groups in service.get_all_groups(skip_validation=True):
+            group_tasks = [fetch_group_members(service, group) for group in groups]
+            for group_task in asyncio.as_completed(group_tasks):
+                group_members = await group_task
+                yield group_members
```
Tankilevitch, line 246: I am confused which is one is the right one? you both return members_dics and the group_members in the same resync method? how should it work?

Tankilevitch (changes requested):

integrations/gitlab/gitlab_integration/ocean.py
mk-armah: Members dict is returned by the `fetch_group_members` function which is responsible for making the decision as to whether to enrich or not enrich a group member #

mk-armah: at `group_tasks = [fetch_group_members(service, group) for group in groups]`, we leverage `fetch_group_members` to fetch all members of the each group and process them as completed.

integrations/gitlab/gitlab_integration/ocean.py
mk-armah: Great, the code will look like this then


`members = [
    member
    async for members_batch in service.get_all_group_members(group)
    for member in members_batch
]

if selector.enrich_with_public_email:
    enriched_member_tasks = [
        service.enrich_member_with_public_email(member)
        for member in members
    ]
    enriched_members = await asyncio.gather(*enriched_member_tasks)
    result = enriched_members
else:
    result = [member.asdict() for member in members]

return result`

mk-armah: I'll refactor

mk-armah: This is better for readability, thanks

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -108,7 +111,9 @@ async def on_start() -> None:
 async def resync_groups(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
     for service in get_cached_all_services():
         async for groups_batch in service.get_all_groups():
-            yield [group.asdict() for group in groups_batch]
+            tasks = [service.enrich_group_with_members(group) for group in groups_batch]
+            enriched_groups = await asyncio.gather(*tasks)
```
Tankilevitch, line 115: when we do such things, we should add it under a feature flag, as we wouldn't want to just do all those extra requests to get the members if it wasn't actually intended by the user.

Therefor I suggest leaving the `group` kind as it was.
And add the following kinds.

`groupMembers` - which will do this exact logic that you have implemented in `resync_groups` - this kind will allow users to decide whether they want to export the group with or without members.

`member / user` - which will bring the information about a user. such as the email etc..

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -207,3 +212,34 @@ async def resync_pipelines(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+            return enriched_members
+
+        return [member.asdict() for member in members]
+
+    for service in get_cached_all_services():
+        async for groups in service.get_all_groups(skip_validation=True):
+            group_tasks = [process_group_members(service, group) for group in groups]
+            for group_task in asyncio.as_completed(group_tasks):
+                group_members = await group_task
+                yield group_members
```
Tankilevitch, line 245: what will happen when a member is in multiple groups? will it get overwritten in port with the latest group that was fetched?

Tankilevitch (changes requested):

integrations/gitlab/gitlab_integration/ocean.py
mk-armah: Yes it would, a function to cache members and return only unsynced / uncached members can solve this. So we don't have to hit port with members that have been synced already

integrations/gitlab/gitlab_integration/events/hooks/group.py
```diff
@@ -1,39 +1,43 @@
...
-
-        group_id = body["group_id"] if "group_id" in body else body["group"]["id"]
-
-        logger.info(f"Handling hook {event} for group {group_id}")
-
-        group = await self.gitlab_service.get_group(group_id)
+    async def _on_hook(
+        self, body: dict[str, Any], gitlab_group: Optional[Group]
+    ) -> None:
+        logger.info(f"Handling {body['event_name']} for group {body['group_id']}")
```
Tankilevitch, line 18: lets set `event_name = body["event_name"]` like we had previously

integrations/gitlab/gitlab_integration/events/hooks/group.py
```diff
@@ -1,39 +1,43 @@
...
+            )
+            await self._register_object_with_members(
+                ObjectKind.GROUPWITHMEMBERS, gitlab_group
+            )
+            logger.info(f"Registered group {body['group_id']}")
         elif (
             group_full_path
             and self.gitlab_service.should_run_for_path(group_full_path)
-            and event_name in ("subgroup_destroy", "group_destroy")
+            and body["event_name"] in ("subgroup_destroy", "group_destroy")
```
Tankilevitch, line 33: use `event_name` rather than `body["event_name"]`

integrations/gitlab/gitlab_integration/events/hooks/base.py
```diff
@@ -49,3 +93,21 @@ async def on_hook(self, event: str, body: dict[str, Any]) -> None:
...
         pass
+
+
+class GroupHandler(HookHandler):
+    async def on_hook(self, event: str, body: dict[str, Any]) -> None:
+        group_id = body.get("group_id", body.get("group", {}).get("id"))
+        group = await self.gitlab_service.get_group(group_id)
+        await self._on_hook(body, group)
+        group_path = body.get("full_path", body.get("group_path"))
+        logger.info(f"Finished handling {event} for group {group_path}")
```
Tankilevitch, line 104: we are missing a log here to indicate that we started handling event

integrations/gitlab/gitlab_integration/events/utils.py
Tankilevitch, line 1: this doesn't seems used anywhere

integrations/gitlab/gitlab_integration/git_integration.py
```diff
@@ -121,6 +121,33 @@ class GitlabResourceConfig(ResourceConfig):
...
+        default=False,
+        description="If set to true, the integration will enrich group members with public email field. Default value is false",
+    )
+    include_inherited_members: bool = Field(
+        alias="includeInheritedMembers",
+        default=False,
+        description="If set to true, the integration will include inherited members in the group members list. Default value is false",
+    )
+    # The "include bot members" flag affects both the "group" and "member" kinds.
+    # To prevent inconsistencies, the behavior or value of this parameter should be consistent for both "groups" and "members".
```
Tankilevitch, line 137:
```suggestion
    # To prevent inconsistencies, the behavior or value of this parameter should be consistent for both "groups-with-members" and "project-with-members".
```

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -29,14 +30,23 @@
 
 from port_ocean.context.event import event
 from port_ocean.core.models import Entity
+from port_ocean.utils.cache import cache_iterator_result
+import functools
 
 PROJECTS_CACHE_KEY = "__cache_all_projects"
+GROUPS_CACHE_KEY = "__cache_all_groups"
+MEMBERS_CACHE_KEY = "__cache_all_members"
```
Tankilevitch, line 38: not used

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -683,6 +695,103 @@ async def get_all_issues(self, group: Group) -> typing.AsyncIterator[List[Issue]
...
+        self, member: RESTObject
+    ) -> dict[str, Any]:
+        user: User = await self.get_user(member.id)
+        member_dict: dict[str, Any] = member.asdict()
+        member_dict["__public_email"] = user.public_email
+        return member_dict
+
+    async def get_user(self, user_id: str) -> User:
+        async with semaphore:
+            logger.info(f"fetching user {user_id}")
```
Tankilevitch, line 778: is there more information that can be added to a `member` other than public email ?

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -683,6 +695,103 @@ async def get_all_issues(self, group: Group) -> typing.AsyncIterator[List[Issue]
...
+            if include_public_email:
+                tasks = [
+                    self.enrich_member_with_public_email(member) for member in members
+                ]
+                members_list.extend(await asyncio.gather(*tasks))
+            else:
+                members_list.extend(member.asdict() for member in members)
+
+        obj_dict: dict[str, Any] = obj.asdict()
+        obj_dict["__members"] = members_list
```
Tankilevitch, line 724: we are not consistent in the way that we are returning the enriched objects, in enrich_project you are returning objects while here you are returning dict, can you elaborate on why?

integrations/gitlab/gitlab_integration/ocean.py
```diff
@@ -154,7 +183,60 @@ async def on_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+                members_tasks = [
+                    service.enrich_object_with_members(
+                        project,
+                        include_inherited_members,
+                        include_bot_members,
+                        include_public_email,
+                    )
+                    for project in projects_enriched_with_extras
+                ]
+                projects_enriched_with_members = await asyncio.gather(*members_tasks)
```
Tankilevitch, line 238: what is making sure we are not experiencing rate limits here?

we get 20 projects, and then spawn 20 tasks and in each project it might have multiple members and then for each member batch you are performing another request.

integrations/gitlab/gitlab_integration/utils.py
```diff
@@ -45,4 +51,7 @@ class ObjectKind:
     PIPELINE = "pipeline"
     PROJECT = "project"
     FOLDER = "folder"
+    MEMBER = "member"
```
Tankilevitch, line 54: doesn't seems used anymore

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -451,11 +462,17 @@ async def get_group(self, group_id: int) -> Group | None:
         else:
             return None
 
-    async def get_all_groups(self) -> typing.AsyncIterator[List[Group]]:
+    @cache_iterator_result()
```
Tankilevitch, line 465: this isn't consistent, on one hand you are using `@cache_iterator_result()` and for `users` you are specifically creating cache. Ideally lets use `cache_iterator_result` if possible

Tankilevitch (changes requested):
Nice job @mk-armah , added a few comments

integrations/gitlab/gitlab_integration/gitlab_service.py
mk-armah: `cache_iterator_results` works with generators, not coroutines...
I wrote one for coroutines [here](https://github.com/port-labs/ocean/blob/a7e4aed2b9a312b145121e6bd4f29783f37e749b/integrations/gcp/gcp_core/cache.py#L9), I'll push to ocean core utils so we use it from there

integrations/gitlab/gitlab_integration/git_integration.py
mk-armah: sorry, this is no longer the case

integrations/gitlab/gitlab_integration/gitlab_service.py
mk-armah: In ProjectWithMembers, we call `enrich_project_with_extras` after which the results are passed down to `enrich_object_with_members`, for `enrich_object_with_members` to get an object, `enrich_project_with_extras ` needed to return an object.

Would you like me to make `enrich_object_with_members` to also manipulate the object directly and return a dict as well ?

integrations/gitlab/gitlab_integration/gitlab_service.py
mk-armah: Here is a sample

```json
{
  "id": 15339857,
  "username": "tompo",
  "name": "Tom Kofi Tankilevitch",
  "state": "active",
  "locked": false,
  "avatar_url": "https://secure.gravatar.com/avatar/81e644d6312fa802fa1a11d605104ba745732a7054be8d3579d3024ff6168bcc?s=80&d=identicon",
  "web_url": "https://gitlab.com/tompo",
  "created_at": "2023-08-27T13:39:29.744Z",
  "bio": "",
  "location": "",
  "public_email": null,
  "skype": "",
  "linkedin": "",
  "twitter": "",
  "discord": "",
  "website_url": "",
  "organization": "",
  "job_title": "",
  "pronouns": null,
  "bot": false,
  "work_information": null,
  "followers": 0,
  "following": 0,
  "is_followed": false,
  "local_time": null
}

```

mk-armah: Member

```json
{
  "id": 15339857,
  "username": "tempo",
  "name": "Tom Kofi Tankilevitch",
  "state": "active",
  "locked": false,
  "avatar_url": "https://secure.gravatar.com/avatar/81e644d6312fa802fa1a11d605104ba745732a7054be8d3579d3024ff6168bcc?s=80&d=identicon",
  "web_url": "https://gitlab.com/tom15339857",
  "access_level": 50,
  "created_at": "2023-08-27T13:39:29.744Z",
  "created_by": {
    "id": 13921399,
    "username": "matanlevi",
    "name": "Matan Levi",
    "state": "active",
    "locked": false,
    "avatar_url": "https://secure.gravatar.com/avatar/814e5edbb657a70c5c9bd09a4171768c245d18cabec5f3b440d43dbc26d36f88?s=80&d=identicon",
    "web_url": "https://gitlab.com/matanlevi"
  },
  "expires_at": null,
  "email": null,
  "membership_state": "active"
}

```

integrations/gitlab/gitlab_integration/gitlab_service.py
mk-armah: [Here](https://github.com/port-labs/ocean/pull/977)

integrations/gitlab/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+0.1.41 (2024-11-13)
```
Tankilevitch, line 10:
```suggestion
0.1.141 (2024-11-13)
```

Tankilevitch (approved):
LGTM

PR #768, opened 2024-07-03
[Integration][Clickup] Add clickup integration

DeeStarks:
# Description

What - This PR adds an integration for Clickup
Why - This will allow users to connect their workspaces from Clickup to Port
How - This was integrated using the Ocean framework. Documentation on usage is available [here](https://github.com/port-labs/port-docs/pull/1366)

## Type of change

## Screenshots

image

image

image

image

## API Documentation

- [Develop an Integration](https://ocean.getport.io/develop-an-integration/)
- [Getting started with Port](https://docs.getport.io/quickstart)
- [Clickup API documentation](https://clickup.com/api/)
- [This integration's documentation](https://github.com/port-labs/port-docs/pull/1366)


integrations/clickup/.dockerignore
```diff
@@ -0,0 +1,94 @@
...
+# Vim swap files
+**/*.swp
+
+# VS Code
+.vscode/
+
+*.md
+**/.ruff_cache
+**/cahangelog
+**/tests
```
PeyGis, line 94: This file needs to be deleted

integrations/clickup/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- towncrier release notes start -->
```
PeyGis, line 8: The changelog content is missing

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+    async def get_single_issue(self, issue_id: str) -> dict[str, Any]:
+        issue_response = await self.client.get(f"{self.clickup_url}/task/{issue_id}")
+        issue_response.raise_for_status()
+        issue = issue_response.json()
+        # converting the datetime timstamps to a datetime object
+        return {
+            **issue,
+            "date_created": timestamp_to_datetime(issue["date_created"]),
+            "date_updated": timestamp_to_datetime(issue["date_updated"]),
+        }
```
PeyGis, line 172: we can move the timestamp to datatime conversion to JQ in the port-app-config

integrations/clickup/clickup/utils.py
```diff
@@ -0,0 +1,10 @@
+from datetime import datetime
+
+
+def timestamp_to_datetime(timestamp: str) -> str | None:
+    """Convert a timestamp to a datetime object."""
+    if not timestamp:
+        return None
+    return datetime.fromtimestamp(int(timestamp) / 1000).strftime(
+        "%Y-%m-%dT%H:%M:%S.%fZ"
+    )
```
PeyGis, line 10: Once we move the conversion from code to JQ, we will not need this file

integrations/clickup/main.py
```diff
@@ -0,0 +1,90 @@
...
+
+    client = ClickupClient(config["clickup_personal_token"])
+    await client.create_events_webhook(app_host)
+
+
+@ocean.on_resync(ObjectKind.TEAM)
+async def on_resync_teams(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    client = ClickupClient(
+        ocean.integration_config["clickup_personal_token"],
+    )
```
PeyGis, line 34: I see you're repeating this over. Maybe you can move it to a function?

integrations/clickup/main.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@ocean.router.post("/webhook")
+async def handle_webhook_request(data: dict[str, Any]) -> dict[str, Any]:
+    client = ClickupClient(ocean.integration_config["clickup_personal_token"])
+
+    event: str = data.get("event", "")
+    logger.info(f"Received clickup webhook event: {event}")
+
+    if "list" in event:
```
PeyGis, line 70: instead of checking if list is in event, can we use the startswith() function?

integrations/clickup/main.py
```diff
@@ -0,0 +1,90 @@
...
+        logger.info(f"Received webhook project event: {data['list_id']}")
+        project = await client.get_single_project(data["list_id"])
+        await ocean.register_raw(ObjectKind.PROJECT, [project])
+    elif "task" in event:
```
PeyGis, line 74: instead of checking if list is in event, can we use the startswith() function?

PeyGis (changes requested):
Left some comments. Please go over them and let me know. Overall, great work

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+    "taskDeleted",
+    "taskCommentPosted",
+    "taskCommentUpdated",
+]
+
+
+class ClickupClient:
+    def __init__(self, clickup_personal_token: str) -> None:
+        self.clickup_url = "https://api.clickup.com/api/v2"
+        self.clickup_personal_token = clickup_personal_token
```
PeyGis, line 25: You can use the clickup_personal_token directly in the header instead of adding it as a class variable

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.api_auth_header = {"Authorization": self.clickup_personal_token}
+        self.client = http_async_client
+        self.client.headers.update(self.api_auth_header)
+
+    async def create_events_webhook(self, app_host: str) -> None:
```
PeyGis, line 31: can rename the function to `create_webhook_events`

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+        teams = self.get_paginated_teams()
+        async for team_batch in teams:
```
PeyGis, line 33: since get_paginated_teams() is an async generator, you could use it directly in the async for loop as

async for team_batch in self.get_paginated_teams():

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+
+    async def get_paginated_teams(
+        self, params: dict[str, Any] = {}
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        teams_response = await self.client.get(
+            f"{self.clickup_url}/team", params=params
+        )
+        teams_response.raise_for_status()
+        teams = teams_response.json()["teams"]
+        yield teams
```
PeyGis, line 82: I don't see where you are handling potential HTTPS errors such as 40X, 50X from the client. Please take a look at other integrations and see how we handle errors and apply it to all your functions.

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+
+    async def _get_paginated_spaces(
+        self, team_id: str, params: dict[str, Any] = {}
+    ) -> list[dict[str, Any]]:
+        spaces_response = await self.client.get(
+            f"{self.clickup_url}/team/{team_id}/space", params=params
+        )
+        spaces_response.raise_for_status()
+        return spaces_response.json()["spaces"]
```
PeyGis, line 91: If we are paginating, I was hoping we could pass some pagination query parameters such as page_size, page_index etc. But I don't see them here. Is this because ClickUp doesn't support pagination? If a user has 10k projects/issues in their workspace, does this get returned all at once? Please investigate and report to us

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+
+    async def get_paginated_projects(
+        self, params: dict[str, Any] = {}
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        # getting all teams so as to retrieve the spaces within each team, then using their
+        # ids to get the projects (lists) within each space
+        teams = self.get_paginated_teams(params)
+        async for team_batch in teams:
```
PeyGis, line 99: this could be updated to

async for team_batch in  self.get_paginated_teams(params):

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+    async def _get_paginated_spaces(
+        self, team_id: str, params: dict[str, Any] = {}
+    ) -> list[dict[str, Any]]:
+        spaces_response = await self.client.get(
+            f"{self.clickup_url}/team/{team_id}/space", params=params
+        )
+        spaces_response.raise_for_status()
+        return spaces_response.json()["spaces"]
+
+    async def get_paginated_projects(
```
PeyGis, line 93: Ocean has helper function to support caching the response from ClickUp so we don't call their API multiple times. Take a look at how we've done it [here](https://github.com/port-labs/ocean/blob/f032aa8b77c1a15cdecfa866719832d7fb167761/integrations/sentry/clients/sentry.py#L89)

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+                    projects_response.raise_for_status()
+                    # because the port-app-config uses the team ID to relate the projects to the teams
+                    # we add the team ID to each project
+                    # also, the clickup returns the datetime as a timestamp, so we convert it to a
+                    # datetime object
+                    projects.extend(
+                        map(
+                            lambda project: {
+                                **project,
+                                "team_id": team["id"],
```
PeyGis, line 117: since team_id is an external variable not from the API response, rename it to `__team_id`

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+                                "start_date": timestamp_to_datetime(
+                                    project["start_date"]
+                                ),
+                                "due_date": timestamp_to_datetime(project["due_date"]),
```
PeyGis, line 121: You can move the timestamp conversion to JQ

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+                            projects_response.json()["lists"],
+                        )
+                    )
+                yield projects
+
+    async def get_single_project(self, project_id: str) -> dict[str, Any]:
+        # Clickup does not provide a direct way to get the team ID from a project (list), so instead
+        # of directly getting the project, we get all the projects and filter by the project ID
+        # using the result of the get_paginated_projects method which includes the team ID
+        projects = self.get_paginated_projects()
```
PeyGis, line 132: Since we're calling self.get_paginated_projects() more than once, we need to add the @cache_iterator_result() decorator on it to cache the response

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,172 @@
...
+                issues = issues_response.json()["tasks"]
+                # converting the datetime timstamps to a datetime object
+                yield [
+                    {
+                        **issue,
+                        "date_created": timestamp_to_datetime(issue["date_created"]),
+                        "date_updated": timestamp_to_datetime(issue["date_updated"]),
+                    }
+                    for issue in issues
+                ]
```
PeyGis, line 161: Move conversion to JQ

PeyGis (changes requested):
I have added more comments. Please let me know if they are not clear or you have an alternative idea

integrations/clickup/clickup/client.py
DeeStarks: I thought that's what the `<Response>.raise_for_status()` is for. I saw it was used in both Linear and Jira

integrations/clickup/clickup/client.py
DeeStarks: Yes, pagination is only supported in tasks based on the documentation, which also doesn't support page size. I can find a way to programmatically handle it

integrations/clickup/clickup/client.py
PeyGis: That's fine then

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,157 @@
...
+
+    async def get_paginated_teams(
+        self, params: dict[str, Any] = {}
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        teams_response = await self.client.get(
+            f"{self.clickup_url}/team", params=params
+        )
+        teams_response.raise_for_status()
+        teams = teams_response.json()["teams"]
+        yield teams
```
PeyGis, line 78: This is not doing any pagination so just switch from AsyncGenerator to a normal list[dict[str, Any]] and use `return` instead of `yield`. Please apply it to all the other methods except the task/issue which is actually doing pagination.

integrations/clickup/main.py
```diff
@@ -0,0 +1,88 @@
...
+            "Without setting up the webhook, the integration will not export live changes from Clickup"
+        )
+        return
+
+    client = create_client()
+    await client.create_webhook_events(app_host)
+
+
+@ocean.on_resync(ObjectKind.TEAM)
+async def on_resync_teams(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
```
PeyGis, line 35: the return type hint has to be `RAW_RESULT` since the get teams is now returning a list of objects. Please apply it to the other resync events except the issues kind

https://github.com/port-labs/ocean/blob/main/port_ocean/core/ocean_types.py

integrations/clickup/main.py
```diff
@@ -0,0 +1,88 @@
...
+    client = create_client()
+
+    async for teams in client.get_paginated_teams():
+        logger.info(f"Received team batch with {len(teams)} teams")
+        yield teams
```
PeyGis, line 40: Also, I would rename all your functions to get_team, get_project etc instead of get_paginated_teams and get_paginated_projects since they are actually not doing any pagination

integrations/clickup/main.py
```diff
@@ -0,0 +1,88 @@
...
+    logger.info(f"Received clickup webhook event: {event}")
+
+    if event.startswith("list"):
+        logger.info(f"Received webhook project event: {data['list_id']}")
+        project = await client.get_single_project(data["list_id"])
+        await ocean.register_raw(ObjectKind.PROJECT, [project])
+    elif event.startswith("task"):
+        logger.info(f"Received webhook task event: {data['task_id']}")
+        issue = await client.get_single_issue(data["task_id"])
+        await ocean.register_raw(ObjectKind.ISSUE, [issue])
```
PeyGis, line 75: In your webhook you have these events

WEBHOOK_EVENTS = [
    "listCreated",
    "listUpdated",
    "listDeleted",
    "taskCreated",
    "taskUpdated",
    "taskDeleted",
    "taskCommentPosted",
    "taskCommentUpdated",
]

At the moment, when a task/list is deleted, it will not be deleted in Port. If you want to have this deleted in Port, use unregister_raw for the delete events

PeyGis (changes requested):
Added more comments

integrations/clickup/clickup/client.py
DeeStarks: done

integrations/clickup/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+          blueprint: '"clickupProject"'
+          identifier: .id | tostring
+          title: .name
+          properties:
+            url: "\"https://app.clickup.com/\" + .team_id + \"/v/l/li/\" + .id"
+            startDate: .start_date | tonumber / 1000 | todate
+            endDate: .due_date | tonumber / 1000 | todate
+            totalIssues: .task_count
+          relations:
+            team: .__team_id
```
Tankilevitch, line 31: too specific, should have added an option to see all the team object, this will allow extensibility in the future

integrations/clickup/.port/resources/port-app-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+    port:
+      entity:
+        mappings:
+          blueprint: '"clickupIssue"'
+          identifier: .id | tostring
+          title: .name
+          properties:
+            url: .url
+            status: .status.status
+            assignee: .assignees | map(.email) | first
```
Tankilevitch, line 44: why not all assignees? having all the list will allow seeing the distribution and spread work inside port between the different assignees, while only setting the first, might be misleading as it can be ordered alphabetically

integrations/clickup/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+
+# Port_Ocean 0.1.0 (2024-07-03)
+
+### Features
+
+- Added Jira integration with support for projects and issues (0.1.0)
```
Tankilevitch, line 14: this is a clickup integration not Jira integration

integrations/clickup/main.py
```diff
@@ -0,0 +1,96 @@
...
+    client = create_client()
+
+    event: str = data.get("event", "")
+    logger.info(f"Received clickup webhook event: {event}")
+
+    if event.startswith("list"):
+        logger.info(f"Received webhook project event: {data['list_id']}")
+        project = await client.get_single_project(data["list_id"])
+
+        if event.endswith("Deleted"):
```
Tankilevitch, line 72: when starting resolving patterns from strings, its worth to leave some comments, examples and reference to how that kind of thing is expected to be received, it can help to onboard developers to the code and make it more readable

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,147 @@
...
+
+            for space in spaces:
+                projects_response = await self.client.get(
+                    f"{self.clickup_url}/space/{space['id']}/list", params=params
+                )
+                projects_response.raise_for_status()
+                # because the port-app-config uses the team ID to relate the projects to the teams
+                # we add the team ID to each project
+                # also, the clickup returns the datetime as a timestamp, so we convert it to a
+                # datetime object
```
Tankilevitch, line 98: we thrive to write a generic code, as although we do have in thought how the mapping will be used, we should be aware that users can change it to their needs and therefor make it generic and as extensible as we can. 

e.g. instead of returning only the team_id, return the full team object.
also no need to mention the datetime and the transformation if you are not doing anything with it here

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,147 @@
...
+        # of directly getting the project, we get all the projects and filter by the project ID
+        # using the result of the get_projects method which includes the team ID
+        project = next(
+            (
+                project
+                for project in await self.get_projects()
+                if project["id"] == project_id
+            ),
+            None,
+        )
```
Tankilevitch, line 121: as although it is nice it re-uses the same code as in the resync method, it is quite an not efficient way to resolve and get the project, as it is very costly to get list all the teams and then all the spaces and then all the projects just to get a single project.

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,147 @@
...
+
+        existing_webhook = next(
+            (
+                webhook
+                for webhook in webhooks["webhooks"]
+                if webhook["endpoint"] == webhook_target_app_host
+            ),
+            None,
+        )
+
```
Tankilevitch, line 45: this whole logic, should be outside of the client, and the client should only expost `get` and `create` methods

Tankilevitch: the resolution logic shouldn't be inside the client

Tankilevitch (changes requested):

integrations/clickup/main.py
```diff
@@ -0,0 +1,96 @@
...
+            await ocean.unregister_raw(ObjectKind.PROJECT, [project])
+        else:
+            await ocean.register_raw(ObjectKind.PROJECT, [project])
+    elif event.startswith("task"):
+        logger.info(f"Received webhook task event: {data['task_id']}")
+        issue = await client.get_single_issue(data["task_id"])
```
Tankilevitch, line 78: what will get returned if its deleted event?

Tankilevitch: I wonder if we even have what to return

integrations/clickup/.port/resources/port-app-config.yaml
DeeStarks: It is a pre-defined blueprint and was advised to not modify. I can adjust it to take in all assignees

integrations/clickup/clickup/client.py
DeeStarks: Yes, that's true. I'll think of a different approach

integrations/clickup/main.py
DeeStarks: Couldn't find a way to solve this. The webhook only returns the task ID, and digging into the `unregister_raw` method, it seems to expect the entire object so it can calculate differences.
Would it be okay to call Port's API from within an integration directly? I'm halfway into this but I realized users will need to generate an API token from the dashboard, and I think they shouldn't have to do that

DeeStarks: For now, I've removed the blocks that check for deleted issues and projects

PR #769, opened 2024-07-03
[Integration][Snyk] - Filter projects

PeyGis:
# Description

What - Based on the new default blueprint model, a project is no longer linked to an issue. But for existing customers, who uses the old model, we need to provide a way for them to still map issues to project. This PR makes it possible for old customers to still bring issues to their project, supporting backwards compatibility
Why -
How -

## Type of change

## Screenshots

## API Documentation


integrations/snyk/main.py
```diff
@@ -71,14 +74,20 @@ async def on_projects_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+        if cast(
+            ProjectResourceConfig, event.resource_config
+        ).selector.attach_issues_to_project:
+            logger.debug("Getting issues for projects in batch")
+            tasks = [process_project_issues(semaphore, project) for project in projects]
+            issues = await asyncio.gather(*tasks)
+            yield [
+                {**project, "__issues": issues}
+                for project, issues in zip(projects, issues)
+            ]
```
Tankilevitch, line 88: lets add a warning when user uses that flag, to indicate that it will be deprecated, as I think this is not too efficient, and also we want users to map the data differently

PR #771, opened 2024-07-03
Add ClickUp integration

martinkatamba:
# Description

This update introduces a seamless integration with ClickUp, allowing you to map teams, projects, and issues directly from your ClickUp environment to your entities in Port.

Key Features:
- Real-Time Sync: Monitor and synchronize changes (create/update/delete) across your ClickUp objects instantly.
- Comprehensive Data Mapping: Fetch and integrate teams, projects, and issues from your ClickUp account into Port effortlessly.
- Continuous Updates: Automatically poll for and apply new updates from your ClickUp account, ensuring your data in Port is always up-to-date.

##Type of Change

[x] New Integration (non-breaking change that adds a new integration)

## Screenshots

Screenshot 2024-07-04 at 1 37 43 AM

Screenshot 2024-07-04 at 1 39 05 AM

Screenshot 2024-07-04 at 1 38 08 AM

Screenshot 2024-07-04 at 1 39 47 AM

Screenshot 2024-07-04 at 1 38 37 AM

Screenshot 2024-07-04 at 1 40 31 AM

## API Documentation

[Port's documentation.](https://docs.getport.io)
[ClickUp API documentation.](https://clickup.com/api/)
[Ocean Framework documentation.](https://ocean.getport.io)


integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
+from typing import Any, AsyncGenerator, List, Dict
+import httpx
+from loguru import logger
+from port_ocean.context import event
+from port_ocean.context.ocean import ocean
+from port_ocean.utils import http_async_client
+
+PAGE_SIZE = 50
```
PeyGis, line 8: Is there a reason why page size is 50? what happens when we bump it to say 100?

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+
+class ClickUpClient:
+    def __init__(self, clickup_url: str, clickup_api_key: str) -> None:
+        """
+        Initializes an instance of the ClickUpClient class.
+
+        Args:
+            clickup_url (str): The base URL for the ClickUp API.
+            clickup_api_key (str): The API key for authenticating requests to the ClickUp API.
+        """
```
PeyGis, line 18: Which this is great, we want to keep the code base clean so you can remove the docs string and only add inline comment where necessary. We believe that function names need to be clear

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+        self.clickup_url = clickup_url
+        self.api_url = f"{self.clickup_url}/api/v2"
+        self.api_key = clickup_api_key
```
PeyGis, line 21: You don't need to store api_key as class variable since you can just use clickup_api_key in the auth header directly

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+    def parse_projects_from_folders(folders: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+        """
+        Extracts and flattens 'Projects' lists from a list of folders.
+
+        Args:
+            folders: A list of dictionaries representing folders with lists.
+
+        Returns:
+            A single list of project dictionaries extracted from folders named "Projects".
+        """
```
PeyGis, line 38: Remove docstring in the other functions too

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+        folder_with_projects = [folder["lists"] for folder in folders]
+        flattened_projects = [project for sublist in folder_with_projects for project in sublist]
+        return flattened_projects
+
+    @staticmethod
+    def parse_project(project: Dict[str, Any], team_id: str) -> Dict[str, Any]:
```
PeyGis, line 44: instead of using Dict and List, let's stick with the default `dict` and `list`

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Args:
+            project (dict): A dictionary representing a project.
+            team_id (str): The team ID to which the project belongs.
+
+        Returns:
+            dict: The updated project dictionary with the team ID.
+        """
+        project["__team_id"] = team_id
+        return project
```
PeyGis, line 56: I don't think you need a function to append the team id to the project you can easily destructure the project dictionary {**project, "__team_id": id}

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+        Args:
+            list_id (str): The ID of the list.
+            params (dict): Query parameters for the API request.
+
+        Returns:
+            dict: JSON response containing the tasks.
+        """
+        response = await self.client.get(f"{self.api_url}/list/{list_id}/task", params=params)
+
+        return response.json() if response else {"tasks": []}
```
PeyGis, line 110: What is the purpose of the `else` part? Why can't you return the json response?

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+        Args:
+            space_id (str): The ID of the space.
+
+        Yields:
+            list[dict[str, Any]]: Lists of project dictionaries.
+        """
+        logger.info("Getting projects from ClickUp")
+
+        params = self._generate_base_req_params()
+        foldered_projects = ClickUpClient.parse_projects_from_folders(
```
PeyGis, line 157: instead of ClickUpClient, why can't you do `self`?

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+        while page * PAGE_SIZE < len(total_projects):
+            logger.info(f"Current query position: page {page}")
+
+            folder_response_list = (await self._get_folders(space_id, params))["folders"]
+            folderless_projects = (await self._get_folderless_projects(space_id, params))["lists"]
+            foldered_projects = ClickUpClient.parse_projects_from_folders(folder_response_list)
+
+            projects = folderless_projects + foldered_projects
+            yield projects
+            page += 1
```
PeyGis, line 177: Clickup does not support pagination for list/folder endpoint so this is not needed

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        logger.info("Getting teams from ClickUp")
+        total_teams = (await self._get_teams())["teams"]
+
+        if not total_teams:
+            logger.warning(
+                "Team query returned 0 teams, did you provide the correct ClickUp API credentials?"
+            )
+
+        yield total_teams
```
PeyGis, line 194: There is no pagination for teams so just self._get_teams() should be sufficient

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+                "Task query returned 0 tasks, did you provide the correct ClickUp API credentials and query parameters?"
+            )
+
+        params["page_size"] = PAGE_SIZE
+        page = 0
+        while page * PAGE_SIZE < len(total_tasks):
+            logger.info(f"Current query position: page {page}")
+            task_response_list = (await self._get_tasks(list_id, params))["tasks"]
+            yield task_response_list
+            page += 1
```
PeyGis, line 254: Without passing the "page" index to the API, this pagination does not work. See the pagination params for [Clickup issues/task](https://clickup.com/api/clickupreference/operation/GetTasks/)

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,254 @@
...
+        Args:
+            list_id (str): The ID of the list.
+
+        Yields:
+            list[dict[str, Any]]: Lists of task dictionaries.
+        """
+        logger.info("Getting tasks (issues) from ClickUp")
+
+        params = self._generate_base_req_params()
+        total_tasks = (await self._get_tasks(list_id, params))["tasks"]
```
PeyGis, line 241: When you get_tasks, Clickup returns the first 100 records. So if you are paginating based on total_task, this will not yield the expected result. Try to have about 120 issues in Clickup and see how this works - to see if you can display all the 120 issues to port

integrations/clickup/main.py
```diff
@@ -0,0 +1,117 @@
...
+    logic_settings = ocean.integration_config
+    app_host = logic_settings.get("app_host")
+
+    if not app_host:
+        logger.warning(
+            "No app host provided, skipping webhook creation. "
+            "Without setting up the webhook, the integration will not export live changes from ClickUp."
+        )
+        return
+    #webhook  setup goes here
```
PeyGis, line 31: If you are not subscribing to webhook then there is no reason to keep this function

integrations/clickup/main.py
```diff
@@ -0,0 +1,117 @@
...
+
+@ocean.on_resync(ObjectKind.TEAM)
+async def on_resync_teams(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    """
+    Resyncs teams from ClickUp.
+
+    Yields:
+        List of team dictionaries.
+    """
```
PeyGis, line 40: Remove the docstrings here too

integrations/clickup/main.py
```diff
@@ -0,0 +1,117 @@
...
+    Resyncs projects from ClickUp.
+
+    Yields:
+        List of project dictionaries.
+    """
+    client = ClickUpClient(
+        ocean.integration_config["clickup_host"],
+        ocean.integration_config["clickup_api_key"],
+    )
```
PeyGis, line 62: You are calling this multiple times in the other resync handlers. I would suggest you wrap this inside a function and call it wherever you need it

integrations/clickup/main.py
```diff
@@ -0,0 +1,117 @@
...
+
+    async for teams in client.get_paginated_teams():
+        for team in teams:
+            team_id = team["id"]
+            async for spaces in client.get_paginated_spaces(team_id):
+                logger.info(f"Received spaces batch with {len(spaces)} spaces for team {team_id}")
+                for space in spaces:
+                    space_id = space["id"]
```
PeyGis, line 70: I would move this block to the client.py so that in this resync handler, I only call the get_paginated_projects() function

integrations/clickup/main.py
```diff
@@ -0,0 +1,117 @@
...
+                    async for projects in client.get_paginated_projects(space_id):
+                        for project in projects:
+                            project_id = project["id"]
```
PeyGis, line 98: Same here. I would move this block to the get_paginated_tasks() function in the client.py so this resync handler is clean - you only call the get_paginated_tasks()

integrations/clickup/main.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Logs the start of the integration and sets up the application if necessary.
+    """
+    logger.info("Starting Port Ocean ClickUp integration")
+
+    if ocean.event_listener_type == "ONCE":
+        logger.info("Skipping webhook creation because the event listener is ONCE")
+        return
+
+    await setup_application()
```
PeyGis, line 117: You don't need this since you are not subscribing to webhooks

integrations/clickup/clickup/client.py
martinkatamba: incase the response is empty the code which uses  this result breaks

integrations/clickup/clickup/client.py
martinkatamba: what do you think about this `        page = 0
        while True:
            params["page"] = page
            logger.info(f"Current query position: page {page}")
            try:
                task_response = await self._get_tasks(list_id, params)
                task_response_list = task_response.get("tasks", [])
                yield task_response_list
                if not task_response_list:
                    break
                logger.error(f"Failed to fetch tasks for page {page}: {e}")
                break
            page += 1`

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,111 @@
...
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        logger.info("Getting spaces from ClickUp")
+
+        try:
+            space_response_list = (await self._get_spaces(team_id))["spaces"]
+            yield space_response_list
+        except httpx.HTTPStatusError as e:
+            logger.error(f"Failed to fetch spaces for team_id: {team_id}: {e}")
+            yield []
+
```
PeyGis, line 89: Since Clickup doesn't support pagination, you don't need to have these async generator methods. Simply use the _get_teams() and _get_spaces()

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,111 @@
...
+            folderless_projects = (await self._get_folderless_projects(space_id))["lists"]
+            projects = folderless_projects + foldered_projects
+
+            if not projects:
+                logger.warning("Space query returned 0 projects")
+
+            yield projects
+        except httpx.HTTPStatusError as e:
+            logger.error(f"Failed to fetch projects for space_id {space_id} : {e}")
+            yield []
```
PeyGis, line 67: I don't see the use of this function since Clickup doesn't support pagination. Just return your foldered + folderless projects

integrations/clickup/clickup/handler.py
```diff
@@ -0,0 +1,65 @@
...
+from clickup.client import ClickUpClient
+from port_ocean.context.ocean import ocean
+
+class ClickUpPortHandler:
+    def __init__(self):
+        self.client = self.get_clickup_client()
+
+    def get_clickup_client(self):
+        try:
+            host = ocean.integration_config["clickup_host"]
```
PeyGis, line 11:
```suggestion
            host = ocean.integration_config.get("clickup_host")
```

PeyGis: This is better since you will add a default value for the host in the spec.yaml

integrations/clickup/clickup/handler.py
```diff
@@ -0,0 +1,65 @@
...
+            raise
+
+    async def fetch_spaces(self, team_id: str):
+        try:
+            async for spaces in self.client.get_spaces(team_id):
+                yield spaces
+        except Exception as e:
+            logger.error(f"Failed to fetch spaces for team {team_id}: {e}")
+            raise
+
```
PeyGis, line 35: this handler file is making the code complex. Just use the functions in client.py directly in main.py. You don't need another layer

PeyGis (changes requested):
More comments

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,139 @@
...
+    def get_clickup_client():
+        try:
+            host = ocean.integration_config.get("clickup_host")
+            api_key = ocean.integration_config.get("clickup_api_key")
+            if not host or not api_key:
+                raise ValueError("ClickUp host or API key is not configured properly.")
+            return ClickUpClient(host, api_key)
+        except KeyError as e:
+            logger.error(f"Configuration key missing: {e}")
+            raise
```
PeyGis, line 26: This is not necessary

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,139 @@
...
+                return projects, team_id
+
+    async def get_teams(self) -> list[dict[str, Any]]:
+        logger.info("Getting teams from ClickUp")
+        try:
+            total_teams = (await self._get_teams())["teams"]
+            return total_teams
+        except httpx.HTTPStatusError as e:
+            logger.error(f"Failed to fetch teams: {e}")
+            return []
```
PeyGis, line 95: What's the point of have a _get_teams() and get_teams() they are both doing the same thing. This is means you don't need this function. Just change the _get_teams, _get_spaces etc directly in the main.py of you need to rename them so they don't have the underscores in front of them

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,139 @@
...
+
+    async def get_spaces(self, team_id: str) -> list[dict[str, Any]]:
+        logger.info("Getting spaces from ClickUp")
+
+        try:
+            space_response_list = (await self._get_spaces(team_id))["spaces"]
+            return space_response_list
+        except httpx.HTTPStatusError as e:
+            logger.error(f"Failed to fetch spaces for team_id: {team_id}: {e}")
+            return []
```
PeyGis, line 105: This is redundant

integrations/clickup/clickup/client.py
```diff
@@ -0,0 +1,139 @@
...
+        projects, _ = await self.fetch_all_projects()
+        for project in projects:
+            project_id = project["id"]
+            try:
+                async for tasks in self.get_paginated_tasks(project_id):
+                    logger.info(f"Received task batch with {len(tasks)} tasks for project {project_id}")
+                    yield tasks
+            except Exception as e:
+                logger.error(f"Failed to fetch tasks for project {project_id}: {e}")
+                raise
```
PeyGis, line 139: Redundant

PeyGis (changes requested):
More comments

PR #790, opened 2024-07-08
[Core] Added environment variables access restriction option

matan84:
# Description

What - Added a new ocean configuration option- Allow Environment variables jq access, which is true by default
Why - To have an option to hide sensitive information from the integration runner
How - Added Integration configuration

## Type of change


Tankilevitch (approved):
Clean 🌊

PR #804, opened 2024-07-10
PORT-5598 | Jira integration - support sprints | Change priority to name

lordsarcastic:
# Description

- Added support for sprints
- Change Issue kind `priority` field to use the priority name instead of the priority id
- Added "total issues" field to project kind

## Type of change
image

More information can be found in the [docs PR](https://pr-1262.d2ngvl90zqbob8.amplifyapp.com/build-your-software-catalog/sync-data-to-catalog/project-management/jira/).


integrations/jira/client.py
```diff
@@ -0,0 +1,242 @@
...
+        self.base_url = f"{self.jira_url}/rest/agile/1.0"
+        self.jira_rest_url = f"{self.jira_url}/rest"
+        self.detail_base_url = f"{self.jira_rest_url}/api/3"
+
+        self.jira_api_auth = BasicAuth(jira_email, jira_token)
+        self.webhooks_url = f"{self.jira_rest_url}/webhooks/1.0/webhook"
+
+        self.client = http_async_client
+        self.client.auth = self.jira_api_auth
+        self.client.timeout = Timeout(30)
```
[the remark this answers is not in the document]
Tankilevitch: I would also increase the timeout, as this is not something that we can somehow tune and we are relying on the jira performance, and we don't want to fail the sync due to response taking more than 30 seconds. Lets increase it to 60.

integrations/jira/main.py
```diff
@@ -75,6 +88,14 @@ async def handle_webhook_request(data: dict[str, Any]) -> dict[str, Any]:
         logger.info(f'Received webhook event for issue: {data["issue"]["key"]}')
         issue = await client.get_single_issue(data["issue"]["key"])
         await ocean.register_raw(ObjectKind.ISSUE, [issue])
+    elif "board" in data:
+        logger.info(f'Received webhook event for board: {data["board"]["id"]}')
+        board = await client.get_single_board(data["board"]["id"])
```
Tankilevitch, line 93: is the code actually handles delete events? as it only does `register_raw`

integrations/jira/client.py
```diff
@@ -0,0 +1,242 @@
...
+        config = typing.cast(JiraIssueResourceConfig, event.resource_config)
+        params = {}
+        if config.selector.jql:
+            params["jql"] = config.selector.jql
+            logger.info(f"Found JQL filter: {config.selector.jql}")
+
+        ISSUES_MAP = {
+            "board": self._get_issues_from_board,
+            "sprint": self._get_issues_from_sprint,
+            "all": self._get_issues_from_org,
```
Tankilevitch, line 164: the client shouldn't be familiar with the event.resource_config, it should be as lean as possible and just expose an interface on how to interact with the third party, therefor let move the extraction of `event.resource_config` outside of the client.

Tankilevitch (changes requested):

integrations/jira/client.py
```diff
@@ -0,0 +1,242 @@
...
+        logger.info("Finished paginated request")
+        return
+
+    async def get_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        async for projects in self._make_paginated_request(
+            f"{self.detail_base_url}/project/search"
+        ):
+            yield projects["values"]
+
+    async def _get_issues_from_board(
```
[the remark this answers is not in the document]
lordsarcastic: If there are duplicated issues, Port will see them as one and we'll end up with only entity in the user dashboard. It is idempotent.

Tankilevitch (changes requested):

integrations/jira/integration.py
```diff
@@ -1,7 +1,56 @@
...
+SprintState = Literal["active", "closed", "future"]
+
+
+class JiraIssueSelector(Selector):
+    jql: str | None = Field(
+        description="Jira Query Language (JQL) query to filter issues",
+    )
+    source: Literal["board", "sprint", "all"] = Field(
+        default="sprint",
+        description="Where issues are sourced from",
```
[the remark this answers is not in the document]
lordsarcastic: If "board" is chosen, only issues that appear in a board that will be ingested. If "sprint", only issues from sprints will be ingested. If "all", all the issues in the Jira project will be ingested.

integrations/jira/integration.py
```diff
@@ -1,7 +1,56 @@
...
+    )
+    # when resyncing issues, there is no way to retrieve the config
+    # set for the `sprint` kind, so we need to duplicate the state
+    # field. This is redundant, but necessary.
+    state: SprintState = Field(
+        alias="sprintState",
+        default="active",
+        description="State of the sprint",
+    )
```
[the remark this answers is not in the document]
lordsarcastic: This is useful when the `source` is set to "sprint". The user can decide what type of sprints the issues will be sourced from. They can decide to only source from the currently active sprint, or from closed sprints, or all sprints

mk-armah:
Can you add screenshots to the description ?

integrations/jira/integration.py
```diff
@@ -1,7 +1,56 @@
...
+SprintState = Literal["active", "closed", "future"]
+
+
+class JiraIssueSelector(Selector):
+    jql: str | None = Field(
+        description="Jira Query Language (JQL) query to filter issues",
+    )
+    source: Literal["sprint", "all"] = Field(
+        default="sprint",
+        description="Where issues are sourced from",
```
Tankilevitch, line 21: lets make default: all, to not break the expected behavior.

Tankilevitch (changes requested):

Tankilevitch (changes requested):
Have you tried this - https://community.developer.atlassian.com/t/jira-api-v3-include-sprint-in-get-issue-search/35411/3 ? I think we should try and allow customers to pass extra fields that they want to get on issues.

integrations/jira/.port/resources/blueprints.json
```diff
@@ -103,6 +110,13 @@
       }
     },
     "relations": {
+      "sprint": {
+        "target": "jiraSprint",
+        "title": "Sprint",
+        "description": "The Jira sprint that contains this issue",
+        "required": false,
+        "many": false
+      },
```
PeyGis, line 119: I see we have a relation to the sprint but in the config, I don't see where we do the mapping

integrations/jira/CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
+## 0.1.93 (2024-10-10)
+
+
+### Features
+
+- Added support for sprints and ingesting issues by sprints (0.1.93)
```
PeyGis, line 15: This needs to change since we are removing sprints

integrations/jira/client.py
```diff
@@ -0,0 +1,154 @@
...
+                    f" and response text: {e.response.text}"
+                )
+                raise
+            except httpx.HTTPError as e:
+                logger.error(f"HTTP occurred while fetching Jira data {e}")
+                raise
+        logger.info("Finished paginated request")
+        return
+
+    async def get_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
```
PeyGis, line 94: the issues method is named get_all_issues() can we rename this to get_all_projects() too?

PeyGis (changes requested):
left some comments

PeyGis (approved):
LGTM

erikzaadi (commented):
Looks great, is there a way to add tests to verify this change?

lordsarcastic:
> Looks great, is there a way to add tests to verify this change?

Nothing's impossible. I'll write some tests for it then

integrations/jira/.port/resources/blueprints.json
```diff
@@ -11,10 +11,17 @@
...
         }
       }
     },
-    "calculationProperties": {}
+    "mirrorProperties": {},
+    "calculationProperties": {},
+    "relations": {}
   },
   {
     "identifier": "jiraIssue",
```
Tankilevitch, line 27: I am thinking maybe we should add a property with sprint, where the description of the property will point the user to the docs?

integrations/jira/CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+
+### Features
+
+- Added a field to display total issues in a project (0.1.96)
+- Added support for ingesting other fields apart from the default fields (Jira Sprint support) (0.1.96)
+
+### Improvements
+
+- Changed issue priority from id to name (0.1.96)
```
Tankilevitch, line 20:
```suggestion
- Added a field to display total issues in a project
- Added support for ingesting other fields apart from the default fields (Jira Sprint support)

### Improvements

- Changed issue priority from id to name
```

integrations/jira/CHANGELOG.md
```diff
@@ -20,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improvements
 
-- Bumped ocean version to ^0.12.6
+- Bumped ocean version to ^0.12.6 (0.1.94)
```
Tankilevitch, line 36:
```suggestion
- Bumped ocean version to ^0.12.6
```

integrations/jira/tests/test_sync.py
Tankilevitch, line 1: test should include actual test of expected data and enrichments.
e.g. 
testing that when passing a `jql` then its adds the `jql` to the query. When adding `fields` that those `fields` are being added to the entity and that you see the expected value.

integrations/jira/main.py
```diff
@@ -35,46 +43,61 @@ async def setup_application() -> None:
...
+        await ocean_action(ObjectKind.PROJECT, [project])
+    elif "issue" in webhook_event:
         logger.info(f'Received webhook event for issue: {data["issue"]["key"]}')
-        issue = await client.get_single_issue(data["issue"]["key"])
-        await ocean.register_raw(ObjectKind.ISSUE, [issue])
+        if delete_action:
+            issue = data["issue"]
+        else:
+            issue = await client.get_single_issue(data["issue"]["key"])
+        await ocean_action(ObjectKind.ISSUE, [issue])
```
Tankilevitch, line 100: I think we should perform a request to get the issue with the `fields` as well, so we don't override it with null once we receive an event. 
Maybe `JQL` as well? as we might receive an event of an issue that we are not interested which in regular `resync` wouldn't been inserted, but in real time event it would've.

integrations/jira/.port/resources/blueprints.json
lordsarcastic: The docs already contains how to add the sprint property. You think we should add it by default?

integrations/jira/main.py
lordsarcastic: All fields are returned by default, so we shouldn't have null overwrites

PR #835, opened 2024-07-23
[Core] ocean core next resync

shalev007:
# Description

**What:**
- Integrated functionality to send the state of the integration on each sync.
- Implemented prediction of the next sync date.

**Why:**
- To ensure that the integration state is consistently updated and monitored.
- To enhance the accuracy and reliability of the next sync date prediction by utilizing server data for SaaS applications, rather than relying solely on environment variables.

**How:**
- Updated the sync logic to include the current integration state in the payload sent to our monitoring system.
- Modified the sync prediction mechanism for SaaS applications to use data from our servers, providing more accurate and context-aware predictions.

## Type of change


port_ocean/ocean.py
```diff
@@ -68,6 +71,26 @@ def __init__(
...
         )
 
+    async def calculate_next_resync(self, now: datetime.datetime) -> float | None:
+        if self.config.runtime != "Saas" and not self.config.scheduled_resync_interval:
+            # There is no scheduled resync outside of Saas runtime or if not configured
+            return None
+
+        interval = self.config.scheduled_resync_interval
+        next_resync = None
+        if self.config.runtime == "Saas":
```
Tankilevitch, line 81: make `Saas` consts

port_ocean/ocean.py
```diff
@@ -68,6 +71,26 @@ def __init__(
...
+            integration = await self.port_client.get_current_integration()
+            interval_str = (
+                integration.get("spec", {})
+                .get("appSpec", {})
+                .get("scheduledResyncInterval")
+            )
```
Tankilevitch, line 87: I think we already have a code that turns the integration to an instance of integration, that way we don't need to actually use dictionary to get that key

port_ocean/ocean.py
```diff
@@ -68,6 +71,26 @@ def __init__(
...
+            interval = convert_time_to_minutes(interval_str)
+
+        next_resync_date = now + datetime.timedelta(minutes=float(interval or 0))
+        next_resync = next_resync_date.now(datetime.timezone.utc).timestamp()
```
Tankilevitch, line 91: lets leave it in seconds and make the UI show it in minutes if we want

port_ocean/ocean.py
```diff
@@ -98,8 +121,16 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+                now = datetime.datetime.now()
+                calculation = asyncio.create_task(self.calculate_next_resync(now))
                 await self.integration.start()
                 await self._setup_scheduled_resync()
+                next_resync = await calculation
+                await self.port_client.update_resync_state(
+                    {
+                        "next_resync": next_resync,
+                    }
+                )
```
Tankilevitch, line 133: I think this needs be updated right after the resync, instead of as part of the initialization itself, that way we are more consistent and provide actual data.

Tankilevitch (changes requested):
CI doesn't and left some more comments

port_ocean/ocean.py
shalev007: WDYM?
I changed it to `timestamp`  to have better control over how we show it, and changed it to UTC so we won't have any timezone issues

port_ocean/ocean.py
shalev007: I think your referring to `ocean.config.integration` but it only contains data from the env variables config and integration identifier

port_ocean/core/event_listener/once.py
```diff
@@ -41,6 +45,34 @@ def __init__(
...
+            self.resync_state["next_resync"] = calculate_next_resync(now, interval)
+        except Exception:
+            logger.exception("Error occurred while calculating next resync")
+            return None
+
+    async def after_resync(self) -> None:
+        if not self.should_update_resync_state():
+            return None
+
+        await ocean.port_client.update_resync_state(self.resync_state)
```
yairsimantov20, line 75: I want this implemented a bit different 
With a lifecycle context manager

port_ocean/utils/misc.py
```diff
@@ -66,3 +67,29 @@ def load_module(file_path: str) -> ModuleType:
...
+    :param time_str: a string representing time in the format "1h" or "1m"
+    """
+    if time_str.endswith("h"):
+        hours = int(time_str[:-1])
+        return hours * 60
+    elif time_str.endswith("m"):
+        minutes = int(time_str[:-1])
+        return minutes
+    else:
+        raise ValueError("Invalid format. Expected a string ending with 'h' or 'm'.")
```
yairsimantov20, line 84: Why? Just send the time to port and let it handle it

port_ocean/utils/misc.py
```diff
@@ -66,3 +67,29 @@ def load_module(file_path: str) -> ModuleType:
...
+
+
+def calculate_next_resync(
+    now: datetime.datetime, interval: int | None = None
+) -> float | None:
+    if not interval:
+        return None
+
+    next_resync_date = now + datetime.timedelta(minutes=float(interval or 0))
```
yairsimantov20, line 93: Why float?
```suggestion
    next_resync_date = now + datetime.timedelta(minutes=float(interval))
```

port_ocean/ocean.py
```diff
@@ -88,7 +97,11 @@ def execute_resync_all() -> None:
                 seconds=interval * 60,
                 # Not running the resync immediately because the event listener should run resync on startup
                 wait_first=True,
-            )(lambda: threading.Thread(target=execute_resync_all).start())
+            )(
+                lambda: threading.Thread(
+                    target=asyncio.run(execute_resync_all())
+                ).start()
+            )
```
yairsimantov20, line 104: Try to implement with asyncio task instead

port_ocean/core/models.py
```diff
@@ -1,10 +1,16 @@
...
 from pydantic import BaseModel
 from pydantic.fields import Field
 
 
+class RuntimeType(Enum):
+    Saas = "Saas"
+    OnPrem = "OnPrem"
+
+
 Runtime = Literal["OnPrem", "Saas"]
```
yairsimantov20, line 14: Why is that needed anymore?

port_ocean/core/event_listener/polling.py
```diff
@@ -83,6 +106,8 @@ async def resync() -> None:
                 signal_handler.register(running_task.cancel)
 
                 await running_task
+                await self.after_resync()
+                self.is_first_resync = False
```
yairsimantov20, line 110: Why is this needed?

yairsimantov20 (changes requested):
Lets talk about the implementation

port_ocean/ocean.py
shalev007: But tasks needs to be awaited which is not compatible with the target type `Callable`

port_ocean/utils/misc.py
shalev007: it's inconsistent with the way we send out time in non-saas integrations

port_ocean/core/event_listener/polling.py
shalev007: This is only used for the resync state update
since polling listener is running a resync automatically only on installation after that all triggers are manual.

port_ocean/core/event_listener/once.py
shalev007: Seems like a great idea!
but I think it's out of this PR's scope and it will inflate this PR more than it should, WDYT?

port_ocean/core/models.py
shalev007: great point

port_ocean/utils/misc.py
```diff
@@ -66,3 +67,29 @@ def load_module(file_path: str) -> ModuleType:
     spec.loader.exec_module(module)
 
     return module
+
+
+def convert_time_to_minutes(time_str: str) -> int:
```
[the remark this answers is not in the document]
shalev007: port_ocean/core/event_listener/once.py::65

port_ocean/core/event_listener/kafka.py
```diff
@@ -122,7 +122,9 @@ async def _handle_message(self, raw_msg: Message) -> None:
 
         if "change.log" in topic and message is not None:
             try:
+                await self._before_resync()
                 await self.events["on_resync"](message)
+                await self._after_resync()
```
Tankilevitch, line 127: I think we can add the `resync` method to the `baseEventListener` that will contain this code that way you won't need to repeat it for each event listener, as well as making sure that every new event listener added will get it out of the box

Tankilevitch (changes requested):

port_ocean/clients/port/client.py
```diff
@@ -75,3 +76,13 @@ async def get_org_id(self) -> str:
...
+
+    async def update_integration_state(self, state: dict[str, Any]) -> dict[str, Any]:
+        logger.debug(f"Updating integration state with: {state}")
+        response = await self.client.patch(
+            f"{self.api_url}/integration/{self.integration_identifier}",
+            headers=await self.auth.headers(),
+            json={"state": state},
+        )
+        handle_status_code(response)
+        return response.json().get("integration", {})
```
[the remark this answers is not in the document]
shalev007: is this a blocker?
is it mandatory for this PR, because if we do have time to invest in this infra then I'd rather use a whole other infra set to save this data in

port_ocean/core/event_listener/polling.py
```diff
@@ -49,7 +49,20 @@ def __init__(
...
         super().__init__(events)
         self.event_listener_config = event_listener_config
-        self._last_updated_at = None
+
+    async def _before_resync(self) -> None:
+        await ocean.app.update_state_before_scheduled_sync()
+
+    async def _after_resync(self) -> None:
+        await ocean.app.update_state_after_scheduled_sync()
+
```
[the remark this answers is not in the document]
shalev007: forgot to delete it after moving it to `BaseEventListener`, nice catch!

port_ocean/core/event_listener/once.py
```diff
@@ -41,6 +44,71 @@ def __init__(
...
+
+        return (
+            convert_time_to_minutes(interval_str),
+            convert_str_to_datetime(start_time_str),
+        )
+
+    async def _before_resync(self) -> None:
+        if not ocean.app.is_saas():
+            await super()._before_resync()
+            return
```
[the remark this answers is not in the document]
shalev007: Then we'd end up updating the integration state twice, once in `BaseEventListener` and then another in `once` event listener

port_ocean/core/event_listener/once.py
```diff
@@ -41,6 +44,71 @@ def __init__(
...
+
+        try:
+            self.cached_integration = await ocean.port_client.get_current_integration()
+            return self.cached_integration
+        except Exception:
+            logger.exception("Error occurred while getting current integration")
+            return None
+
+    async def get_saas_integration_prediction_data(
+        self,
```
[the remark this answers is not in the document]
shalev007: this function returns the necessary data for saas integrations to make the next_resync prediction, I don't love this name myself TBH, what would you call it?

port_ocean/core/event_listener/once.py
```diff
@@ -41,6 +44,71 @@ def __init__(
...
+    ) -> tuple[int | None, datetime.datetime | None]:
+        if not ocean.app.is_saas():
+            return (None, None)
+
+        integration = await self.get_current_integration_cached()
+        if not integration:
+            return (None, None)
+
+        integration = await ocean.port_client.get_current_integration()
+        interval_str = (
```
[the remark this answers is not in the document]
shalev007: This is of course a bug

port_ocean/ocean.py
```diff
@@ -62,17 +65,85 @@ def __init__(
...
-    async def _setup_scheduled_resync(
+        # TODO: remove this once we separate the state from the integration
+        self.last_resync_start: datetime.datetime | None = None
+        self.last_integration_updated_at: str = ""
+
+    def get_last_updated_at(self) -> str | None:
+        return self.last_integration_updated_at
+
+    def set_last_updated_at(self, last_updated_at: str) -> None:
+        self.last_integration_updated_at = last_updated_at
```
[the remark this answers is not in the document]
shalev007: removing them and then changing the `Ocean` property directly?

port_ocean/core/event_listener/kafka.py
shalev007: Is it a blocker?

port_ocean/core/event_listener/polling.py
```diff
@@ -69,20 +82,17 @@ async def resync() -> None:
...
                 logger.info("Detected change in integration, resyncing")
-                self._last_updated_at = last_updated_at
+                ocean.app.set_last_updated_at(last_updated_at)
                 running_task: Task[Any] = get_event_loop().create_task(
                     self.events["on_resync"]({})  # type: ignore
                 )
                 signal_handler.register(running_task.cancel)
 
                 await running_task
+                await self._after_resync()
```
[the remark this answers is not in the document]
shalev007: Yes it will not stay up and running, I think whatever error that will come up to this it will not exit the app.
But we need to take into account that if this request fails in polling event-listener we might run an un called for resync since the latest `updatedAt` will not inline with the current one

port_ocean/core/event_listener/once.py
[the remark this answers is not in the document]
shalev007: delete and comment

port_ocean/core/event_listener/once.py
```diff
@@ -41,6 +44,71 @@ def __init__(
...
+                "Integration scheduled resync interval not found for integration state update"
+            )
+            return (None, None)
+
+        start_time_str = integration.get("statusInfo", {}).get("updatedAt")
+
+        if not start_time_str:
+            logger.error(
+                "Integration creation time not found for integration state update"
+            )
```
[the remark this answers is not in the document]
shalev007: add comment

port_ocean/utils/misc.py
```diff
@@ -66,3 +68,59 @@ def load_module(file_path: str) -> ModuleType:
...
+    Convert a string representing time to a datetime object.
+    :param time_str: a string representing time in the format "2021-09-01T12:00:00Z"
+    """
+    if time_str.endswith("Z"):
+        aware_date = datetime.datetime.fromisoformat(time_str.replace("Z", "+00:00"))
+    aware_date = datetime.datetime.fromisoformat(time_str)
+    return datetime.datetime.fromtimestamp(aware_date.timestamp())
+
+
+def convert_time_to_minutes(time_str: str) -> int:
```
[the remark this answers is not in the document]
shalev007: moved them to time

port_ocean/ocean.py
```diff
@@ -62,17 +65,79 @@ def __init__(
...
+    ) -> None:
+        _interval = interval or self.config.scheduled_resync_interval
+
+        self.last_resync_start = datetime.datetime.now()
+        integration = await self.port_client.update_integration_state(
+            {
+                "status": "running",
+                "last_resync_start": self.last_resync_start.timestamp(),
+                "last_resync_end": None,
+                "interval_in minuets": _interval,
```
Tankilevitch, line 101:
```suggestion
                "interval_in_minuets": _interval,
```

port_ocean/ocean.py
```diff
@@ -62,17 +65,79 @@ def __init__(
...
+        integration = await self.port_client.update_integration_state(
+            {
+                "status": "completed",
+                "last_resync_start": (
+                    self.last_resync_start.timestamp()
+                    if self.last_resync_start
+                    else None
+                ),
+                "last_resync_end": datetime.datetime.now().timestamp(),
+                "interval_in minuets": _interval,
```
Tankilevitch, line 125:
```suggestion
                "interval_in_minuets": _interval,
```

port_ocean/core/event_listener/base.py
```diff
@@ -36,6 +37,29 @@ def _stop(self) -> None:
...
+    async def _resync(
+        self,
+        resync_args: dict[Any, Any],
+    ) -> None:
+        """
+        Triggers the "on_resync" event.
+        """
+        await self._before_resync()
+        await self.events["on_resync"](resync_args)
+        await self._after_resync()
```
Tankilevitch, line 61: I know its not in much in the scope, but I am a bit afraid of the integration leaving wrong "last" state.
```suggestion
    async def _resync(
        self,
        resync_args: dict[Any, Any],
    ) -> None:
        """
        Triggers the "on_resync" event.
        """
        await self._before_resync()
        try:
            await self.events["on_resync"](resync_args)
        except Exception:
            await self._on_resync_failure()
        else:
            await self._after_resync()
```

port_ocean/core/event_listener/once.py
```diff
@@ -41,6 +44,68 @@ def __init__(
...
+        self.cached_integration: dict[str, Any] | None = None
+
+    async def get_current_integration_cached(self) -> dict[str, Any]:
+        if self.cached_integration:
+            return self.cached_integration
+
+        self.cached_integration = await ocean.port_client.get_current_integration()
+        return self.cached_integration
+
+    async def get_saas_integration_prediction_data(
```
Tankilevitch, line 56:
```suggestion

    async def get_saas_resync_initialization_and_interval(
```

Tankilevitch (changes requested):

port_ocean/clients/port/client.py
```diff
@@ -75,3 +76,16 @@ async def get_org_id(self) -> str:
...
+    async def update_integration_state(self, state: dict[str, Any]) -> dict[str, Any]:
+        logger.debug(f"Updating integration state with: {state}")
+        response = await self.client.patch(
+            f"{self.api_url}/integration/{self.integration_identifier}/state",
+            headers=await self.auth.headers(),
+            json=state,
+        )
+        handle_status_code(response, should_raise=False, should_log=True)
+        if not response.is_error:
+            logger.info("Integration state updated successfully")
```
Tankilevitch, line 89: and use the `should_log` here as well

port_ocean/core/event_listener/once.py
```diff
@@ -41,6 +44,93 @@ def __init__(
...
+            return (None, None)
+
+        last_updated_saas_integration_config_str = integration.get(
+            "statusInfo", {}
+        ).get("updatedAt")
+
+        # we use the last updated time of the integration config as the start time since in saas application the interval is configured by the user from the portal
+        if not last_updated_saas_integration_config_str:
+            logger.error("updatedAt not found for integration")
+            return (None, None)
```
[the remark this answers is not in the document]
Tankilevitch: for example, more detailed log for whether this was expected or not

Tankilevitch (changes requested):

port_ocean/clients/port/client.py
```diff
@@ -75,3 +76,16 @@ async def get_org_id(self) -> str:
         handle_status_code(response)
 
         return response.json()["organization"]["id"]
+
+    async def update_integration_state(self, state: dict[str, Any]) -> dict[str, Any]:
+        logger.debug(f"Updating integration state with: {state}")
```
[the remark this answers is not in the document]
shalev007: Why?
We don't need this to be changed, these parameters stay constant across event-listener

port_ocean/clients/port/client.py
shalev007: WDYM?

port_ocean/core/event_listener/once.py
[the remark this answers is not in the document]
shalev007: Added

port_ocean/ocean.py
```diff
@@ -62,17 +65,84 @@ def __init__(
         self.integration = (
             integration_class(ocean) if integration_class else BaseIntegration(ocean)
         )
+        self.initiated_at = datetime.datetime.now(tz=datetime.timezone.utc)
 
-    async def _setup_scheduled_resync(
+        # TODO: remove this once we separate the state from the integration
+        self.last_integration_updated_at: str = ""
```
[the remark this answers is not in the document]
shalev007: Added a comment

port_ocean/clients/port/client.py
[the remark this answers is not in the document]
Tankilevitch: this is code in the client class, therefor it should be an interface for other uses as well. and therefor each call to the client should decide whether it wants to raise the error or not

port_ocean/clients/port/client.py
Tankilevitch: once you add the `should_raise` and `should_log` then decide if to log based on the parameter passed

Tankilevitch (approved):
LGTM

port_ocean/clients/port/client.py
```diff
@@ -75,3 +76,19 @@ async def get_org_id(self) -> str:
...
+        response = await self.client.patch(
+            f"{self.api_url}/integration/{self.integration_identifier}/state",
+            headers=await self.auth.headers(),
+            json=state,
+        )
+        handle_status_code(response, should_raise, should_log)
+        if response.is_success and should_log:
+            logger.info("Integration state updated successfully")
+
+        return response.json().get("integration", {})
```
yairsimantov20, line 94:
```suggestion
        return response.json()["integration"]
```

port_ocean/core/event_listener/base.py
```diff
@@ -36,6 +38,39 @@ def _stop(self) -> None:
...
+        """
+        Triggers the "on_resync" event.
+        """
+        await self._before_resync()
+        try:
+            await self.events["on_resync"](resync_args)
+            await self._after_resync()
+        except Exception as e:
+            await self._on_resync_failure(e)
+            raise e
```
yairsimantov20, line 72: can we instead use a context manager syntax?

```
@asynccontextmanager
async resync_life_cycle():
    await self._before_resync()
    try:
        yield
    except Exception as e:
            await self._on_resync_failure(e)
            raise e
```

```suggestion
    async def _before_resync(self) -> None:
        """
        Can be used for event listeners that need to perform some action before resync.
        """
        await ocean.app.update_state_before_scheduled_sync()

    async def _after_resync(self) -> None:
        """
        Can be used for event listeners that need to perform some action after resync.
        """
        await ocean.app.update_state_after_scheduled_sync()

    async def _on_resync_failure(self, e: Exception) -> None:
        """
        Can be used for event listeners that need to handle resync failures.
        """
        await ocean.app.update_state_after_scheduled_sync(IntegrationStateStatus.Failed)

    async def _resync(
        self,
        resync_args: dict[Any, Any],
    ) -> None:
        """
        Triggers the "on_resync" event.
        """
        async with resync_life_cycle():
            await self.events["on_resync"](resync_args)
```

port_ocean/utils/time.py
```diff
@@ -0,0 +1,54 @@
...
+
+def convert_str_to_utc_datetime(time_str: str) -> datetime.datetime | None:
+    """
+    Convert a string representing time to a datetime object.
+    :param time_str: a string representing time in the format "2021-09-01T12:00:00Z"
+    """
+    aware_date = datetime.datetime.fromisoformat(time_str)
+    if time_str.endswith("Z"):
+        aware_date = datetime.datetime.fromisoformat(time_str.replace("Z", "+00:00"))
+    return aware_date.astimezone(datetime.timezone.utc)
```
yairsimantov20, line 13:
```suggestion
    return datetime.strptime(date_str, "%Y-%m-%dT%H:%M:%SZ").astimezone(datetime.timezone.utc)
```

port_ocean/utils/misc.py
```diff
@@ -1,15 +1,21 @@
...
 from uuid import uuid4
-
 import tomli
 import yaml
 
 
+class IntegrationStateStatus(Enum):
+    Running = "running"
+    Failed = "failed"
+    Completed = "completed"
```
yairsimantov20, line 16: any reason those are lower case?

port_ocean/ocean.py
```diff
@@ -62,17 +66,93 @@ def __init__(
...
-            loop.run_until_complete(self.integration.sync_raw_all())
-            loop.close()
+            try:
+                await self.integration.sync_raw_all()
+                await self.update_state_after_scheduled_sync()
+            except Exception as e:
+                await self.update_state_after_scheduled_sync(
+                    IntegrationStateStatus.Failed
+                )
+                raise e
```
yairsimantov20, line 155: This logic does not belong in this class
it is irrelevant for the core
this feature is only a side effect to the resync no the main core feature

yairsimantov20 (dismissed):
A lot of code here is irrelevant to ocean core and can be calculated on the port side

port_ocean/clients/port/client.py
```diff
@@ -75,3 +76,19 @@ async def get_org_id(self) -> str:
...
+        if should_log:
+            logger.debug(f"Updating integration state with: {state}")
+        response = await self.client.patch(
+            f"{self.api_url}/integration/{self.integration_identifier}/state",
+            headers=await self.auth.headers(),
+            json=state,
+        )
+        handle_status_code(response, should_raise, should_log)
+        if response.is_success and should_log:
+            logger.info("Integration state updated successfully")
```
Tankilevitch, line 92:
```suggestion
    async def update_integration_resync_state(
        self, state: dict[str, Any], should_raise: bool = True, should_log: bool = True
    ) -> dict[str, Any]:
        if should_log:
            logger.debug(f"Updating integration state with: {state}")
        response = await self.client.patch(
            f"{self.api_url}/integration/{self.integration_identifier}/resync-state",
            headers=await self.auth.headers(),
            json=state,
        )
        handle_status_code(response, should_raise, should_log)
        if response.is_success and should_log:
            logger.info("Integration resync state updated successfully")
```

PR #841, opened 2024-07-25
[AzureDevops] Added check only for project id when comparing publisher inputs

matan84:
# Description

What - Fixed webhook creation bug where each restart would create all webhooks again
Why - Bug
How - Webhook comparing publisherInputs would always fail because ADO pushes additional input

## Type of change

### DIFF
--- a/integrations/azure-devops/azure_devops/webhooks/webhook_event.py
+++ b/integrations/azure-devops/azure_devops/webhooks/webhook_event.py
@@ -29,7 +29,7 @@ class WebhookEvent(BaseModel):
                 if not self.publisherInputs and not subscribed_event.publisherInputs:
                     return subscribed_event
 
-                # Azure Devops sends more than just the projectId in the publisherInputs, 
+                # Azure Devops sends more than just the projectId in the publisherInputs,
                 # And we only need to verify the projectId
                 if self.publisherInputs and subscribed_event.publisherInputs:
                     if subscribed_event.publisherInputs.get(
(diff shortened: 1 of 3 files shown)


phalbert (approved):
Tested locally, and LGTM

integrations/azure-devops/azure_devops/webhooks/webhook_event.py
```diff
@@ -25,9 +25,15 @@ def get_event_by_subscription(
...
             ):
-                return subscribed_event
+                if not self.publisherInputs and not subscribed_event.publisherInputs:
+                    return subscribed_event
+
+                if self.publisherInputs and subscribed_event.publisherInputs:
+                    if subscribed_event.publisherInputs.get(
+                        "projectId", None
+                    ) == self.publisherInputs.get("projectId", None):
+                        return subscribed_event
```
Tankilevitch, line 36: add comments

Tankilevitch (approved):

integrations/azure-devops/azure_devops/webhooks/webhook_event.py
omby8888:
```suggestion
                if not (self.publisherInputs or subscribed_event.publisherInputs) or (self.publisherInputs and subscribed_event.publisherInputs and subscribed_event.publisherInputs.get(
                        "projectId", None
                    ) == self.publisherInputs.get("projectId", None):):
                    return subscribed_event
```

PR #845, opened 2024-07-29
[AWS] Bug fix global resources not synced for all accounts

shalev007:
# Description

What - 
Fixed an issue in the AWS integration where S3 buckets were not synced for all accounts.

Why - 
S3 buckets are a global resource type, meaning we only search for the first region showing these buckets and then stop searching to avoid rate limiting with AWS and wasting time. This caused users with multi-account setups to only list their first account's buckets.

How - 
Added an iteration of accounts on top of the existing process to ensure that S3 buckets are synced across all accounts.

## Type of change

## Screenshots

## API Documentation

### DIFF
--- a/integrations/aws/main.py
+++ b/integrations/aws/main.py
@@ -1,11 +1,13 @@
 import json
 import typing
 
+import aioboto3
 from fastapi import Response, status
 import fastapi
 from starlette import responses
 from pydantic import BaseModel
 
+from aws.aws_credentials import AwsCredentials
 from port_ocean.core.models import Entity
 
 from utils.resources import (
@@ -18,6 +20,8 @@ from utils.resources import (
 
 from utils.aws import (
     describe_accessible_accounts,
+    get_accounts,
+    get_default_region_from_credentials,
     get_sessions,
     update_available_access_credentials,
     validate_request,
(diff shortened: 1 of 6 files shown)


integrations/aws/main.py
```diff
@@ -33,21 +37,49 @@
...
+                logger.info(f"Trying to resync {kind} in all regions until success")
+                async for session in credentials.create_session_for_each_region():
+                    s = await session
+                    try:
+                        async for batch in resync_cloudcontrol(kind, s):
+                            yield batch
+                        break
+                    except Exception as e:
+                        continue
+
```
Tankilevitch, line 61: what if there is any other error? what will we do with it? seems like you totally silence it

integrations/aws/main.py
```diff
@@ -33,21 +37,49 @@
...
+        if is_access_denied_exception(e):
+            if handle_exceptions:
+                logger.info(f"Trying to resync {kind} in all regions until success")
+                async for session in credentials.create_session_for_each_region():
+                    s = await session
+                    try:
+                        async for batch in resync_cloudcontrol(kind, s):
+                            yield batch
+                        break
+                    except Exception as e:
```
Tankilevitch, line 59: I am less of a fan of doing such a complicated logics inside error handling, my suggestion would be to return an value to indicate that it failed on authorization and if that was received then act on performing the logic inside of the current `except` block

Tankilevitch (changes requested):

integrations/aws/main.py
shalev007: The error message is inside `resync_cloudcontrol`

integrations/aws/main.py
```diff
@@ -33,21 +37,51 @@
...
+            if handle_exceptions:
+                logger.info(f"Trying to resync {kind} in all regions until success")
+                async for async_session in credentials.create_session_for_each_region():
+                    session = await async_session
+                    try:
+                        async for batch in resync_cloudcontrol(kind, session):
+                            yield batch
+                        break
+                    except Exception as e:
+                        continue
```
Tankilevitch, line 60:
```suggestion
async def _handle_global_resource_resync(
    kind: str,
    credentials: AwsCredentials,
    current_session: aioboto3.Session,
    handle_exceptions: bool = True,
) -> ASYNC_GENERATOR_RESYNC_TYPE:
    fetched_default_region = True
    default_region = get_default_region_from_credentials(credentials)
    try:
        for region in credentials.enabled_regions:
            if region != default_region:
                continue
            else:
                async for batch in resync_cloudcontrol(kind, current_session):
                    yield batch
    except Exception as e:
        if is_access_denied_exception(e):
            fetched_default_region = False

    if fetched_default_region:
        return

    for region in credentials.enabled_regions:
        async for batch in resync_cloudcontrol(kind, current_session):
            yield batch
```

Tankilevitch, line 60: also we shouldn't silent fail, only when access denied or not found and also then to log it

integrations/aws/main.py
```diff
@@ -33,21 +37,51 @@
...
+        logger.info(f"Trying to resync {kind} in all regions until success")
+        async for session in credentials.create_session_for_each_region():
+            try:
+                async for batch in resync_cloudcontrol(kind, session):
+                    yield batch
+                break
+            except Exception as e:
+                if not is_access_denied_exception(e):
+                    raise e
+
```
Tankilevitch, line 65: lets log if its access denied

integrations/aws/main.py
```diff
@@ -33,21 +37,51 @@
...
+) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    denied_access_to_default_region = False
+    default_region = get_default_region_from_credentials(credentials)
+    default_session = await credentials.create_session(default_region)
+    try:
+        async for batch in resync_cloudcontrol(kind, default_session):
+            yield batch
+    except Exception as e:
+        if is_access_denied_exception(e):
+            denied_access_to_default_region = True
```
Tankilevitch, line 51: same here

Tankilevitch (changes requested):

integrations/aws/main.py
shalev007: it will log it twice since it's already logging it in the `resync_cloudcontrol` function

integrations/aws/main.py
shalev007: it will log it twice since it's already logging it in the resync_cloudcontrol function

Tankilevitch (approved):
LGTM

PR #855, opened 2024-07-31
[Integration][SonarQube] Added _target='blank' attribute to spec links

kodjomiles:
Update the description content and the attribute target="_blank" to anchor tags in descriptions to open a new browser tab instead of the current browser.


phalbert (changes requested):
Nit change

kodjomiles:
> Nit change

@phalbert can you please elaborate here

PR #867, opened 2024-08-01
[Integration][Datadog] Add metrics availabity data on service and hosts

phalbert:
# Description

- Add new kinds of `metric` and `serviceMetric` to get specified metrics

## Type of change

## Screenshots

Screenshot 2024-08-13 at 07 59 38

Screenshot 2024-08-13 at 07 54 03
Screenshot 2024-08-13 at 07 54 16
Screenshot 2024-08-13 at 07 54 35

## API Documentation


phalbert:
A few other improvements I could make. cc @PeyGis , @Tankilevitch 

1. **Dashboard and Query Data**: Implement a caching mechanism to store fetched dashboard definitions and query results for a certain duration. 

2. **Time Window**: Allow users to configure the time window (currently hardcoded to 10 minutes) for the `check_metric_data_exists` function. This gives users more control over the metric availability checks.

Tankilevitch (changes requested):
I think the implementation is too specific for the user use-case.

We have multiple resources touched to support that capability. And I think they needs to be separated.

`Service`
`Host`
`Dashboard`
`Dashboard-Metrics`

IMO we need to approach it by taking advantage of port builder capabilities, relations, mirrors, aggregation properties. 

Service -> Dashboard Metrics or Service -> Dashboard -> Metrics 
Or
Metrics -> Service

phalbert:
> I think the implementation is too specific for the user use-case.
> 
> We have multiple resources touched to support that capability. And I think they needs to be separated.
> 
> `Service` `Host` `Dashboard` `Dashboard-Metrics`
> 
> IMO we need to approach it by taking advantage of port builder capabilities, relations, mirrors, aggregation properties.
> 
> Service -> Dashboard Metrics or Service -> Dashboard -> Metrics Or Metrics -> Service

Screenshot 2024-08-07 at 00 27 23

This change now adds available metrics as relational array of metric ids that are available for a host or service

PeyGis (changes requested):

integrations/datadog/client.py
```diff
@@ -233,6 +296,373 @@ async def get_slo_history(
...
+        if not dashboard_id:
+            return None
+        url = f"{self.api_url}/api/v1/dashboard/{dashboard_id}"
+        return await self._send_api_request(url)
+
+    async def enrich_dashboard_url(self, dashboard: dict[str, Any]) -> dict[str, Any]:
+        """
+        Enriches a dashboard item with the full Datadog web URL.
+        """
+        dashboard["url"] = f"{self.datadog_web_url}{dashboard['url']}"
```
PeyGis, line 398: we can construct the url using the solution I shared in Slack: where we can use jq to access .env.INTEGRATION_VARIABLE

integrations/datadog/client.py
```diff
@@ -233,6 +296,373 @@ async def get_slo_history(
...
+
+        while True:
+            url = f"{self.api_url}/api/v1/dashboard"
+            result = await self._send_api_request(
+                url,
+                params={
+                    "start": page,
+                    "count": page_size,
+                    "filter[shared]": "false",
+                    "filter[deleted]": "false",
```
PeyGis, line 413: since they are both set to false, do we know their default values?

integrations/datadog/client.py
```diff
@@ -233,6 +296,373 @@ async def get_slo_history(
...
+                },
+            )
+            dashboards = result.get("dashboards")
+            if not dashboards:
+                break
+
+            # Enrich items concurrently using asyncio.gather
+            dashboards = await asyncio.gather(
+                *[self.enrich_dashboard_url(dashboard) for dashboard in dashboards]
+            )
```
PeyGis, line 423: if we go with the use of jq to access the datadog web url env variable, then we will not need this

integrations/datadog/client.py
```diff
@@ -233,6 +296,373 @@ async def get_slo_history(
...
+            The structure of each widget dictionary depends on the Datadog API response.
+        """
+        async for dashboards in self.get_dashboards():
+            metrics = await process_in_queue(
+                [dashboard["id"] for dashboard in dashboards],
+                self.get_single_dashboard,
+                concurrency=MAX_CONCURRENT_REQUESTS,
+            )
+
+            for metric in metrics:
```
PeyGis, line 455: I am not familiar with datadog dashboard but does the widget contain the metrics? From the docs, I only see widgets instead of metrics

integrations/datadog/client.py
```diff
@@ -233,6 +296,373 @@ async def get_slo_history(
...
+        self, widget: dict[str, Any], dashboard_id: str
+    ) -> dict[str, Any]:
+        """
+        Adds the dashboard ID to a widget dictionary.
+        """
+        widget["dashboard_id"] = dashboard_id
+        return widget
+
+    async def get_dashboard_metrics(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        """
```
PeyGis, line 438: one thing I am curious about is the representation of the data from the dashboard and dashboard widgets. I was hoping we could embed the actual widget in Port using the url but from the code, it looks like we are only keeping track of it's properties

integrations/datadog/client.py
```diff
@@ -233,6 +296,373 @@ async def get_slo_history(
...
+            yield dashboards
+            page += 1
+
+    async def add_dashboard_id_to_widget(
+        self, widget: dict[str, Any], dashboard_id: str
+    ) -> dict[str, Any]:
+        """
+        Adds the dashboard ID to a widget dictionary.
+        """
+        widget["dashboard_id"] = dashboard_id
```
PeyGis, line 434: instead of setting the id of the dashboard, can we set the whole dashboard object so that users can choose how they want to relate the dashboard to other resources, not be id, but by name maybe?

if agreed, then we can use this format

widget["__dashboard"] = dashboard

PeyGis (changes requested):

matan84 (changes requested):
Left a few comments, will re-review once fixed

integrations/datadog/client.py
phalbert: Every widget contains metrics.

integrations/datadog/client.py
phalbert: no longer relevant bse the implementation changed

integrations/datadog/client.py
```diff
@@ -233,6 +296,117 @@ async def get_slo_history(
...
+        """
+        url = f"{self.api_url}/api/v1/metrics"
+
+        metrics = []
+
+        services = []
+        if service == "*":
+            # get all services
+            async for service_list in self.get_services():
+                services.extend(service_list)
```
Tankilevitch, line 370: should we actually wait for all the services to start fetching service metrics? why not after receiving a batch starting querying its metrics?

integrations/datadog/client.py
```diff
@@ -233,6 +296,117 @@ async def get_slo_history(
...
+
+        logger.info(f"Fetching metrics for {len(services)} services")
+
+        for service in services:
+            service_id = service["attributes"]["schema"]["dd-service"]
+
+            params = {"env": env, "service": service_id}
+            query_with_values = self._create_query_with_values(
+                f"{query}{{service:{service_id},env:{env}}}", params
+            )
```
Tankilevitch, line 388: I think user will want to have a different entity for a service when its in `prod` and when its in `dev`, I would also make sure I list all environments and iterate over them

integrations/datadog/client.py
```diff
@@ -233,6 +296,117 @@ async def get_slo_history(
...
+
+            time_in_minutes = 10
```
Tankilevitch, line 390: const

Tankilevitch: is it something user might want to configure? or its better we are managing it?

integrations/datadog/client.py
```diff
@@ -233,6 +296,117 @@ async def get_slo_history(
...
+            end_time = int(time.time())
+            start_time = end_time - (time_in_minutes * 60)
+
+            url = f"{self.api_url}/api/v1/query?from={start_time}&to={end_time}&query={query_with_values}"
+            result = await self._fetch_with_rate_limit_handling(
+                url, semaphore=self._metrics_semaphore
+            )
+
```
Tankilevitch, line 398: you are not even running it concurrently how come you need the semaphore? I would expect the code to spawn multiple requests to get that metric data

Tankilevitch: as suggested above, once you get a batch, spawn a few concurrent requests per to the allowed in datadog by the rate limit

Tankilevitch (changes requested):

integrations/datadog/client.py
phalbert: maybe we can allow user to configure it

integrations/datadog/main.py
```diff
@@ -76,10 +78,33 @@ async def on_resync_services(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+
+    if kind == ObjectKind.SERVICE_METRIC:
+        logger.info(
+            f"Fetching metrics for {params.metric} for service {params.service}"
+        )
+
+        metric_name = dd_client.extract_metric_name(params.metric)
+        metric_metadata = await dd_client.get_metric_metadata(metric_name)
+        metric_metadata["metric"] = metric_name
+        await ocean.register_raw(ObjectKind.METRIC, [metric_metadata])
```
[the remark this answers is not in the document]
phalbert: okay

integrations/datadog/.port/resources/port-app-config.yaml
```diff
@@ -102,3 +102,41 @@ resources:
...
+          title: .short_name // .metric
+          blueprint: '"datadogMetric"'
+          properties:
+            description: .description
+            short_name: .short_name
+            type: .type
+            unit: .unit
+            per_unit: .per_unit
+            statsd_interval: .statsd_interval
+            integration: .integration
```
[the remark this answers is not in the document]
phalbert: no, we are not exporting all. I have changed it anyway to have the metric in the serviceMetric blueprint

integrations/datadog/client.py
phalbert: i have done this

integrations/datadog/.port/resources/port-app-config.yaml
```diff
@@ -102,3 +102,41 @@ resources:
...
+  - kind: serviceMetric
+    selector:
+      query: "true"
+      metric: "avg:system.mem.used"
+      env: prod
+      service: '*'
```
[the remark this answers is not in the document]
phalbert: solved

integrations/datadog/client.py
phalbert: user can now configure

integrations/datadog/client.py
phalbert: resolved now

integrations/datadog/.port/resources/port-app-config.yaml
```diff
@@ -102,3 +102,27 @@ resources:
...
+          properties:
+            query: .__query
+            series: .series
+            res_type: .res_type
+            from_date: '.from_date / 1000 | todate'
+            to_date: '.to_date / 1000 | todate'
+            metric_name: .__metric
+            env: .__env
+          relations:
+            service: .__service
```
Tankilevitch, line 128: looks good, but I think this and the blueprint shouldn't be by default, as user wouldn't want to export system.mem.used by default, so lets create docs containing this example.

Tankilevitch: also lets say user want to actually see the current value of the metric how would that be achieved

integrations/datadog/.port/spec.yaml
```diff
@@ -5,11 +5,12 @@ features:
...
-      - kind: datadogMonitor
-      - kind: datadogService
-      - kind: datadogSlo
-      - kind: datadogSloHistory
+      - kind: host
+      - kind: monitor
+      - kind: service
+      - kind: slo
+      - kind: sloHistory
+      - kind: serviceMetric
```
Tankilevitch, line 13: nice catch

integrations/datadog/overrides.py
```diff
@@ -33,12 +33,31 @@ class SLOHistoryResourceConfig(ResourceConfig):
...
+    metric: str = Field(alias="metric", default="")
+    env: str = Field(alias="env", default="")
+    service: str = Field(alias="service", default="")
+    timeframe: int = Field(
+        alias="timeframe", description="Time frame in minutes", default=10
     )
 
 
+class DatadogResourceSelector(Selector):
+    datadogSelector: DatadogSelector = Field(
```
Tankilevitch, line 46:
```suggestion
    datadog_selector: DatadogSelector = Field(
```

integrations/datadog/overrides.py
```diff
@@ -33,12 +33,31 @@ class SLOHistoryResourceConfig(ResourceConfig):
     selector: SLOHistorySelector
 
 
-class DataDogPortAppConfig(PortAppConfig):
-    resources: list[SLOHistoryResourceConfig | ResourceConfig] = Field(
-        default_factory=list
+class DatadogSelector(BaseModel):
+    metric: str = Field(alias="metric", default="")
```
Tankilevitch, line 37: should be required

integrations/datadog/overrides.py
```diff
@@ -33,12 +33,31 @@ class SLOHistoryResourceConfig(ResourceConfig):
...
+    env: str = Field(alias="env", default="")
+    service: str = Field(alias="service", default="")
```
Tankilevitch, line 39: why default is not `*`?

integrations/datadog/main.py
```diff
@@ -76,10 +77,27 @@ async def on_resync_services(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
     dd_client = init_client()
 
     async for services in dd_client.get_services():
-        logger.info(f"Received batch with {len(services)} service catalogs")
+        logger.info(f"Received batch with {len(services)} services")
         yield services
 
 
+@ocean.on_resync(ObjectKind.SERVICE_METRIC)
+async def on_resync_metrics(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
```
Tankilevitch, line 85:
```suggestion
async def on_resync_service_metrics(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
```

integrations/datadog/client.py
```diff
@@ -233,6 +297,176 @@ async def get_slo_history(
...
+                        "__env": env_to_fetch,
+                    }
+                )
+                metrics.append(result)
+
+            yield metrics
+
+    async def get_metrics(
+        self,
+        query: str,
```
Tankilevitch, line 426:
```suggestion
        metric: str,
```

integrations/datadog/client.py
```diff
@@ -233,6 +297,176 @@ async def get_slo_history(
...
+        env: str = "*",
+        service: str = "*",
+        time_window: int = FETCH_WINDOW_TIME_IN_MINUTES,
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        """
+        Fetches metrics for specified services and environment.
+
+        Args:
+            query (str): The Datadog query string to execute.
```
Tankilevitch, line 435:
```suggestion
            metric (str): The Datadog metric to fetch.
```

integrations/datadog/client.py
```diff
@@ -233,6 +297,176 @@ async def get_slo_history(
...
+            env (str): The environment to filter by (default: "prod").
```
Tankilevitch, line 436: this is wrong, actually maybe we should set default to `prod` also in the selector, I think its better practice than `*`

integrations/datadog/client.py
```diff
@@ -233,6 +297,176 @@ async def get_slo_history(
             )
             return {}
 
+    async def get_tags(self) -> dict[str, Any]:
+        url = f"{self.api_url}/api/v1/tags/hosts"
```
Tankilevitch, line 301: why /hosts if we said `serviceMetrics` I am a bit confused

integrations/datadog/client.py
```diff
@@ -233,6 +297,176 @@ async def get_slo_history(
...
+
+        Yields:
+            AsyncGenerator[list[dict[str, Any]], None]: Each individual metric as it's fetched.
+        """
+
+        envs_to_fetch = (
+            [env] if env != "*" else self.get_env_tags(await self.get_tags())
+        )
+        if not envs_to_fetch:
+            logger.warning("No environments found, exiting...")
```
Tankilevitch, line 448:
```suggestion
            logger.warning(f"No environments found, can't fetch metrics for metric {metric}")
```

integrations/datadog/client.py
```diff
@@ -233,6 +297,176 @@ async def get_slo_history(
...
+            return
+
+        if service == "*":
+            async for service_list in self.get_services():
+                async for metric in self._fetch_metrics_for_services(
+                    query, envs_to_fetch, service_list, time_window
+                ):
+                    yield metric
+        else:
+            result = await self.get_single_service(service)
```
Tankilevitch, line 458: just to make sure, it returns empty result? and not an error of 404?

integrations/datadog/client.py
```diff
@@ -233,6 +297,176 @@ async def get_slo_history(
...
+            for result, env_to_fetch in zip(results, envs_to_fetch):
+                # Update result with metadata
+                result.update(
+                    {
+                        "__service": service_id,
+                        "__metric": self.extract_metric_name(query),
+                        "__query_id": f"{query}/service:{service_id}/env:{env_to_fetch}",
+                        "__query": query,
+                        "__env": env_to_fetch,
+                    }
```
Tankilevitch, line 418: const it

Tankilevitch (changes requested):

integrations/datadog/.port/resources/port-app-config.yaml
phalbert: It is a timeseries so the user can see the latest series by specifying 1 minute

integrations/datadog/overrides.py
phalbert: oh, yes. nice catch

integrations/datadog/overrides.py
phalbert: definitely

integrations/datadog/client.py
phalbert: tags are picked from hosts in the datadog agent. The agent will pick all tags from a host while sending metrics, tracces, logs, e.t.c

integrations/datadog/client.py
phalbert: Actually, '*' is better since we can't set `prod` as another user may have `production` or `live`

integrations/datadog/client.py
phalbert: nice catch. I have revised this

integrations/datadog/client.py
phalbert: elaborate. i don't get it yet

integrations/datadog/.port/resources/port-app-config.yaml
Tankilevitch: Let's update default to 1 min

Tankilevitch: And have an example on how to actually map the value

integrations/datadog/client.py
Tankilevitch: The underscores __ should be in consts

integrations/datadog/overrides.py
```diff
@@ -33,12 +33,29 @@ class SLOHistoryResourceConfig(ResourceConfig):
     selector: SLOHistorySelector
 
 
-class DataDogPortAppConfig(PortAppConfig):
-    resources: list[SLOHistoryResourceConfig | ResourceConfig] = Field(
-        default_factory=list
+class DatadogSelector(BaseModel):
+    metric: str = Field(alias="metric", required=True)
+    env: str = Field(alias="env", default="*")
```
Tankilevitch, line 38: Default prod

Tankilevitch (changes requested):

integrations/datadog/overrides.py
phalbert: How about if another user called it `production`?

integrations/datadog/client.py
```diff
@@ -14,6 +17,11 @@
 MAX_PAGE_SIZE = 100
 MAX_CONCURRENT_REQUESTS = 2
 
+MAXIMUM_CONCURRENT_REQUESTS_METRICS = 20
+MAXIMUM_CONCURRENT_REQUESTS_DEFAULT = 1
+MINIMUM_LIMIT_REMAINING = 1
+DEFAULT_SLEEP_TIME = 0.1
```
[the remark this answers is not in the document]
phalbert: yes

integrations/datadog/.port/resources/port-app-config.yaml
phalbert: I have added the docs https://github.com/port-labs/port-docs/pull/1502

integrations/datadog/overrides.py
Tankilevitch: I thought its auto defined by datadog 🤔

integrations/datadog/client.py
```diff
@@ -233,6 +303,173 @@ async def get_slo_history(
...
+        """
+        logger.debug(f"Creating query with values: {metric_query} | {variable_values}")
+        for variable_name, value in variable_values.items():
+            metric_query = metric_query.replace(
+                f"${variable_name}", f"{variable_name}:{value}"
+            )
+        return metric_query
+
+    def extract_metric_name(self, query_string: str) -> str | None:
+        """Extracts the Datadog metric name from a query string.
```
Tankilevitch, line 334:
```suggestion

    def _extract_metric_name(self, query_string: str) -> str | None:
        """Extracts the Datadog metric name from a query string.
```

integrations/datadog/client.py
```diff
@@ -233,6 +303,173 @@ async def get_slo_history(
...
+        metric_query: str,
+        env: str = "*",
+        service: str = "*",
+        time_window: int = FETCH_WINDOW_TIME_IN_MINUTES,
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        """
+        Fetches metrics for specified services and environment.
+
+        Args:
+            metric (str): The Datadog metric to fetch.
```
Tankilevitch, line 439: add example how `metric_query` looks like.
also align parameter naming in the docs

integrations/datadog/client.py
```diff
@@ -233,6 +303,173 @@ async def get_slo_history(
...
+            query_string: The Datadog query string.
+
+        Returns:
+            The metric name, or None if no metric name is found.
+
+        Examples:
+            - "sum:container.memory.usage" -> "container.memory.usage"
+            - "avg:system.cpu.used{service:my-service}" -> "system.cpu.used"
+        """
+        match = re.search(r"(?:\w+:)?(\w+\.\w+\.\w+)", query_string)
```
Tankilevitch, line 346: why we are doing that? as eventually I do care about the `avg` in `avg:system.cpu.used`? 
is there a different naming terminology between `avg:system.cpu.used` to `system.cpu.used` in datadog?

Tankilevitch (changes requested):

integrations/datadog/overrides.py
phalbert: @Tankilevitch, what is reliable is that we can always find the tag env but when setting up the datadog agent on a host, a user specifies the value of the tag. I setup host and specified prod and staging. But another user could set production, test or something else

integrations/datadog/client.py
phalbert: alright

integrations/datadog/client.py
phalbert: this was just metadata info. we still have the full metric name in the title of the entity and query property


```
{
  "identifier": "avg:system.mem.used/service:mayicard-mayicore/env:staging",
  "title": "avg:system.mem.used{service:mayicard-mayicore,env:staging}",
  "icon": null,
  "blueprint": "datadogServiceMetric",
  "team": [],
  "properties": {
    "query": "avg:system.mem.used",
    "series": [],
    "res_type": "time_series",
    "from_date": "2024-08-16T09:15:34Z",
    "to_date": "2024-08-16T09:25:34Z",
    "metric_name": "system.mem.used",
    "env": "staging"
  },
  "relations": {
    "service": "mayicard-mayicore"
  }
}
```

Tankilevitch (approved):
LGTM

PR #870, opened 2024-08-01
[Core] Fix unpacking issue in unregister raw function

mk-armah:
# Description

**Fix unpacking error in unregister raw:**

**What** -  Resolved an issue in the `unregister_raw` method where an incorrect unpacking operation was causing a `ValueError ("not enough values to unpack (expected 2, got 1)")`. 

**Why** - This error occurred because the method was incorrectly expecting a tuple of entities and errors from the asyncio.gather call, but only receiving a single list due to the absence of zip unpacking. 

**How** -The fix involved using zip to correctly unpack the results into entities and errors, and then flattening these lists to handle multiple resource mappings properly. This change ensures that the method correctly processes the output of the asynchronous tasks and handles errors appropriately.

## Type of change

## Screenshots

## API Documentation


CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+## 0.9.8 (2024-08-01)
+
+
+### Bug Fixes
+
+- Fix unpacking error in unregister raw
```
matan84, line 15: Add more data here- what was broken, and how it was fixed

matan84 (changes requested):
Good job, left one comment

yairsimantov20 (approved):
🌊 
Good job!

matan84 (approved):
LGTM

PR #881, opened 2024-08-04
Port 8461 ocean integration config is not a valid dict

yairsimantov20:
# Description

What - The gitlab integration is failing to startup when using the env var syntax to populate the integration config with an object that holds the tokken mapping which is also an object
Why - the inner object was calculated as a string
How - parsing it with pydantic parser if its a string

## Type of change

## Screenshots

## API Documentation


port_ocean/config/base.py
```diff
@@ -17,7 +17,7 @@
 def read_yaml_config_settings_source(
     settings: "BaseOceanSettings", base_path: str
 ) -> dict[str, Any]:
-    yaml_file = getattr(settings.__config__, "yaml_file", "")
+    yaml_file = getattr(settings.Config, "yaml_file", "")
```
Tankilevitch, line 21: seems like the `base_path` is redundant now, and you can use `settings._bash_path` inside the method

port_ocean/config/settings.py
```diff
@@ -73,6 +76,26 @@ class IntegrationConfiguration(BaseOceanSettings, extra=Extra.allow):
     integration: IntegrationSettings = IntegrationSettings(type="", identifier="")
     runtime: Runtime = "OnPrem"
 
+    @root_validator()
+    def validate_integration_config(cls, values: dict[str, Any]) -> dict[str, Any]:
+        if not (config_model := values.get("_integration_config_model")):
+            return values
+
+        if not (integration_config := values.get("integration")):
+            return values
```
Tankilevitch, line 85: why we need this case?

port_ocean/config/settings.py
```diff
@@ -73,6 +76,26 @@ class IntegrationConfiguration(BaseOceanSettings, extra=Extra.allow):
     integration: IntegrationSettings = IntegrationSettings(type="", identifier="")
     runtime: Runtime = "OnPrem"
 
+    @root_validator()
+    def validate_integration_config(cls, values: dict[str, Any]) -> dict[str, Any]:
+        if not (config_model := values.get("_integration_config_model")):
```
Tankilevitch, line 81: I don't fully get the full logic of the method, is the function is only responsible for validation? or it also sets the `integration_config` why is that?

Tankilevitch (changes requested):
🌊

port_ocean/config/settings.py
yairsimantov20: removed

port_ocean/config/settings.py
```diff
@@ -73,6 +76,24 @@ class IntegrationConfiguration(BaseOceanSettings, extra=Extra.allow):
...
+        def parse_config(model: Type[BaseModel], config: Any) -> BaseModel:
+            if isinstance(config, str):
+                return parse_raw_as(model, config)
+            else:
+                return parse_obj_as(model, config)
+
+        integration_config = values["integration"]
+        integration_config.config = parse_config(
+            config_model, integration_config.config
+        )
```
Tankilevitch, line 93: lets add a comment why we are doing that here

Tankilevitch (approved):
LGTM 🚀

PR #889, opened 2024-08-05
[Integration][AWS] throw error if no permissions on region

shalev007:
# Description

What - We have implemented blacklist error handling for AWS errors in our system. This ensures that certain AWS error types are specifically caught and handled differently from other errors.

Why - Each AWS resource handles "not found" errors differently, making it challenging to distinguish between a resource not being found and an actual error in the system. This inconsistency can lead to misinterpretation of errors and inappropriate error handling. By adding blacklist error handling, we can provide more accurate and consistent error handling across different AWS resources.

How - We have introduced a blacklist mechanism that explicitly catches specific AWS error codes related to "not found" scenarios. These error codes are then handled separately from other types of errors, allowing us to treat them as non-critical issues and proceed accordingly. This ensures that our system can gracefully handle situations where AWS resources are legitimately not found without mistaking them for system errors. This approach improves our system's robustness and reliability in interacting with AWS services.

## Type of change

## Screenshots

## API Documentation

### DIFF
--- a/integrations/aws/main.py
+++ b/integrations/aws/main.py
@@ -228,6 +228,10 @@ async def webhook(update: ResourceUpdate, response: Response) -> fastapi.Respons
                 if is_access_denied_exception(e):
                     raise e
                 resource = None
+                
+            # some aws resources return not found as an empty response
+            if not resource:
+                resource = None
 
             for kind in matching_resource_configs:
                 blueprints = matching_resource_configs[kind]
--- a/integrations/aws/main.py
+++ b/integrations/aws/main.py
@@ -226,9 +226,16 @@ async def webhook(update: ResourceUpdate, response: Response) -> fastapi.Respons
                     resource_type, identifier, account_id, region
                 )
             except Exception as e:
-                if is_access_denied_exception(e) or is_server_error(e):
+                if is_access_denied_exception(e):
                     logger.error(
-                        f"Skipping resyncing {resource_type} in region {region} due to missing access permissions"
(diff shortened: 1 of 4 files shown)


integrations/aws/main.py
```diff
@@ -224,7 +225,19 @@ async def webhook(update: ResourceUpdate, response: Response) -> fastapi.Respons
...
+                    )
+                    return fastapi.Response(
+                        status_code=status.HTTP_200_OK,
+                        content=json.dumps({"ok": True}),
+                    )
+                resource = None
+
+            # some aws resources return not found as an empty response
+            if not resource:
                 resource = None
```
Tankilevitch, line 241: redundant?

integrations/aws/main.py
```diff
@@ -224,7 +225,19 @@ async def webhook(update: ResourceUpdate, response: Response) -> fastapi.Respons
...
+            except Exception as e:
+                if is_access_denied_exception(e) or is_server_error(e):
+                    logger.error(
+                        f"Skipping resyncing {resource_type} in region {region} due to missing access permissions"
+                    )
+                    return fastapi.Response(
+                        status_code=status.HTTP_200_OK,
+                        content=json.dumps({"ok": True}),
+                    )
+                resource = None
```
Tankilevitch, line 237: if access denied we do want to try to delete I think

shalev007: Why is that?

Tankilevitch (approved):
bump project.toml and version and fix suggestion

PR #918, opened 2024-08-19
[GCP] Scale GCP Integration

mk-armah:
# Description

**What:**  
- Implemented a custom retry mechanism to replace GCP's AsyncRetry due to reliability issues.
- Added rate limiting for "Search All Resources" requests, configurable via the `SEARCH_ALL_RESOURCES_PER_MINUTE_QUOTA` environment variable.

**Why:**  
- The default GCP AsyncRetry mechanism exhibited reliability issues during operations, leading to the need for a more robust custom implementation.
- Rate limiting was introduced to ensure compliance with GCP's quota restrictions, allowing for configurable limits on the number of "Search All Resources" requests made per minute.

**How:**  
- Developed and integrated a custom retry mechanism tailored to handle specific scenarios more effectively than GCP's default.
- Introduced a rate limiting feature, controlled by the `SEARCH_ALL_RESOURCES_PER_MINUTE_QUOTA` environment variable, to manage API request flow.

## Type of change

## Screenshots
Screenshot 2024-08-27 at 9 49 51 AM


integrations/gcp/gcp_core/search/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,92 @@
...
+                return list(response.dimensions_infos)
+
+    async def _get_quota(self, container_id: str) -> int:
+        name = self.quota_name(container_id)
+
+        try:
+            quota_infos = await self._request_quota_info(name)
+            quota_info = quota_infos[0] if quota_infos else None
+
+            if not quota_info or quota_info.get("applicable_locations") != ["global"]:
```
matan84, line 48: Doesn't `not quota_info` mean he doesn't have any quota for this container?

integrations/gcp/gcp_core/search/helpers/ratelimiter/utils.py
matan84, line 1: Rename to cache.py

matan84: And move it somewhere else. it's not related to the ratelimiter specifically by default

matan84 (changes requested):
Left a few comments
In my view, these new files can be in the same heirarchy as the search, just with indicative names
If you think otherwise lets talk

Overall- Really impressive work here

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -38,70 +96,60 @@ async def search_all_resources(
 async def search_all_resources_in_project(
     project: dict[str, Any], asset_type: str, asset_name: str | None = None
 ) -> ASYNC_GENERATOR_RESYNC_TYPE:
```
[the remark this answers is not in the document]
mk-armah: Ouch, my bad

integrations/gcp/gcp_core/search/helpers/ratelimiter/base.py
mk-armah: yes, but specifically, for the asset api's quota,

integrations/gcp/.port/spec.yaml
```diff
@@ -10,6 +10,11 @@ configurations:
     type: string
     require: false
     description: Base64 encoding of your Google Cloud's ADC Credentials configuration file. For more information, see the <a href="https://docs.getport.io/build-your-software-catalog/sync-data-to-catalog/cloud-providers/gcp/installation" target="_blank">Google documentation</a>
+  - name: searchAllResourcesPerMinuteQuota
+    type: integer
+    require: false
+    description: The allowed number of requests to the GCP assets inventory. By default it is set to 400, but can be adjusted per to your GCP plan. For more information see the <a href="https://cloud.google.com/asset-inventory/docs/quota#project_quota" target="_blank">GCP quota documentation</a>
+    default: 400
 deploymentMethodRequirements:
```
Tankilevitch, line 18: is this still needed?

integrations/gcp/CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 <!-- towncrier release notes start -->
 
+## 0.1.46 (2024-08-26)
+
+
+### Improvements
+
+- Introduced AsyncGeneratorRetry, a custom retry mechanism designed to handle entire resource retrieval retries instead of individual page retries. This change addresses the recurring DEADLINE_EXCEEDED errors that occurred with GCP's AsyncRetry during page-based retries, significantly improving reliability in GCP integrations.
+- Implemented a sophisticated rate limiting system based on quota levels, allowing for precise control over request limits per project, per service, and per quota.
+- Added a `SEARCH_ALL_RESOURCES_PER_MINUTE_QUOTA` environment variable for controlling the rate limit quota should the integration fail to query quota from GCP as stated in the last point. Default is 400
```
Tankilevitch, line 17: in the code you use the parameter from config while here you are referring to environment variable

integrations/gcp/gcp_core/search/iterators.py
```diff
@@ -15,15 +14,18 @@ async def iterate_per_available_project(
...
+                return
+
             async for batch in stream_async_iterators_tasks(*tasks):
                 yield batch
-    except StopAsyncIteration:
-        raise NoProjectsFoundError(
-            "Performed a search_projects actions, and found no accessable Projects. This may be due to unset permissions or no current existing projects."
-        )
+    except Exception as e:
+        logger.exception(f"Error iterating over projects: {e}")
```
Tankilevitch, line 31: this can be problematic to just silent errors, as if we failed to bring data for some kind and the resync didn't receive it as failed it means that the resync passed successfuly and data that it wasn't able to fetch will be deleted as part of the resync process. This is not something we want to do as we would rather not delete from port if the state is incomplete rather than delete data that failed getting from the third part.

Only reasons to silent errors from third party is 403 or 404 where we either don't have access or there is no such resources to fetch.

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -38,144 +45,157 @@ async def search_all_resources(
...
+            }
+            for asset in assets
+        ]
+        return latest_resources
+
+    async with semaphore:
+        """
+        List of supported assets: https://cloud.google.com/asset-inventory/docs/supported-asset-types
+        Search for resources that the caller has ``cloudasset.assets.searchAllResources`` permission on within the project's scope.
+        """
```
Tankilevitch, line 64: should be at the start of the method

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -38,144 +45,157 @@ async def search_all_resources(
...
+                    search_all_resources_request,
+                    parse_asset_response,
+                    rate_limiter=asset_rate_limiter,
+                ):
+                    yield assets
+
+            except PermissionDenied:
+                logger.error(
+                    f"Service account doesn't have permissions to list topics from project {project_name}"
+                )
```
Tankilevitch, line 95: I think it doesn't have permissions to list this resource kind, not something specific for topics

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -38,144 +45,157 @@ async def search_all_resources(
...
+            except NotFound:
+                logger.debug(
+                    f"Couldn't perform search_all_resources on project {project_name} since it's deleted."
+                )
```
Tankilevitch, line 99: how this could be? that we are able to fetch it but not list it? I think worth having it as `info` as we do have this log that we start fetching from it

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,94 @@
+from abc import ABC, abstractmethod
+from typing import Any, Optional, TYPE_CHECKING
+from aiolimiter import AsyncLimiter
+from google.cloud.cloudquotas_v1 import CloudQuotasAsyncClient, GetQuotaInfoRequest  # type: ignore
+from loguru import logger
+from enum import Enum
+from port_ocean.context.ocean import ocean
+from gcp_core.cache import cache_coroutine_result  # type: ignore
```
Tankilevitch, line 8: why we type: ignore?

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,94 @@
...
+        quota = await self._get_quota(container_id)
+        limiter = AsyncLimiter(max_rate=quota, time_period=self.time_period)
+        return limiter
+
+    def quota_name(self, container_id: str) -> str:
+        return f"{self.container_type.value}/{container_id}/locations/global/services/{self.service}/quotaInfos/{self.quota_id}"
+
+    async def __getitem__(self, container_id: str) -> AsyncLimiter:
+        name = self.quota_name(container_id)
+        return await self.register(container_id, name)
```
Tankilevitch, line 94: add comment about this class usage and why we do it

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,94 @@
...
+    FOLDER = "folders"
+    ORGANIZATION = "organizations"
+
+
+class GCPResourceQuota(ABC):
+    quota_id: str | None = None
+    service: str | None = None
+    container_type: ContainerType = ContainerType.PROJECT
+    _default_quota: int = _DEFAULT_RATE_LIMIT_QUOTA
+
```
Tankilevitch, line 32: add comment about this class usage and why we need it and why we chose to implement it in this way

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,94 @@
...
+    @abstractmethod
+    def quota_name(self, container_id: str) -> str:
+        """Generate the fully qualified name for the quota resource."""
+        pass
+
+
+class GCPResourceRateLimiter(GCPResourceQuota):
+    time_period: float = _DEFAULT_RATE_LIMIT_TIME_PERIOD
+
+    @cache_coroutine_result()
```
Tankilevitch, line 83: how long would it be cached for?

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,94 @@
...
+    _default_quota: int = _DEFAULT_RATE_LIMIT_QUOTA
+
+    async def _request_quota_info(self, name: str) -> MutableSequence["DimensionsInfo"]:
+        async with AsyncLimiter(
+            max_rate=_DEFAULT_RATE_LIMIT_QUOTA,
+            time_period=_DEFAULT_RATE_LIMIT_TIME_PERIOD,
+        ):
+            async with CloudQuotasAsyncClient() as quotas_client:
+                request = GetQuotaInfoRequest(name=name)
+                response = await quotas_client.get_quota_info(request=request)
```
Tankilevitch, line 40: does it requires extra permissions to perform this request? is it something we should add as part of our docs and TF?

integrations/gcp/gcp_core/search/utils.py
```diff
@@ -1,50 +1,63 @@
...
+    """
+    page = 0
+    page_token = None
+
+    while True:
+        if page_token:
+            request["page_token"] = page_token
+
+        if rate_limiter:
+            logger.debug(f"Rate limiting enabled for `{method}`")
```
Tankilevitch, line 34: add what is the rate limit

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
mk-armah: Yes it does require the Cloud API Quota API to be on for the project (container) and also some permission relating to the service being tracked allocated as such. Adding to the docs is vital

integrations/gcp/gcp_core/search/iterators.py
mk-armah: logger.exception actually throws an the error 🤓

integrations/gcp/gcp_core/search/resource_searches.py
mk-armah: correct, an oversight

integrations/gcp/gcp_core/search/utils.py
mk-armah: its logged during the creation of the rate limiter object, but at this stage it will be difficult to get that

integrations/gcp/.port/spec.yaml
matan84: Yes, This is for non-terraform installations

integrations/gcp/gcp_core/search/utils.py
matan84: I agree that this shouldn't be logged here (Since it will mean that for each page this will be printed) but we definitely should log what is the max_rate that we are running with per resource per project.

Add this log to be above the while: true + the AsyncLimiter has a max_rate + time_period properies you can use to print

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -38,144 +45,157 @@ async def search_all_resources(
...
-        except PermissionDenied as e:
-            logger.exception(
-                f"Couldn't access the API Cloud Assets to get kind {asset_type}. Please set cloudasset.assets.searchAllResources permissions for project {project_name}"
-            )
-            raise e
 
 
 async def list_all_topics_per_project(
-    project: dict[str, Any],
+    project: dict[str, Any], **kwargs: Any
```
matan84, line 107:
```suggestion
    project: dict[str, Any]
```

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
matan84: We have to add this to the TF as well. Please specify which exact additional permissions are required here.

integrations/gcp/gcp_core/search/utils.py
```diff
@@ -1,50 +1,63 @@
...
+
+
+@async_generator_retry
+async def paginated_query(
+    client: Any,
+    method: str,
+    request: dict[str, Any],
+    parse_fn: Any,
+    rate_limiter: Optional[AsyncLimiter] = None,
+    *args: Any,
```
matan84, line 20:
```suggestion
    timeout: int = DEFAULT_REQUEST_TIMEOUT
```

Also remove kwargs

matan84 (changes requested):
added some changes

integrations/gcp/.port/spec.yaml
mk-armah: the `searchAllResourcesPerMinuteQuota` var, is what is used should the integration fail to fetch cloud quota api for any of the services, would you rather I have it defined as a constant in the code instead ?

mk-armah: A more granular approach in my opinion would be to allow optionally setting quota on each kind (service) to ingest via the `port-app-config`. 
That way, without having to activate cloud quota api, one can easily infer the allowed quota's for the service api and add it to the port app config specifications, the integration then utilizes these parameters to control rate limits per kind

integrations/gcp/CHANGELOG.md
mk-armah: In the code I'm actually fetching from `integration_config`.  the env ?🤔

integrations/gcp/.port/spec.yaml
matan84: This limit isn't really applying to the hard limit made by GCP. I agree this will be much cleaner, but considering we are hard limited by GCP we cannot allow two different limitations to exist, one logical and another actual limit

integrations/gcp/gcp_core/search/resource_searches.py
mk-armah: yeah, that.. actually, its meant to be there, I'm passing the kind from here
Screenshot 2024-08-27 at 10 41 35 AM

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
Tankilevitch: what happens if there are no permissions to get it? how users upgrades to this permission?

integrations/gcp/.port/spec.yaml
Tankilevitch: lets have it as an environment variable for now, with a default of 400. 
And do you use the same client for real time events as well?

integrations/gcp/gcp_core/utils.py
```diff
@@ -55,3 +69,37 @@ def get_current_resource_config() -> (
...
     )
+
+
+def get_credentials_json() -> str:
+    b64_credentials = ocean.integration_config["encoded_adc_configuration"]
+    credentials_json = base64.b64decode(b64_credentials).decode("utf-8")
+    return credentials_json
+
+
+def service_account_project() -> str:
```
matan84, line 80:
```suggestion
def get_service_account_project_id() -> str:
```

integrations/gcp/gcp_core/search/utils.py
matan84, line 1: Let's rename this file to pagintated_query.py

integrations/gcp/gcp_core/search/utils.py
```diff
@@ -1,50 +1,68 @@
...
+
+        items = parse_fn(response)
+        if items:
+            page += 1
+            logger.info(
+                f"Found {len(items)} items on page {page} for `{method}` with request: {request}"
+            )
+            yield items
+        else:
+            logger.warning(
```
matan84, line 59:
```suggestion
            logger.info(
```

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -25,169 +24,186 @@
...
 ) -> ASYNC_GENERATOR_RESYNC_TYPE:
-    """
```
matan84, line 82: Revert the comment delete, This specification is needed here

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -25,169 +24,186 @@
...
-                topics = parse_protobuf_messages(paginated_response.topics)
-                if topics:
-                    logger.info(
-                        f"Found {len(topics)} {AssetTypesWithSpecialHandling.TOPIC}'s"
-                    )
+        project_name = project["name"]
+        logger.info(
+            f"Searching all {AssetTypesWithSpecialHandling.TOPIC}'s in project {project_name}"
+        )
+        async with kwargs["topic_semaphore"]:
```
matan84, line 107: Do we need a semaphore or a aiolimiter here? I believe we need a aiolimiter, correct me if im wrong

integrations/gcp/gcp_core/search/resource_searches.py
matan84: Agreed that these logs should be info logs

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -25,169 +24,186 @@
...
+        latest_resources = [
+            {
+                **parse_latest_resource_from_asset(asset),
+                EXTRA_PROJECT_FIELD: project,
+            }
+            for asset in assets
+        ]
+        return latest_resources
+
+    async with kwargs["asset_semaphore"]:
```
matan84, line 61: Not sure we need the semaphore. We didn't encounter any problems with concurrency, the ratelimiter should suffice here

integrations/gcp/gcp_core/search/resource_searches.py
```diff
@@ -25,169 +24,186 @@
...
 ) -> ASYNC_GENERATOR_RESYNC_TYPE:
     """
     List of supported assets: https://cloud.google.com/asset-inventory/docs/supported-asset-types
     Search for resources that the caller has ``cloudasset.assets.searchAllResources`` permission on within the project's scope.
     """
-    project_name = project["name"]
-    logger.info(f"Searching all {asset_type}'s in project {project_name}")
-    async with AssetServiceAsyncClient() as async_assets_client:
+
+    def parse_asset_response(response: Any) -> list[dict[Any, Any]]:
```
matan84, line 50: Let's extract this method to a private method somewhere.

integrations/gcp/gcp_core/helpers/retry/async_retry.py
matan84, line 1: This file is very clean, really like what you did here :)

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,178 @@
...
+    This class allows efficient control over the rate of API requests per container (e.g., projects, folders) based on the actual quota allocated by GCP.
+    """
+
+    time_period: float = _DEFAULT_RATE_LIMIT_TIME_PERIOD
+
+    @cache_coroutine_result()
+    async def register(self, container_id: str, *arg: Optional[Any]) -> AsyncLimiter:
+        quota = await self._get_quota(container_id, *arg)
+
+        effective_quota_limit: int = int(max(round(quota * _PERCENTAGE_OF_QUOTA, 1), 1))
```
matan84, line 128: Do you need to apply the percentage again? You do that in the `effective_quota_limit` variable above

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,178 @@
...
+from collections.abc import MutableSequence
+import asyncio
+
+
+_DEFAULT_RATE_LIMIT_TIME_PERIOD: float = 60.0
+_DEFAULT_RATE_LIMIT_QUOTA: int = int(
+    ocean.integration_config["search_all_resources_per_minute_quota"]
+)
+_PERCENTAGE_OF_QUOTA: float = 0.8
+_MAXIMUM_CONCURRENT_REQUESTS: int = 10
```
matan84, line 20: Why 10?

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,178 @@
...
+from gcp_core.cache import cache_coroutine_result
+from collections.abc import MutableSequence
+import asyncio
+
+
+_DEFAULT_RATE_LIMIT_TIME_PERIOD: float = 60.0
+_DEFAULT_RATE_LIMIT_QUOTA: int = int(
+    ocean.integration_config["search_all_resources_per_minute_quota"]
+)
+_PERCENTAGE_OF_QUOTA: float = 0.8
```
matan84, line 19:
```suggestion
_PERCENTAGE_OF_QUOTA: float = 0.9
```

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,178 @@
...
+            _maximum_concurrent_requests = await self._maximum_concurrent_requests(
+                container_id
+            )
+            self._semaphore = asyncio.BoundedSemaphore(_maximum_concurrent_requests)
+        return self._semaphore
+
+    async def _maximum_concurrent_requests(self, container_id: str) -> int:
+        name = self.quota_name(container_id)
+        quota = await self.register(container_id, name)
+        limit = max(
```
matan84, line 168: Could you please explain how did you come to this calculation here?

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
```diff
@@ -0,0 +1,178 @@
...
+            int(quota.max_rate / self.default_maximum_limit), self.default_maximum_limit
+        )
+        logger.info(
+            f"The integration will process {limit} {self.container_type.value}'s at a time based on {container_id}'s/{self.service}/{self.quota_id} quota's capacity"
+        )
+        if limit <= self.default_maximum_limit:
+            logger.warning(
```
matan84, line 175: Nice!!!!

matan84 (changes requested):
Requested some changes and explanations, Really impressive stuff @mk-armah !!!

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
mk-armah: No no, we don't

integrations/gcp/gcp_core/search/resource_searches.py
mk-armah: You are absolutely right, I'll wouldn't be necessary to apply a semaphore on querying the topics kind for at the moment

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
mk-armah: The closer we get to the quota, the higher our chances of getting close to the rate limit and also, further to the retry mechanism.
Ideally, we don't want to be approaching the retry mechanism a lot because at the moment its retry's an entire resource and not a page. 

I must also add, that though we experience little to no rate limits now from tests I've done, I do see randomly (rare but observed) that the rate limit does hit, good part is that in such scenario's the retry mechanism does set in to compensate. However, getting away from the quota value by 20-30% does reduce chances of this happening.

integrations/gcp/gcp_core/helpers/ratelimiter/base.py
mk-armah: This gives an estimate of how many projects can be allowed to run within the `quota's limits` .

The higher the number of concurrent requests, the lesser the number of resources each project gets to query at a time before the quota gets exhausted. If the number of projects attached to the service account is relatively large, it means the integration will take a longer before ingesting resources to port since each of those projects would have to finish paginating through their resources (kind) first before an ingestion happens. However, by controlling this value, we directly control our ingestion `frequency` by simply obeying the quota of the limiter.




   - The `max` function is used to determine the larger value between the calculated limit and the `self.default_maximum_limit`.
   - This ensures that the system either uses the calculated limit (if it's greater) or defaults to the `self.default_maximum_limit` (if the calculated limit is smaller).

### Example:
Let's say the quota's maximum rate (`quota.max_rate`) is 320, and the `self.default_maximum_limit` is set to 10.

   - `limit = floor(320 / 10) = 32`
   - `semaphore = max(32, 10) = 32`

### Explanation:
In this case, the calculated limit is 32, which is greater than the default maximum limit of 10. Hence, the integration will permit a batch of `32 projects to battle for 320 requests` at a time. This increases the chance of us completing a sync for some projects within the batch. 
Note that for a sync to happen, all pages of a project's resource must be iterated through, it could be any number of pages though, but counts of all the requests needed to complete the iterations are deducted from the 320 for all projects within the batch.

mk-armah: @matan84  makes sense ?

matan84: Yup, makes sense

matan84 (changes requested):
Left really few comments

matan84 (approved):
LGTM

PR #945, opened 2024-08-26
[Integration][Octopus] Updated serverUrl description to be more expressive

oiadebayo:
# Description

What - Updated the description of serverUrl in the spec.yaml to be more expressive
Why - 
How -

## Type of change

## Screenshots

## API Documentation

### DIFF
--- a/integrations/octopus/.port/spec.yaml
+++ b/integrations/octopus/.port/spec.yaml
@@ -19,7 +19,7 @@ configurations:
   - name: serverUrl
     required: true
     type: url
-    description: The base URL of your Octopus Deploy instance. It should include the protocol (e.g., https://).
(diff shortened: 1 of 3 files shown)


integrations/octopus/.port/spec.yaml
```diff
@@ -19,7 +19,7 @@ configurations:
   - name: serverUrl
     required: true
     type: url
-    description: The base URL of your Octopus Deploy instance. It should include the protocol (e.g., https://).
+    description: The base URL of your Octopus Deploy instance. It should include the protocol (e.g. <a href="https://demo.octopus.app/" target="_blank">https://demo.octopus.app/</a> or <a href="https://example.com/" target="_blank">https://example.com/</a>).
```
PeyGis, line 22: Is there a reason we have two examples? one for demo and another for example.com?

PeyGis (changes requested):
Please bump the integration version in the pyproject.toml also

integrations/octopus/.port/spec.yaml
oiadebayo: The url could be a subdomain on octopus or self hosted domain

PR #952, opened 2024-08-26
[Integration][Sonarqube] - Fix sonarQube client has no attributes metrics

PeyGis:
# Description

What - The integration was throwing an error where the `self.metrics` was not found in the `get_all_sonarqube_analyses`(). This was because each resync handler creates a new instance of the sonarQube client, causing the client to lose the state of it's saved variables. An example is when the self.metrics in the previous handler is lost in subsequent clients. 
Why - Affecting the ingestion of onprem analysis kind as it relies on the metrics
How - Implemented a singleton pattern where we use one global instance of the sonarQube client across the application

## Type of change

## Screenshots

Screenshot 2024-08-26 at 9 07 58 PM

## API Documentation

### DIFF
--- a/integrations/sonarqube/CHANGELOG.md
+++ b/integrations/sonarqube/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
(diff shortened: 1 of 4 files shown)


Tankilevitch (approved):
LGTM, let's make sure that for scheduled and real time there are no multi threading issues. Different event loops issues

PR #956, opened 2024-08-27
[Framework] Fix bug search relation error unhashable dict

omby8888:
# Description

What - Refer to entity key in upsert entity response, if the response is not 200 then turn raw search relations into None to not harm the deletion proccess

Why - Response from upsert entity api isn't processed right and causes ocean core to use the raw entity that contains dict typed relations instead of string which is the relation's entity identifier

How - refer to the entity key in a response that looks like {ok: true, entity: {...}}, if the response is not 200 turn the dict typed relations into None 

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,19 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+## 0.10.3 (2024-08-27)
+
+### Bug Fixes
+
+- Handle upsert entity failure when saving modified data for search relations calculations
+
+
 ## 0.10.2 (2024-08-26)
 
 ### Bug Fixes
 
 - Reverted last bugfix
+
 
 ## 0.10.1 (2024-08-26)
 
--- a/port_ocean/clients/port/mixins/entities.py
+++ b/port_ocean/clients/port/mixins/entities.py
@@ -61,16 +61,19 @@ class EntityClientMixin:
         result_entity = (
             Entity.parse_obj(result["entity"]) if result.get("entity") else entity
         )
+        # In order to save memory we'll keep only the identifier and relations of the
+        # upserted entity result for later calculations
+        reduced_entity = Entity(identifier=result_entity.identifier)
 
         # Turning dict typed relations (raw search relations) is required
         # for us to be able to successfully calculate the participation related entities
         # and ignore the ones that don't as they weren't upserted
-        result_entity.relations = {
+        reduced_entity.relations = {
             key: None if isinstance(relation, dict) else relation
             for key, relation in result_entity.relations.items()
         }
 
-        return result_entity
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,12 +7,20 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+## 0.10.3 (2024-08-27)
+
+### Bug Fixes
+
+- Handle upsert entity failure when saving modified data for search relations calculations
```
talsabagport, line 14:
```suggestion
- Fixed upsert entity failure when saving modified data for search relations calculations
```

port_ocean/clients/port/mixins/entities.py
```diff
@@ -58,11 +58,16 @@ async def upsert_entity(
...
-        result_entity = Entity.parse_obj(result)
-        # Set the results of the search relation and identifier to the entity
-        entity.identifier = result_entity.identifier or entity.identifier
-        entity.relations = result_entity.relations or entity.relations
-        return entity
+        result_entity = (
+            Entity.parse_obj(result["entity"]) if result.get("entity") else entity
+        )
+
+        result_entity.relations = {
```
talsabagport, line 65: Add a comment to explain it

talsabagport (changes requested):

talsabagport (approved):

port_ocean/clients/port/mixins/entities.py
```diff
@@ -58,11 +58,19 @@ async def upsert_entity(
             )
         handle_status_code(response, should_raise)
         result = response.json()
-        result_entity = Entity.parse_obj(result)
-        # Set the results of the search relation and identifier to the entity
-        entity.identifier = result_entity.identifier or entity.identifier
-        entity.relations = result_entity.relations or entity.relations
-        return entity
+        result_entity = (
+            Entity.parse_obj(result["entity"]) if result.get("entity") else entity
```
Tankilevitch, line 62: returning all the result entity means that we return the full entity that is inside port, this means that if we just ingested the relations of an entity but it has 20 more properties it will be returned here and passed forward to the resync phase. This can be problematic from a memory point of view and that is why we only took the identifier and the relations from the result entity and not the full entity.

Tankilevitch (changes requested):

port_ocean/clients/port/mixins/entities.py
omby8888: we need the blueprint as well to delete the identifier of a specific blueprint only as there may be entities with same identifier but different blueprints

Tankilevitch (approved):
LGTM

PR #967, opened 2024-08-29
[Integration][OpsGenie] - Performance Improvement

PeyGis:
# Description

What - Made performance improvements on the OpsGenie integration.

Why - The integration was very slow due to multiple API calls it makes

How - Improved the integration in several ways
### Additions
1. Added new kinds for team and schedule. With this addition, users can bring teams and schedule rotations from OpsGenie to Port
2. Added support for filtering data directly from OpsGenie API. This is configured in the port-app-config
3. Added logs to make it easy to debug the integration in the event of an unexpected behaviour

### Breaking Changes
1. Removed the teams properties from the OpsGenieService blueprint and made it a relation instead.
2. Removed the extra API calls that were made to the OpsGenie API to describe all impacted services for each incident. This breaking change boosts the performance of the integration as we can still relate incidents to services without making extra API call
3. Removed the extra API calls that were made to find associated incident for each alert. The team found a much better way to still relate incidents to alerts using JQ. This change also improves the overall performance of the integration as we don't have to make extra API calls.

## Type of change

### Core testing checklist

- [x] Docs PR link [here](https://docs.getport.io/build-your-software-catalog/sync-data-to-catalog/incident-management/opsgenie)

### Preflight checklist

- [ ] Handled rate limiting
- [x] Handled pagination
- [ ] Support Multi account

## Screenshots
Screenshot 2024-09-02 at 9 08 44 PM

## API Documentation

Services - https://docs.opsgenie.com/docs/service-api
Incidents - https://docs.opsgenie.com/docs/incident-api
Alerts -  https://docs.opsgenie.com/docs/alert-api
More - https://docs.opsgenie.com/docs/api-overview


integrations/opsgenie/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
+## 0.1.74 (2024-08-29)
+
+
+### Improvements
+
+- Improved on the performance on the integration by implementing caching, adding logs for verbosity and making efficient lookup of resources (0.1.74)
```
Tankilevitch, line 15: lets bump minor

integrations/opsgenie/client.py
```diff
@@ -40,109 +41,130 @@ async def _get_single_resource(
...
+                    return schedule
 
-        api_version = await self.get_resource_api_version(ObjectKind.INCIDENT)
-        url = f"{self.api_url}/{api_version}/incidents/{incident_identifier}/associated-alert-ids"
-        associated_alerts = (await self._get_single_resource(url))["data"]
-        event.attributes[cache_key] = associated_alerts
-        return associated_alerts
+        return {}
 
     async def get_impacted_services(
```
Tankilevitch, line 123: do we still need the impacted_services? isn't it enough to just relate it using the `id` of the service?

integrations/opsgenie/main.py
```diff
@@ -33,31 +39,42 @@ async def enrich_services_with_team_data(
...
+            f"Got {len(impacted_service_ids)} unique impacted services from {len(incident_chunk)} incidents with a chunk size of {chunk_size}"
         )
-        incident["__impactedServices"] = impacted_services
-        return incident
 
+        # Fetch all impacted services for this chunk in one API call
+        async with semaphore:
+            all_impacted_services = await opsgenie_client.get_impacted_services(
+                list(impacted_service_ids)
+            )
```
Tankilevitch, line 66: if we do the relation based on `id` we don't need to fetch all of these entities anymore

Tankilevitch (changes requested):

integrations/opsgenie/main.py
```diff
@@ -16,48 +19,61 @@ def init_client() -> OpsGenieClient:
...
-        impacted_services = await opsgenie_client.get_impacted_services(
-            incident["impactedServices"]
-        )
-        incident["__impactedServices"] = impacted_services
-        return incident
+    schedule_ids = [schedule["id"] for schedule in schedules.values()]
 
+    async def fetch_oncall(schedule_id: str) -> tuple[str, dict[str, Any]]:
+        async with semaphore:
+            return schedule_id, await opsgenie_client.get_oncall_user(schedule_id)
```
Tankilevitch, line 36: can't we solve this by some solution on port side, like calculation property or something? I would like to try to reduce this kind of extra requests

Tankilevitch: I wonder if this can be a separate kind, something like service-on-call or something like that

integrations/opsgenie/integration.py
```diff
@@ -0,0 +1,109 @@
...
+
+        return value
+
+
+class AlertAndIncidentResourceConfig(ResourceConfig):
+    class AlertAndIncidentSelector(Selector):
+        api_query_params: APIQueryParams | None = Field(
+            alias="apiQueryParams",
+            description="The query parameters to filter alerts or incidents",
+        )
```
Tankilevitch, line 84: lets define it outside of the `AlertAndIncidentResourceConfig` it will help with easier referencing

integrations/opsgenie/integration.py
```diff
@@ -0,0 +1,109 @@
...
+    ResourceConfig,
+    PortAppConfig,
+    Selector,
+)
+from port_ocean.core.integrations.base import BaseIntegration
+
+
+class APIQueryParams(BaseModel):
+    created_at: str | None = Field(
+        alias="createdAt", description="The date and time the alert was created"
```
Tankilevitch, line 15: its being used only by the alerts kind, so lets provide it a more specific name

integrations/opsgenie/integration.py
```diff
@@ -0,0 +1,109 @@
...
+    is_seen: Literal["true", "false"] | None = Field(
+        description="Whether the alert has been seen"
+    )
+    acknowledged: Literal["true", "false"] | None = Field(
+        description="Whether the alert has been acknowledged"
+    )
+    snoozed: Literal["true", "false"] | None = Field(
+        description="Whether the alert has been snoozed"
+    )
+    priority: Literal["P1", "P2", "P3", "P4", "P5"] | None = Field(
```
Tankilevitch, line 38: is there possibility to specify your own priorities in opsgenie? I wonder if we actually want to restrict it, as it means any addition on opsgenie side, will require us to update the code

integrations/opsgenie/integration.py
```diff
@@ -0,0 +1,109 @@
...
+    status: Literal["open", "resolved", "closed"] | None = Field(
+        description="The status of the alert"
+    )
+    is_seen: Literal["true", "false"] | None = Field(
+        description="Whether the alert has been seen"
+    )
+    acknowledged: Literal["true", "false"] | None = Field(
+        description="Whether the alert has been acknowledged"
+    )
+    snoozed: Literal["true", "false"] | None = Field(
```
Tankilevitch, line 35: why this is not a boolean?

integrations/opsgenie/integration.py
```diff
@@ -0,0 +1,109 @@
...
+        for field, value in self.dict(exclude_none=True).items():
+            if isinstance(value, list):
+                params.append(f"{field}:{','.join(value)}")
+            else:
+                params.append(f"{field}:{value}")
+
+        return {"query": " AND ".join(params)}
+
+    class Config:
+        allow_population_by_field_name = True
```
Tankilevitch, line 63: add why

Tankilevitch (changes requested):
looks good left some comments

integrations/opsgenie/integration.py
PeyGis: This is used in both alerts and incidents

Tankilevitch (approved):
LGTM

PR #978, opened 2024-09-03
[Integration][SonarQube] - Support Portfolios

lordsarcastic:
# Description

What - Added support for SonarQube portfolio

Why -

How -

## Type of change

### Core testing checklist

- [x] Docs PR link [here](https://github.com/port-labs/port-docs/pull/1546)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [ ] Support Multi account

### Entities list
Screenshot 2024-09-04 at 13 57 55

### Entity detail
Screenshot 2024-09-04 at 13 57 18

## API Documentation
- [Portfolio details API](https://next.sonarqube.com/sonarqube/web_api/api/views/show)
- [Portfolio list API](https://next.sonarqube.com/sonarqube/web_api/api/views/list)


integrations/sonarqube/client.py
```diff
@@ -468,6 +472,39 @@ async def get_all_sonarqube_analyses(
...
+            endpoint=Endpoints.PORTFOLIO_DETAILS,
+            query_params={"key": portfolio_key},
+        )
+        return response
+
+    def _get_portfolio_subportfolios(
+        self, portfolio: dict[str, Any]
+    ) -> list[dict[str, Any]]:
+        logger.info(f"Fetching subportfolios for: {portfolio['key']}")
+        subportfolios = portfolio.get("subViews", []) or []
```
PeyGis, line 492:
```suggestion
        subportfolios = portfolio.get("subViews", [])
```

integrations/sonarqube/client.py
```diff
@@ -468,6 +472,39 @@ async def get_all_sonarqube_analyses(
...
+
+    async def get_portfolio_details(self, portfolio_key: str) -> dict[str, Any]:
+        logger.info(f"Fetching portfolio details for: {portfolio_key}")
+        response = await self.send_api_request(
+            endpoint=Endpoints.PORTFOLIO_DETAILS,
+            query_params={"key": portfolio_key},
+        )
+        return response
+
+    def _get_portfolio_subportfolios(
```
PeyGis, line 488: we can rename the function to _fetch_subportfolios or _extract_subportfolios to make it less verbose

integrations/sonarqube/client.py
```diff
@@ -468,6 +472,39 @@ async def get_all_sonarqube_analyses(
...
+        logger.info(f"Fetching subportfolios for: {portfolio['key']}")
+        subportfolios = portfolio.get("subViews", []) or []
+        all_portfolios = []
+        for subportfolio in subportfolios:
+            if subportfolio.get("qualifier") in PORTFOLIO_VIEW_QUALIFIERS:
+                all_portfolios.append(subportfolio)
+            all_portfolios.extend(self._get_portfolio_subportfolios(subportfolio))
+        return all_portfolios
+
+    async def get_all_portfolios(self) -> AsyncGenerator[list[dict[str, Any]], None]:
```
PeyGis, line 500: let's add some logs here too?

integrations/sonarqube/client.py
```diff
@@ -468,6 +472,39 @@ async def get_all_sonarqube_analyses(
...
+        portfolios = await self._get_all_portfolios()
+
+        for portfolio in portfolios:
+            portfolio_data = await self.get_portfolio_details(portfolio["key"])
+            yield [portfolio_data]
+            yield self._get_portfolio_subportfolios(portfolio_data)
```
PeyGis, line 506: can we add some try catch here so that in the event there's an unexpected error, we can catch it?

PeyGis: also, can we fetch all the portfolio details at once using async gather, and then take our time to extra it's sub portfolio data recursively? 

```
    portfolio_keys = [portfolio["key"] for portfolio in portfolios]
    portfolio_details = await asyncio.gather(
        *[self.get_portfolio_details(key) for key in portfolio_keys]
    )
```
for each portfolio details list
  yield portfolio
  yield extract sub portfolio( portfolio)

PeyGis (changes requested):
Overall looks great. Just some few comment

integrations/sonarqube/client.py
```diff
@@ -22,10 +22,14 @@ class Endpoints:
     ONPREM_ISSUES = "issues/list"
     SAAS_ISSUES = "issues/search"
     ANALYSIS = "activity_feed/list"
+    PORTFOLIO_DETAILS = "views/show"
+    PORTFOLIOS = "views/list"
 
 
 PAGE_SIZE = 100
 
+PORTFOLIO_VIEW_QUALIFIERS = ["VW", "SVW"]
```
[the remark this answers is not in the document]
lordsarcastic: Yes, only views of these types are actually portfolios

integrations/sonarqube/client.py
lordsarcastic: It is intentional. We encountered a bug recently on Kubecost integration where in the event a key has an explicit value of `None` it will return that instead of returning the default set in the `get` method call.

PeyGis: But the whole purpose of getting the sub views is just so we can iteratively return it's details. So portfolio.get("subViews", []) will ensure that if the subViews is None, we just return an empty list. the additional [] seems redundant

PeyGis (approved):
LGTM

integrations/sonarqube/client.py
```diff
@@ -468,6 +472,49 @@ async def get_all_sonarqube_analyses(
...
+            all_portfolios.extend(self._extract_subportfolios(subportfolio))
+        return all_portfolios
+
+    async def get_all_portfolios(self) -> AsyncGenerator[list[dict[str, Any]], None]:
+        logger.info(f"Fetching all portfolios in organization: {self.organization_id}")
+        portfolios = await self._get_all_portfolios()
+
+        try:
+            portfolio_keys = [portfolio["key"] for portfolio in portfolios]
+            portfolios_data = await asyncio.gather(
```
matan84, line 506: Let's batch concurrent requests to prevent the integration from running possibly thousands of tasks concurrently

matan84 (dismissed):
Added comments

erikzaadi (approved):

PR #997, opened 2024-09-06
[Integration][GCP] Added an alternative for retrieving quota project id

mk-armah:
# Description

**What**  
This PR fixes a bug related to fetching the GCP project ID from service account credentials. Previously, the function would fail to retrieve the correct project ID when the `quota_project_id` key was missing from the credentials file.

**Why**  
GCP configuration differs based on how the credentials are generated. When using `gcloud auth application-default login`, the project ID is stored under the `quota_project_id` key. However, when a service account key is generated from the console, the project ID is stored under the `project_id` key. This discrepancy caused the integration to fail in identifying the project ID when the credentials were generated from the console, impacting quota computation operations.

This bug affected users by causing failures in environments where `quota_project_id` was not available. The fix ensures that both configurations are supported, making the application more robust and reliable in different environments.

**How**  
- Introduced a mechanism to check for both `project_id` and `quota_project_id` keys in the credentials file.
- Improved error handling to raise meaningful exceptions when credentials are missing or incomplete.

## Type of change

- [ ] Docs PR link [here](#)


integrations/gcp/gcp_core/utils.py
```diff
@@ -101,7 +101,13 @@ def get_service_account_project_id() -> str:
     "get project id associated with service account"
     try:
         default_credentials = json.loads(get_credentials_json())
-        project_id = default_credentials["quota_project_id"]
+        project_id = default_credentials.get("project_id_") or default_credentials.get(
+            "quota_project_id_"
+        )
+
+        if not project_id:
+            raise KeyError("project_id or quota_project_id")
```
matan84, line 109: Change the phrasing here

integrations/gcp/gcp_core/utils.py
```diff
@@ -101,7 +101,13 @@ def get_service_account_project_id() -> str:
     "get project id associated with service account"
     try:
         default_credentials = json.loads(get_credentials_json())
-        project_id = default_credentials["quota_project_id"]
+        project_id = default_credentials.get("project_id_") or default_credentials.get(
```
matan84, line 104:
```suggestion
        project_id = default_credentials.get("project_id") or default_credentials.get("quota_project_id")
```

integrations/gcp/gcp_core/utils.py
```diff
@@ -119,7 +125,6 @@ def get_service_account_project_id() -> str:
         raise ValueError(
             f"Couldn't figure out the service account's project id. Error: {str(e)}"
         )
-    raise ValueError("Couldn't figure out the service account's project id.")
```
matan84, line 122: Don't delete this

matan84 (changes requested):
requested changes

integrations/gcp/gcp_core/utils.py
mk-armah: It can be reached ?

matan84: Just in case, so that we're not running into some None issues

integrations/gcp/gcp_core/utils.py
matan84: I meant to something amongst the lines of:
`Couldn't find keys project_id or quota_project_id in your credentials file.`

integrations/gcp/gcp_core/utils.py
mk-armah: we have except Exception 😂

integrations/gcp/gcp_core/utils.py
mk-armah: okay, got you, the reason why its this way is because, line 122, add the prefix to the error string

mk-armah: Eventually, this is what we get when a key error is raise

`Couldn't figure out the service account's project id. Key: project_id or quota_project_id doesn't exist in the credentials file."`

mk-armah: image

matan84: OHHH nice of nvm you can leave `project_id or quota_project_id`

PR #1018, opened 2024-09-15
Add core smoke test

erikzaadi:
# Description

Add a smoke test that uses the fake integration to connect to a live port instance and verify that entities arrive properly

### Content

- [x] Git ignore junit files
- [x] Update docker image labels to ensure new ghcr image permissions - according to [this Github policy](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility)
- [x] Add base bash port client (Will be ported to other languages)
- [x] Add smoke test that runs the fake integration in a container and assert on created entities in port
- [x] Add github action for core changes (that triggers smoke test)

## Type of change


scripts/run-smoke-test.sh
```diff
@@ -0,0 +1,104 @@
...
+
+SCRIPT_BASE="$(cd -P "$(dirname "$0")" && pwd)"
+ROOT_DIR="$(cd -P "${SCRIPT_BASE}/../" && pwd)"
+
+RANDOM_ID="${RANDOM}${RANDOM}" # DOGE - RANDOM
+INTEGRATION_IDENTIFIER="smoke-test-integration-${RANDOM_ID}"
+BLUEPRINT_DEPARTMENT="fake-department-${RANDOM_ID}"
+BLUEPRINT_PERSON="fake-person-${RANDOM_ID}"
+BLUEPRINT_IDENTIFIERS="${BLUEPRINT_DEPARTMENT} ${BLUEPRINT_PERSON}"
+FAKE_INTEGRATION_VERSION=${FAKE_INTEGRATION_VERSION:-0.1.1-dev}
```
Tankilevitch, line 18: what is the effect if its static dev version?

scripts/run-smoke-test.sh
```diff
@@ -0,0 +1,104 @@
...
+PORT_BASE_URL_FOR_DOCKER=${PORT_BASE_URL}
+
+if [[ ${PORT_BASE_URL} =~ localhost ]]; then
+    PORT_BASE_URL_FOR_DOCKER=${PORT_BASE_URL//localhost/host.docker.internal}
```
Tankilevitch, line 22: please elaborate 🙏

scripts/run-smoke-test.sh
```diff
@@ -0,0 +1,104 @@
...
+fi
+
+TEMP_DIR=$(mktemp -d -t smoke-test-integration.XXXXXXX)
```
Tankilevitch, line 25: what is this for?

scripts/run-smoke-test.sh
Tankilevitch, line 1: overall i feel like there are a lot of comments missing to make this readable

scripts/run-smoke-test.sh
```diff
@@ -0,0 +1,104 @@
...
+RESOURCE_DIR_SUFFIX="integrations/fake-integration/.port/resources"
+cp -r "${ROOT_DIR}"/${RESOURCE_DIR_SUFFIX} "${TEMP_DIR}"
+
+sed -i.bak "s/fake-department/${BLUEPRINT_DEPARTMENT}/g" "${TEMP_DIR}"/resources/blueprints.json
+sed -i.bak "s/fake-person/${BLUEPRINT_PERSON}/g" "${TEMP_DIR}"/resources/blueprints.json
+sed -i.bak "s/\"fake-department\"/\"${BLUEPRINT_DEPARTMENT}\"/g" "${TEMP_DIR}"/resources/port-app-config.yml
+sed -i.bak "s/\"fake-person\"/\"${BLUEPRINT_PERSON}\"/g" "${TEMP_DIR}"/resources/port-app-config.yml
```
Tankilevitch, line 32: should we rename the file? as its very very specific for fake-integration?

integrations/_infra/Dockerfile
```diff
@@ -4,6 +4,7 @@ ARG BUILD_CONTEXT
 ARG INTEGRATION_VERSION
 
 LABEL INTEGRATION_VERSION=${INTEGRATION_VERSION}
+LABEL org.opencontainers.image.source https://github.com/port-labs/ocean
```
Tankilevitch, line 7: add comment for reasoning

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,58 @@
+name: Smoke Test
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  detect-changes:
+    uses: ./.github/workflows/detect-changes-matrix.yml
```
Tankilevitch, line 9: shouldn't this be the opposite? as the smoke tests is for changes in core rather than in integrations

Tankilevitch (changes requested):

scripts/run-smoke-test.sh
erikzaadi: It's possible to override if wanted (will take defined env var or the default value)

scripts/run-smoke-test.sh
erikzaadi: TL;DR to make it work on docker locally while accessing port api on a local running process, you need to use 'host.docker.internal' within the container.

scripts/run-smoke-test.sh
erikzaadi: This is where I save the altered `blueprints.json` and `app-config-port.yml`

erikzaadi: it's also mounted to the container

scripts/run-smoke-test.sh
erikzaadi: it's hard code for the fake integration at this delivery, can work on make it generic now or as a separate PR

.github/workflows/smoke-test.yml
erikzaadi: This is me making a small AC preparation crime.
Wanted to create a convention where you'd add something according to convention to run the same smoke test for an integration later.

The detect changes also detects changes in core

integrations/_infra/Dockerfile
erikzaadi: Done

scripts/run-smoke-test.sh
erikzaadi: Added some comments and tried to simplify where possible, please have a looksie

erikzaadi:
@Tankilevitch made some more changes for readability

scripts/run-smoke-test.sh
erikzaadi: `mv {bash,python}`

Tankilevitch (approved):
LGTM, nice smoke test ;)

PR #1019, opened 2024-09-16
Ocean Gitlab V2 Integration 

MusahMusah:
# Description

What - Implemented Gitlab Integration

Why - The goal is to ensure that the integration is efficient, intuitive, and user-friendly, while minimizing dependencies. This enables better control over rate limits, supports real-time events, and allows for easier maintenance and testing.

How - By building an async HTTP client using Ocean's custom implementation, integrating rate-limiting mechanisms, and streamlining the setup process to make parameters more intuitive. Webhook support and real-time event handling will enhance user interaction, and verbosity and testing ensure clarity and reliability throughout development.

## Type of change

I implemented my test cases by simulating those from integrations/fake-integration. However, when I run the tests, I consistently encounter the following error:

`FAILED tests/test_integration.py::test_all_resync_methods - pydantic.error_wrappers.ValidationError: 1 validation error for IntegrationConfiguration
unhashable type: 'list' (type=type_error)
`

Due to this error, I have temporarily commented out my test code [here](https://github.com/port-labs/ocean/pull/1019/files#diff-943497e1611996d1bd50717ddd79696a8f9c2b8d3bf03190177d75344f0199f1). I believe the issue may stem from the Ocean framework itself, which is why I have not removed the tests entirely. The commented-out code is retained for reference and future debugging.

Note: The test was passing in a previous version of Ocean that I had cloned locally. I suspect the error, which arises from invoking Ocean's `get_raw_result_on_integration_sync_kinds` test utility function, may have been introduced in an earlier version of the framework.

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [x] Support Multi account

## Screenshots
image
image
image
image

## API Documentation
Here are the GitLab REST API endpoints that I referenced for my integrations:
- [Gitlab Projects](https://docs.gitlab.com/ee/api/projects.html)
- [Gitlab Issues](https://docs.gitlab.com/ee/api/issues.html)
- [Gitlab Groups](https://docs.gitlab.com/ee/api/groups.html)
- [Gitlab Merge Requests](https://docs.gitlab.com/ee/api/merge_requests.html)
  
**WEBHOOKS**
- [Gitlab Webhook Events](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#issue-events)

**RATE LIMITS**
- [User Rate limit](https://docs.gitlab.com/ee/administration/settings/user_and_ip_rate_limits.html)

**PAGINATION**
- [Offset Based Pagination](https://docs.gitlab.com/ee/api/rest/#pagination)


integrations/gitlab_v2/client.py
```diff
@@ -0,0 +1,296 @@
...
+    "approved",
+    "unapproved",
+    "approval",
+    "unapproval",
+]
+DELETE_WEBHOOK_EVENTS: list[str] = ["close", "merge"]
+WEBHOOK_EVENTS_TO_TRACK: dict[str, bool] = {
+    "issues_events": True,
+    "merge_requests_events": True,
+}
```
PeyGis, line 27: gitlab has more webhook events. please track all of them instead of these 2

integrations/gitlab_v2/client.py
```diff
@@ -0,0 +1,296 @@
...
+WEBHOOK_NAME: str = "Port-Ocean-Events-Webhook"
+
+
+class GitlabClient:
+    def __init__(self, gitlab_host: str, gitlab_token: str) -> None:
+        self.projects_url = f"{gitlab_host}/api/v4/projects"
+        self.merge_requests_url = f"{gitlab_host}/api/v4/merge_requests"
+        self.issues_url = f"{gitlab_host}/api/v4/issues"
+        self.groups_url = f"{gitlab_host}/api/v4/groups"
+
```
PeyGis, line 37: This is going to cause complexity in the future if we keep adding more kinds. Just have one single url and that's what you should pass to the async http client

integrations/gitlab_v2/integration.py
```diff
@@ -0,0 +1,32 @@
...
+    Selector,
+)
+from port_ocean.core.integrations.base import BaseIntegration
+from pydantic import Field
+
+
+class GitlabProjectSelector(Selector):
+    onlyGrouped: bool = Field(
+        default=True, description="Retrieve only grouped projects"
+    )
```
PeyGis, line 16: may I know you meant by grouped projects?

integrations/gitlab_v2/main.py
```diff
@@ -0,0 +1,157 @@
...
+class ResourceKind(StrEnum):
+    GROUP = "group"
+    PROJECT = "project"
+    MERGE_REQUEST = "merge_request"
+    ISSUE = "issue"
+
+
+class WebHookEventType(StrEnum):
+    MERGE_REQUEST = "merge_request"
+    ISSUE = "issue"
```
PeyGis, line 28: add more webhook types

integrations/gitlab_v2/main.py
```diff
@@ -0,0 +1,157 @@
...
+    # Check if tokens is a list and filter valid tokens (strings only)
+    if isinstance(tokens, list):
+        tokens_are_valid = filter(lambda token: isinstance(token, str), tokens)
+
+        # Ensure all tokens are valid strings
+        if not all(tokens_are_valid):
+            raise InvalidTokenException("Invalid access tokens, ensure all tokens are valid strings")
+    else:
+        raise InvalidTokenException("Invalid access tokens, confirm you passed in a list of tokens")
+
```
PeyGis, line 46: the validation of tokens should be something that should be handled by GitLab and not us

integrations/gitlab_v2/main.py
```diff
@@ -0,0 +1,157 @@
...
+    return await bootstrap_client()
+
+
+def initialize_client(gitlab_access_token: str) -> GitlabClient:
+    return GitlabClient(
+        ocean.integration_config["gitlab_host"],
+        gitlab_access_token,
+    )
```
PeyGis, line 54: why do you pass the access token to this function? why can't you do the same as the gitlab_host?

integrations/gitlab_v2/main.py
```diff
@@ -0,0 +1,157 @@
...
+            "No app host provided, skipping webhook creation. "
+            "Without setting up the webhook, the integration will not export live changes from Gitlab"
+        )
+        return
+
+    if ocean.event_listener_type == "ONCE":
+        logger.info("Skipping webhook creation because the event listener is ONCE")
+        return
+
+    gitlab_client = initialize_client(tokens[0])
```
PeyGis, line 71: can you share how the token is passed to the integration?

integrations/gitlab_v2/main.py
```diff
@@ -0,0 +1,157 @@
...
+    if webhook_event == WebHookEventType.MERGE_REQUEST:
+        payload = extract_merge_request_payload(data)
+        await ocean_action(ResourceKind.MERGE_REQUEST, [payload])
+    elif webhook_event == WebHookEventType.ISSUE:
+        payload = extract_issue_payload(data)
+        logger.info(f"Upserting issue with payload: {payload}")
+        await ocean_action(ResourceKind.ISSUE, [payload])
+    else:
+        logger.info(f"Unhandled webhook event type: {webhook_event}")
+        return {"ok": True}
```
PeyGis, line 99: handle the other webhook events

integrations/gitlab_v2/main.py
```diff
@@ -0,0 +1,157 @@
...
+async def resync_project(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    tokens = ocean.integration_config["gitlab_access_tokens"]["tokens"]
+    for token in tokens:
+        client = initialize_client(token)
+        config = typing.cast(GitlabProjectResourceConfig, event.resource_config)
+
+        async for projects in client.get_projects():
+            logger.info(f"Received {kind} batch with {len(projects)} projects")
+            if config.selector.onlyGrouped:
+                projects = [project for project in projects if project.get("__group")]
```
PeyGis, line 126: I don't see what this filter is solving?

integrations/gitlab_v2/tests/test_integration.py
```diff
@@ -0,0 +1,88 @@
...
+# ]
+#
+# FAKE_GROUPS: list[dict[str, Any]] = [
+#     {
+#         "id": 1,
+#         "title": "Test Group",
+#         "visibility": "private",
+#         "web_url": "https://gitlab.com/test-namespace/test-group",
+#         "description": "Group description",
+#     }
```
PeyGis, line 32: everything here is commented. if you don't need it, delete them

integrations/gitlab_v2/utils.py
```diff
@@ -0,0 +1,39 @@
...
+def extract_issue_payload(data: dict[str, Any]) -> dict[str, Any]:
+    logger.info(f"Extracting issue for project: {data['project']['id']}")
+    return {
+        "id": data["object_attributes"]["id"],
+        "title": data["object_attributes"]["title"],
+        "link": data["object_attributes"]["url"],
+        "description": data["object_attributes"]["description"],
+        "created_at": data["object_attributes"]["created_at"],
+        "updated_at": data["object_attributes"]["updated_at"],
+        "author": {
```
PeyGis, line 33: this should be done using jq in the port-app-config.yaml

PeyGis (changes requested):
looks good. please go over my comments

integrations/gitlab_v2/client.py
MusahMusah: Okay noted

integrations/gitlab_v2/utils.py
MusahMusah: alright then.

integrations/gitlab_v2/integration.py
MusahMusah: The intension was to give users ability to filter only projects that belongs to atleast one group or projects that have existing relationship with group(s).

integrations/gitlab_v2/main.py
MusahMusah: The intent is to support multiple token implementation from different GitLab accounts, allowing for a seamless resync of all resources at once, rather than requiring multiple integrations. However, I can revert to supporting a single token if this approach does not align with the intended requirements. Please let me know if i should stick to just single access token for now.

MusahMusah: The validation only ensures tokens are provided in below format :
`{"tokens": ["token1", "token2"]}`

integrations/gitlab_v2/main.py
MusahMusah: The tokens can be passed as an environment variable 
`OCEAN__INTEGRATION__CONFIG__GITLAB_ACCESS_TOKENS={"tokens": ["token1", "token2"]}`

Also validated in the `spec.yaml` like so:

> configurations:
  - name: gitlabAccessTokens
    required: true
    type: object
    sensitive: true
    description: 'Gitlab access token. See the <a target="_blank" href="https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html">Gitlab Documentation</a>'

integrations/gitlab_v2/tests/test_integration.py
MusahMusah: I commented this out because:

I implemented my test cases by simulating those from integrations/fake-integration. However, when I run the tests, I consistently encounter the following error:
FAILED tests/test_integration.py::test_all_resync_methods - pydantic.error_wrappers.ValidationError: 1 validation error for IntegrationConfiguration unhashable type: 'list' (type=type_error) 

Note: The test was passing in a previous version of Ocean that I had cloned locally. I suspect the error, which arises from invoking Ocean's get_raw_result_on_integration_sync_kinds test utility function, may have been introduced in an earlier version of the framework.

integrations/gitlab_v2/main.py
MusahMusah: The access token is passed to the function because it can vary dynamically depending on the request, as my implementation supports multiple access tokens.  This allows the flexibility to switch between tokens as needed. In contrast, the GitLab host is a constant value that doesn’t change, so it's directly accessed from the configuration rather than passed as a parameter, making the token the only part that requires flexibility during initialization.

MusahMusah: Please let me know if there is no need to support multiple access tokens

integrations/gitlab_v2/main.py
MusahMusah: Issue has been addressed and this filter has been removed as my implementation now works with a single resync method as suggested by other reviewer.

integrations/gitlab_v2/main.py
MusahMusah: I'm now handling "push", "merge_request", and "issue" events. These are the events associated to the resources given. Kindly specify which other events you would like my implementation to handle as well. thanks so much.

MusahMusah:
Hi @PeyGis

Thank you for taking the time to review my PR. I have addressed most of the issues you raised and left a few comments for clarification. I'm looking forward to your feedback on those.

integrations/gitlab_v2/main.py
PeyGis: sure. But ideally, when users provide tokens, we want it to be in this format
{"access-token-1": "list of projects or sub groups it should pull data from, you can also use * so it brings all", "access-token-2": "list of projects or sub groups it should pull data from", "access-token-X": "list of xxx"}

This way, when the user installs the integration, it will fetch data from all the tokens that have been provided

PeyGis: https://docs.getport.io/build-your-software-catalog/sync-data-to-catalog/git/gitlab/installation#tokenmapping

You can read more here

integrations/gitlab_v2/main.py
PeyGis: supporting multiple tokens is one of the main focus of this assignment so yeah, good job on that. I was only concerned about the way you've structured it

integrations/gitlab_v2/main.py
MusahMusah: I see now, thanks for clarifying this.

PR #1041, opened 2024-09-20
[Integration][AWS] - Fix ExpiredTokenException 

mk-armah:
# Description

**What**  
- Improved the mechanism for parallel fetching of AWS account resources.
- Fixed `ExpiredTokenException` by replacing the event-based caching system with a time-dependent caching mechanism. The new approach ensures that the role is reassumed and session credentials are refreshed when 80% of the session duration has been used.

**Why**  
- The previous event-based caching system led to the `ExpiredTokenException`, causing session credentials to expire unexpectedly.
- Implementing a time-dependent caching mechanism ensures that session credentials are refreshed proactively, preventing disruptions.

**How**  
- Replaced the resync-dependent caching system with a time-based cache that monitors the session expiry.
- Added logic to reassume the role and refresh credentials once 80% of the session duration has passed, improving session reliability.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/aws/utils/aws.py
```diff
@@ -1,33 +1,32 @@
...
+
 _session_manager: SessionManager = SessionManager()
 
+CACHE_DURATION_SECONDS = (
+    0.80 * ASSUME_ROLE_DURATION_SECONDS
+)  # Refresh role credentials after exhausting 80% of the session duration
+
 
-async def update_available_access_credentials() -> None:
+@cached(ttl=CACHE_DURATION_SECONDS, cache=Cache.MEMORY)
```
PeyGis, line 19: I would like to understand what happens when the cache is cleared after the duration? And how is the simple Memory different from using FastAPIs event.attribute functionality?

i guess my question is, since you changed the approach from event based to time based, what happens to the assumed role credentials after the ttl?

Also after the cache is cleared, how does it manage the re-entry. Thus, after the first ttl duration has elapsed, we signal that the credentials needs to be refreshed. Now what happens after the second hour. how does the ttl behave?

PeyGis (changes requested):
good job on using the sessionDuration in the assume role client. I left a comment on how the aiocache works for the ttl

integrations/aws/aws/session_manager.py
```diff
@@ -93,7 +96,9 @@ async def _get_organization_session(self) -> aioboto3.Session | None:
         async with application_session.client("sts") as sts_client:
             try:
                 organizations_client = await sts_client.assume_role(
-                    RoleArn=organization_role_arn, RoleSessionName="AssumeRoleSession"
+                    RoleArn=organization_role_arn,
+                    RoleSessionName="AssumeRoleSession",
```
matan84, line 100:
```suggestion
                    RoleSessionName="OceanAssumeRoleSession",
```

integrations/aws/aws/session_manager.py
```diff
@@ -151,6 +156,7 @@ async def _assume_role_and_update_credentials(
             account_role = await sts_client.assume_role(
                 RoleArn=f'arn:aws:iam::{account["Id"]}:role/{self._get_account_read_role_name()}',
                 RoleSessionName="AssumeRoleSession",
```
matan84, line 158:
```suggestion
                RoleSessionName="OceanMemberAssumeRoleSession",
```

integrations/aws/main.py
```diff
@@ -35,6 +35,10 @@
     is_access_denied_exception,
     is_server_error,
 )
+from port_ocean.utils.async_iterators import stream_async_iterators_tasks
+import asyncio
+
+MAX_CONCURRENT_TASKS = 50
```
matan84, line 41: Why is this the limit?

integrations/aws/main.py
```diff
@@ -69,19 +73,31 @@ async def _handle_global_resource_resync(
...
+        async with semaphore:
+            if is_global:
+                async for batch in _handle_global_resource_resync(kind, credentials):
+                    yield batch
+            else:
+                async for session in credentials.create_session_for_each_region():
+                    try:
+                        async for batch in resync_cloudcontrol(kind, session):
+                            yield batch
+                    except Exception:
```
matan84, line 94: I know you didn't add this except Exception, but let's remove it. I think this can be really problematic. If you think otherwise, let's talk

matan84 (changes requested):
Added some comments. Please make sure to add tests

integrations/aws/main.py
Tankilevitch: lets catch the error, and raise at the end if any exception was found, that way we still fetch for all other accounts without failing them.

integrations/aws/aws/session_manager.py
Tankilevitch: is it breaking?

integrations/aws/utils/aws.py
mk-armah: Cache is meant to expire after exhausting 80% of the session duration;
when the cache is cleared, the next call to `update_available_access_credentials` resets the session, producing new credentials thereby extending the expiry time once again...
This can happen at any point within a resync.

TTL, last for the specified time only, `update_available_access_credentials` is being called as many times as possible to ensure we know whenever the TTL is close to expiry.

integrations/aws/main.py
mk-armah: Controlling the number of accounts with the stream at a particular time ?

matan84: I meant, Why did you pick 50 for the max concurrent tasks? is there a doc you are basing on? if yes - let's add it as comment

integrations/aws/utils/aws.py
mk-armah: Right, I understand, actually @Tankilevitch  and I decided on this value intuitively control the stream

integrations/aws/main.py
mk-armah: Right, I understand, actually @Tankilevitch  and I decided on this value intuitively control the stream

mk-armah: what do you think ?

matan84: Ok, understood.

integrations/aws/main.py
```diff
@@ -69,19 +73,37 @@ async def _handle_global_resource_resync(
...
+                        regions.append(session.region_name)
+                        errors.append(exc)
+                        continue
+            if errors:
+                message = f"Failed to fetch {kind} for these regions {regions} with {len(errors)} errors in account {credentials.account_id}"
+                raise ExceptionGroup(message, errors)
+
+    tasks = [handle_account(credentials) async for credentials in get_accounts()]
+    if tasks:
+        async for batch in stream_async_iterators_tasks(*tasks):
```
matan84, line 105: Let's batch processes together, I think there's no limitaiton here (i.e. what if we have 20,000 accounts?)

integrations/aws/main.py
```diff
@@ -69,19 +73,37 @@ async def _handle_global_resource_resync(
...
-            async for batch in _handle_global_resource_resync(kind, credentials):
-                yield batch
-        else:
-            async for session in credentials.create_session_for_each_region():
-                try:
-                    async for batch in resync_cloudcontrol(kind, session):
-                        yield batch
-                except Exception:
-                    continue
+    semaphore = asyncio.BoundedSemaphore(MAX_CONCURRENT_TASKS)
```
matan84, line 79: Is there anywhere else you can initiate this semaphore?

integrations/aws/main.py
```diff
@@ -69,19 +73,37 @@ async def _handle_global_resource_resync(
...
+
+    async def handle_account(
```
matan84, line 81: Extract this function.

matan84: The main purpose of the resync_all is to control the rate and the parallelism of the handle_account. Also, rename this handle account to something more verbose.

integrations/aws/main.py
```diff
@@ -94,76 +116,102 @@ async def resync_account(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
             "elasticache",
             "describe_cache_clusters",
             "CacheClusters",
             "Marker",
-        ):
+        )
+        async for session in get_sessions()
+    ]
+    if tasks:
+        async for batch in stream_async_iterators_tasks(*tasks):
```
matan84, line 132: Again, place some limitation on the parallelism here

matan84 (changes requested):
Added some more comments

integrations/aws/main.py
mk-armah: can you come again, this isn't clear to me ?

integrations/aws/main.py
mk-armah: the object isn't meant to be used by another function unless re-instantiated for some of use case, in my opinion that's the best place for it ?

mk-armah: we can also instantiate somewhere else, but the point is, a single instance or object wouldn't be reusable for all other functions.

matan84 (approved):
LGTM

PR #1063, opened 2024-10-06
[Integration][AWS] | Improved Concurrency Control and Eliminated Likelihood of Thundering Herd

mk-armah:
# Description

**What:**

- Refactored semaphore implementation to effectively limit concurrency across tasks.
- Added a util `semaphore_async_iterator` to enable seamless control over concurrent executions per kind (can be re-used in other integrations).
- Removed iterative calls to the cache for tracking token expiry, reducing the likelihood of a thundering herd problem.

**Why:**

- The previous implementation limited concurrency within tasks (accounts), ideally, this use case requires concurrent limits is to be global, thus across accounts.
- Iterative cache calls could potentially cause a thundering herd problem when cache expires and token needed to be refreshed.

**How:**

- Applied the semaphore correctly by wrapping task creation with the semaphore, ensuring proper concurrency control.
- Implemented unit tests using `pytest` to verify semaphore functionality and concurrency limits.
- Optimized cache usage by eliminating unnecessary iterative calls for tracking expiry.

## Type of change

### Core testing checklist

- [x] If new resource kind is added or updated in the integration, add example raw data, mapping, and expected result to the `examples` folder in the integration directory.
- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [x] Support Multi account

## Screenshots

NB: Warning log in the screenshot below is just for demonstration purposes.
image

## API Documentation


integrations/aws/utils/resources.py
```diff
@@ -127,7 +127,7 @@ async def resync_custom_kind(
     next_token = None
     if not describe_method_params:
         describe_method_params = {}
-    while await update_available_access_credentials():
```
omby8888, line 130: ?

mk-armah: While it enhances the frequency at which we check for the token's existence, when it detects that the token isn't available, it causes a thundering herd when the `resync_custom_kind` function is executed concurrently.

omby8888 (approved):
Nice Work!! 🌊

PR #1105, opened 2024-10-29
[Integration][SonarQube] Bug | 404 response when querying SonarQube issues

lordsarcastic:
# Description

What - Fixed a bug where SonarQube instances lower than 10.2 report a 404 error when retrieving issues

Why - This is because the issues list API on SonarQube was introduced in 10.2 and also as an internal API

How - Checking for the integration version and choosing the right API based on the integration version

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots
image

## API Documentation

[SonarQube issues list API](https://next.sonarqube.com/sonarqube/web_api/api/issues/list?internal=true)

### DIFF
--- a/integrations/sonarqube/client.py
+++ b/integrations/sonarqube/client.py
@@ -46,7 +46,6 @@ class Endpoints:
 
 PAGE_SIZE = 100
 
-
 PORTFOLIO_VIEW_QUALIFIERS = ["VW", "SVW"]
 
 
--- a/integrations/sonarqube/client.py
+++ b/integrations/sonarqube/client.py
@@ -38,14 +38,17 @@ class Endpoints:
     WEBHOOKS = "webhooks"
     MEASURES = "measures/component"
     BRANCHES = "project_branches/list"
-    ONPREM_ISSUES = "issues/list"
-    SAAS_ISSUES = "issues/search"
+    ISSUES_LIST = "issues/list"
+    ISSUES_SEARCH = "issues/search"
     ANALYSIS = "activity_feed/list"
     PORTFOLIO_DETAILS = "views/show"
     PORTFOLIOS = "views/list"
+    VERSION = "server/version"
 
 
 PAGE_SIZE = 100
+
+MINIMUM_SUPPORTED_ISSUES_LIST_VERSION = 10.2
 
 PORTFOLIO_VIEW_QUALIFIERS = ["VW", "SVW"]
 
@@ -88,6 +91,15 @@ class SonarQubeClient:
                 "Content-Type": "application/json",
             },
         }
+
+    async def _get_sonarqube_instance_version(self) -> str:
+        response = await self.http_client.get(
+            f"{self.base_url}/api/{Endpoints.VERSION}"
+        )
+        response.raise_for_status()
+        semver_list = response.text.split(".")
+        version = float(f"{semver_list[0]}.{semver_list[1]}")
+        return version
 
     async def send_api_request(
         self,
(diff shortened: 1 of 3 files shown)


integrations/sonarqube/client.py
```diff
@@ -338,12 +350,20 @@ async def get_issues_by_component(
         component_key = component.get("key")
         endpoint_path = ""
 
-        if self.is_onpremise:
+        current_version = await self._get_sonarqube_instance_version()
+
+        if current_version >= MINIMUM_SUPPORTED_ISSUES_LIST_VERSION:
+            logger.info(
+                f"SonarQube version is {current_version}. Using issues/list API"
```
PeyGis, line 357:
```suggestion
            logger.info(
                f"SonarQube version is {current_version}. Using {Endpoints.ISSUE_LIST} API"
```

integrations/sonarqube/client.py
```diff
@@ -338,12 +350,20 @@ async def get_issues_by_component(
         component_key = component.get("key")
         endpoint_path = ""
 
-        if self.is_onpremise:
+        current_version = await self._get_sonarqube_instance_version()
+
+        if current_version >= MINIMUM_SUPPORTED_ISSUES_LIST_VERSION:
```
PeyGis, line 355: Can you verify if this logic will work for SonarCloud too? From the way I see it, they don't have issues/list, only issues/search

https://sonarcloud.io/web_api/api/issues/search?deprecated=false

PeyGis (changes requested):
looks great. left some comments

integrations/sonarqube/client.py
```diff
@@ -336,20 +337,13 @@ async def get_issues_by_component(
...
         component_key = component.get("key")
-        endpoint_path = ""
-
-        if self.is_onpremise:
-            query_params = {"project": component_key}
-            endpoint_path = Endpoints.ONPREM_ISSUES
-        else:
-            query_params = {"componentKeys": component_key}
-            endpoint_path = Endpoints.SAAS_ISSUES
+        query_params = {"componentKeys": component_key}
```
PeyGis, line 340:
```suggestion
        if self.is_onpremise:
            query_params = {"components": component_key}
        else:
            query_params = {"componentKeys": component_key}
```

integrations/sonarqube/client.py
```diff
@@ -38,15 +38,16 @@ class Endpoints:
     WEBHOOKS = "webhooks"
     MEASURES = "measures/component"
     BRANCHES = "project_branches/list"
-    ONPREM_ISSUES = "issues/list"
-    SAAS_ISSUES = "issues/search"
+    ISSUES_SEARCH = "issues/search"
     ANALYSIS = "activity_feed/list"
     PORTFOLIO_DETAILS = "views/show"
     PORTFOLIOS = "views/list"
+    VERSION = "server/version"
```
PeyGis, line 45: delete since we don't need it

integrations/sonarqube/client.py
```diff
@@ -38,15 +38,16 @@ class Endpoints:
...
 
 
 PAGE_SIZE = 100
 
+
```
PeyGis, line 50: remove blank line

PeyGis (changes requested):
left some comments and suggestions

integrations/sonarqube/client.py
lordsarcastic: Only components since we are using only one API

Tankilevitch (approved):

PR #1121, opened 2024-11-05
[Integration][async-gitlab] - <GitLab integration with Async support>

victor-devv:
# Description

What - Ocean Gitlab integration with async support using the http async client. Includes support for resources which includes groups, projects, merge requests and issues; as well as webhooks.

Why -

How -

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/async-gitlab/.port/resources/blueprints.json
```diff
@@ -0,0 +1,228 @@
...
+            }
+          }
+        },
+        "required": []
+      },
+      "mirrorProperties": {},
+      "calculationProperties": {
+        "lead_time_days": {
+          "title": "Lead time (Days)",
+          "calculation": "if .properties.status == \"merged\" then ((.properties.mergedAt | sub(\"\\\\.[0-9]+\\\\+00:00$\"; \"Z\") | strptime(\"%Y-%m-%dT%H:%M:%SZ\") | mktime) - (.properties.createdAt | sub(\"\\\\.[0-9]+\\\\+00:00$\"; \"Z\") | strptime(\"%Y-%m-%dT%H:%M:%SZ\") | mktime)) / 86400 | tonumber else null end",
```
kodjomiles, line 152: why is lead time a calculation property here, The calculation should be done in the mapping configuration and default to null if the pr hasn't been merged. we have access to both mergedAt and createdAt at the time of ingesting. this would help the reuse of leadTime. calculation properties are done on the fly. there are instances that we need to use leadTime.

PR #1146, opened 2024-11-15
[Integration][SonarQube] Added support for querying projects via the GA API

lordsarcastic:
# Description

What - Changed projects API from internal to GA and made use of internal API optional. Add more logs to SonarQube integration

Why - To aid investigations of bugs in the SonarQube integration and to avoid reliance on internal and unstable APIs

How - Added selector property to enable use of internal API to enrich more data. Added more details to the logs and keeping them at the debug level

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/sonarqube/client.py
```diff
@@ -95,8 +96,11 @@ async def send_api_request(
         query_params: Optional[dict[str, Any]] = None,
         json_data: Optional[dict[str, Any]] = None,
     ) -> dict[str, Any]:
+        _headers = {**self.http_client.headers, "Authorization": "Bearer <masked>"}
```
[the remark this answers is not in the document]
lordsarcastic: Part of requirements to log as many information as possible including tokens but the tokens should be masked

integrations/sonarqube/CHANGELOG.md
```diff
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Increased logs presence in integration (0.1.111)
+- Replaced calls to internal API for projects to GA version, making internal API use optional (0.1.111)
+
+### Bug Fixes
+
+- Fixed bug where issues/list API is not available for older SonarQube instance versions (0.1.105)
+
+
+## 0.1.111 (2024-11-15)
+
```
PeyGis, line 24: I thought we fixed this already in earlier release. Can we update the changelog so it focuses on this project API transition and enhanced logs?

integrations/sonarqube/client.py
```diff
@@ -206,7 +217,7 @@ async def get_components(
 
         try:
             response = await self.send_paginated_api_request(
-                endpoint=Endpoints.PROJECTS,
+                endpoint=Endpoints.PROJECTS_INTERNAL,
```
PeyGis, line 220: let's use the GA project API here for the get_components() and that's what will be used by new customer who install the integration

integrations/sonarqube/integration.py
```diff
@@ -184,6 +184,11 @@ def default_metrics() -> list[str]:
         metrics: list[str] = Field(
             description="List of metric keys", default=default_metrics()
         )
+        use_internal_api: bool = Field(
+            alias="useInternalApi",
+            description="Use internal API to fetch more data",
+            default=True,
```
PeyGis, line 190: Let's set the default to False

integrations/sonarqube/client.py
```diff
@@ -288,12 +299,26 @@ async def get_all_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
...
+            components = await self.get_components()
+            for component in components:
+                all_projects[component["key"]] = {
+                    **all_projects.get(component["key"], {}),
+                    **component,
+                }
+
+        for project in all_projects.values():
+            project_data = await self.get_single_project(project=project)
             yield [project_data]
```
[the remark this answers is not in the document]
mk-armah: There seemed to be an oversight here even with the already existing code. It was yielded one project at a time,  Lets correct that

integrations/sonarqube/client.py
```diff
@@ -288,12 +299,26 @@ async def get_all_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
...
+        if selector.use_internal_api:
+            components = await self.get_components()
+            for component in components:
+                all_projects[component["key"]] = {
+                    **all_projects.get(component["key"], {}),
+                    **component,
+                }
+
+        for project in all_projects.values():
+            project_data = await self.get_single_project(project=project)
```
mk-armah, line 321:
```suggestion
yield [await self.get_single_project(project=component) for component in components]
```

PeyGis (changes requested):
left some comments

mk-armah (changes requested):

integrations/sonarqube/client.py
mk-armah: pass endpoint as a parameter to `get_component` function, default to `PROJECTS`

integrations/sonarqube/client.py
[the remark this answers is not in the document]
lordsarcastic: The `get_single_project` method enriches the project with important data such as links, branches and so on.

integrations/sonarqube/client.py
```diff
@@ -288,13 +303,29 @@ async def get_all_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
...
+
+        all_projects = {}
+        for project in await self.send_paginated_api_request(
+            endpoint=Endpoints.PROJECTS, data_key="components"
+        ):
+            project_key = project.get("key")
+            all_projects[project_key] = project
+
+        if selector.use_internal_api:
+            # usint the internal API to fetch more data
```
Tankilevitch, line 317: typo

integrations/sonarqube/client.py
```diff
@@ -288,13 +303,29 @@ async def get_all_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
...
+            for component in components:
+                all_projects[component["key"]] = {
+                    **all_projects.get(component["key"], {}),
+                    **component,
+                }
+
+        yield [
+            await self.get_single_project(component)
+            for component in all_projects.values()
+        ]
```
Tankilevitch, line 328: I think we need pagination in batches and not aggregating all and then returning all

Tankilevitch (changes requested):

integrations/sonarqube/client.py
```diff
@@ -288,13 +303,29 @@ async def get_all_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
...
-        for component in components:
-            project_data = await self.get_single_project(project=component)
-            yield [project_data]
+        selector = cast(SonarQubeProjectResourceConfig, event.resource_config).selector
+        self.metrics = selector.metrics
+
+        all_projects = {}
+        for project in await self.send_paginated_api_request(
+            endpoint=Endpoints.PROJECTS, data_key="components"
+        ):
```
mk-armah, line 312: why not 
`for project in await self.get_components(Endpoints.PROJECTS)` ?

integrations/sonarqube/client.py
mk-armah, line 317:
```suggestion
            # using the internal API to fetch more data
```

integrations/sonarqube/client.py
Tankilevitch, line 328: this can be very prune to errors if one fails get_single_project it will fail all projects

integrations/sonarqube/client.py
```diff
@@ -288,13 +303,29 @@ async def get_all_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
...
+            for component in components:
+                all_projects[component["key"]] = {
+                    **all_projects.get(component["key"], {}),
+                    **component,
+                }
+
+        yield [
+            await self.get_single_project(component)
+            for component in all_projects.values()
+        ]
```
mk-armah, line 329: I'm not sure we need a yield here if all its doing is aggregating one time, return might be best in such cases.
Can we refactor to some pagination like  approach ? .

mk-armah (dismissed):
left some comments

integrations/sonarqube/integration.py
```diff
@@ -184,6 +184,11 @@ def default_metrics() -> list[str]:
         metrics: list[str] = Field(
             description="List of metric keys", default=default_metrics()
         )
+        use_internal_api: bool = Field(
+            alias="useInternalApi",
+            description="Use internal API to fetch more data",
+            default=False,
```
Tankilevitch, line 190: Should be true, to be backwards compatible 

And the default template mapping and blueprints should use the False option

mk-armah: @lordsarcastic

integrations/sonarqube/main.py
```diff
@@ -18,14 +25,73 @@ def init_sonar_client() -> SonarQubeClient:
...
+
+    if api_filter.on_provisioned_only:
+        project_params["onProvisionedOnly"] = api_filter.on_provisioned_only
+
+    if api_filter.projects:
+        project_params["projects"] = ",".join(api_filter.projects)
+
+    if api_filter.qualifiers:
+        project_params["qualifiers"] = ",".join(api_filter.qualifiers)
+
```
PeyGis, line 46: the SonarQubeProjectApiFilter  has a `generate_request_params()`, is there any reason why you're not reusing that method since you've even made the same logic there?

```
        if self.projects:
            value["projects"] = ",".join(self.projects)

        if self.qualifiers:
            value["qualifiers"] = ",".join(self.qualifiers)
```

integrations/sonarqube/main.py
```diff
@@ -18,14 +25,73 @@ def init_sonar_client() -> SonarQubeClient:
...
+    return project_params
+
+
+def produce_component_params(
+    client: SonarQubeClient, selector: Any, initial_params: dict[str, Any] = {}
```
PeyGis, line 51: where /when will the `initial_params` get initialized or have a value?

integrations/sonarqube/main.py
```diff
@@ -18,14 +25,73 @@ def init_sonar_client() -> SonarQubeClient:
...
+        elif event.resource_config:
+            # This might be called from places where event.resource_config is not set
+            # like on_start() when creating webhooks
+
+            selector = cast(CustomSelector, event.resource_config.selector)
+            component_query_params.update(selector.generate_request_params())
+
+    # remove project query params
+    for key in ["analyzedBefore", "onProvisionedOnly", "projects", "qualifiers"]:
+        component_query_params.pop(key, None)
```
PeyGis, line 70: the .generate_request_params() excludes props with none values. why are we doing this again?

integrations/sonarqube/main.py
```diff
@@ -18,14 +25,73 @@ def init_sonar_client() -> SonarQubeClient:
...
+
     fetched_projects = False
-    async for project_list in sonar_client.get_all_projects():
+    selector = cast(SonarQubeProjectResourceConfig, event.resource_config).selector
+    sonar_client.metrics = selector.metrics
+
+    project_query_params = produce_project_params(selector.api_filters)
+
+    component_params = produce_component_params(sonar_client, selector)
+
```
PeyGis, line 88: the get_all_project() should get one request param. You can pass either of these instead of passing both

integrations/sonarqube/main.py
```diff
@@ -35,11 +101,33 @@ async def on_project_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+    query_params = selector.generate_request_params()
+    project_query_params = produce_project_params(selector.project_api_filters)
+    initial_component_query_params = (
+        selector.project_api_filters.generate_request_params()
+        if selector.project_api_filters
+        else {}
+    )
+    component_params = produce_component_params(
+        sonar_client, selector, initial_component_query_params
+    )
```
PeyGis, line 123: it will be great if you have a centralized way of generating the project query params based on whether we are using internal or GA API. At the moment, it looks like we are duplicating this logic

integrations/sonarqube/main.py
```diff
@@ -35,11 +101,33 @@ async def on_project_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
     fetched_issues = False
-    async for issues_list in sonar_client.get_all_issues():
+    async for issues_list in sonar_client.get_all_issues(
+        query_params=query_params,
+        project_query_params=project_query_params,
+        component_query_params=component_params,
```
PeyGis, line 128: we should pass only 1 query params object instead of 2

integrations/sonarqube/client.py
```diff
@@ -113,46 +111,38 @@ async def send_api_request(
             )
             raise
 
-    async def send_paginated_api_request(
+    async def _handle_paginated_request(
```
PeyGis, line 114: let's rename to _send_paginated_request() so it's consistent with what we have in other integration

integrations/sonarqube/client.py
```diff
@@ -284,82 +263,115 @@ async def get_single_project(self, project: dict[str, Any]) -> dict[str, Any]:
...
+    async def _get_projects(
+        self, params: dict[str, Any]
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        async for projects in self._handle_paginated_request(
+            endpoint=Endpoints.PROJECTS,
+            data_key="components",
+            method="GET",
+            query_params=params,
+        ):
+            yield projects
```
PeyGis, line 275: why do we need this intermediate method for fetching projects?

integrations/sonarqube/client.py
```diff
@@ -284,82 +263,115 @@ async def get_single_project(self, project: dict[str, Any]) -> dict[str, Any]:
...
+
+    @cache_iterator_result()
+    async def get_all_projects(
+        self,
+        params: dict[str, Any] = {},
+        use_internal_api: bool = True,
+        component_params: dict[str, Any] | None = None,
```
PeyGis, line 282: we should stick with one api query param instead of receiving both params and component params

integrations/sonarqube/client.py
```diff
@@ -284,82 +263,115 @@ async def get_single_project(self, project: dict[str, Any]) -> dict[str, Any]:
...
+            async for components in self._get_components(component_params):
+                all_projects.update(
+                    {
+                        component["key"]: {
+                            **all_projects.get(component["key"], {}),
+                            **component,
+                        }
+                        for component in components
+                    }
+                )
```
PeyGis, line 314: this is awesome

PeyGis (changes requested):
thanks for all the efforts. left some comments

integrations/sonarqube/main.py
lordsarcastic: I realized that I'd have to explicitly pop out the other items I didn't need. Since I have already included the hard coded strings "projects", "qualifiers" and so on  in the next function below, explicitly stating the other properties here to be removed would result in me specifying all the properties in the `integration` file here. So I decided to only have these properties hard coded here .

integrations/sonarqube/main.py
lordsarcastic: The `pop` method will error out if the item isn't there. Also, we don't want any of these properties sent when we are calling the internal components API method.

integrations/sonarqube/main.py
lordsarcastic: The `produce_component_params` method is used for generating query params when using the internal API.  The other function is when using the GA API

integrations/sonarqube/client.py
lordsarcastic: To ensure separation of concerns, else the `get_all_projects` method will be doing too much

integrations/sonarqube/main.py
```diff
@@ -18,64 +25,97 @@ def init_sonar_client() -> SonarQubeClient:
...
+@ocean.on_resync(ObjectKind.GA_PROJECTS)
+async def on_ga_project_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    logger.info(f"Listing Sonarqube resource: {kind}")
+
+    selector = cast(SonarQubeGAProjectResourceConfig, event.resource_config).selector
+    sonar_client.metrics = selector.metrics
+
+    async for projects in sonar_client.get_components(
+        query_params=selector.generate_request_params()
+    ):
```
PeyGis, line 74: why is the ga_project kind using the get_components() which calls the internal API instead of the get_projects() which calls the GA API?

integrations/sonarqube/.port/resources/blueprints.json
```diff
@@ -124,6 +124,141 @@
     },
     "relations": {}
   },
+  {
+    "identifier": "sonarQubeGAProject",
+    "title": "SonarQube GA Project",
```
PeyGis, line 129: I see what you're doing here. What I'll prefer is that we have only 1 blueprint for project in the default resource. So let's delete the old SonarQubeProject blueprint, rename this GA project blueprint to SonarQubeProject. This way, when the integration is installed, users see only I SonarQube project blueprint.

integrations/sonarqube/.port/resources/port-app-config.yaml
```diff
@@ -36,6 +36,40 @@ resources:
...
+      metrics:
+        - code_smells
+        - coverage
+        - bugs
+        - vulnerabilities
+        - duplicated_files
+        - security_hotspots
+        - new_violations
+        - new_coverage
+        - new_duplicated_lines_density
```
PeyGis, line 51: this is good. to avoid having 2 mappings, let's delete the project kind mapping so that for fresh installations, they will use the ga_projects

PeyGis (changes requested):
we're almost there. left some comments

integrations/sonarqube/.port/resources/port-app-config.yaml
```diff
@@ -20,22 +17,24 @@ resources:
     port:
       entity:
         mappings:
-          blueprint: '"sonarQubeProject"'
+          blueprint: '"sonarQubeGAProject"'
```
PeyGis, line 20:
```suggestion
          blueprint: '"sonarQubeProject"'
```

PeyGis (changes requested):

integrations/sonarqube/integration.py
```diff
@@ -185,32 +215,59 @@ def default_metrics() -> list[str]:
...
             alias="projectApiFilters",
             description="Allows users to control which projects to query the issues for",
         )
 
-    kind: Literal["issues"]
+    kind: Literal["issues"]  # type: ignore
     selector: SonarQubeIssueSelector
 
 
+AppConfig = Annotated[
```
Tankilevitch, line 258:
```suggestion
SonarResourcesConfig = Annotated[
```

integrations/sonarqube/main.py
```diff
@@ -18,64 +25,95 @@ def init_sonar_client() -> SonarQubeClient:
...
+            component_query_params.update(selector.generate_request_params())
+    return component_query_params
+
+
 sonar_client = init_sonar_client()
 
 
 @ocean.on_resync(ObjectKind.PROJECTS)
 async def on_project_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
     logger.info(f"Listing Sonarqube resource: {kind}")
```
Tankilevitch, line 53: not indicative log, lets remove it

integrations/sonarqube/main.py
```diff
@@ -18,64 +25,95 @@ def init_sonar_client() -> SonarQubeClient:
...
-    async for project_list in sonar_client.get_all_projects():
-        yield project_list
-        fetched_projects = True
-
-    if not fetched_projects:
-        logger.error("No projects found in Sonarqube")
-        raise RuntimeError(
-            "No projects found in Sonarqube, failing the resync to avoid data loss"
-        )
+
```
Tankilevitch, line 54: lets log a warning that we want customers to use the newer resource

integrations/sonarqube/main.py
```diff
@@ -18,64 +25,95 @@ def init_sonar_client() -> SonarQubeClient:
...
+    component_params = produce_component_params(sonar_client, selector)
+
+    async for projects in sonar_client.get_components(query_params=component_params):
+        logger.info(f"Received project batch of size: {len(projects)}")
+        yield projects
+
+
+@ocean.on_resync(ObjectKind.GA_PROJECTS)
+async def on_ga_project_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    logger.info(f"Listing Sonarqube resource: {kind}")
```
Tankilevitch, line 67: remove

integrations/sonarqube/main.py
```diff
@@ -18,64 +25,95 @@ def init_sonar_client() -> SonarQubeClient:
...
-        )
 
 
 @ocean.on_resync(ObjectKind.ANALYSIS)
 @ocean.on_resync(ObjectKind.SASS_ANALYSIS)
 async def on_saas_analysis_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
     if not ocean.integration_config["sonar_is_on_premise"]:
-        fetched_analyses = False
+        logger.info("Sonar is not on-premise, processing SonarCloud on saas analysis")
         async for analyses_list in sonar_client.get_all_sonarcloud_analyses():
```
Tankilevitch, line 99: why we removed it?

integrations/sonarqube/main.py
```diff
@@ -89,9 +127,9 @@ async def handle_sonarqube_webhook(webhook_data: dict[str, Any]) -> None:
         webhook_data.get("project", {})
     )  ## making sure we're getting the right project details
     project_data = await sonar_client.get_single_project(project)
-    issues_data = await sonar_client.get_issues_by_component(project)
     await ocean.register_raw(ObjectKind.PROJECTS, [project_data])
```
Tankilevitch, line 130: what about ga_project?

integrations/sonarqube/.port/resources/port-app-config.yaml
```diff
@@ -27,15 +24,17 @@ resources:
...
-            lastAnalysisDate: .__branch.analysisDate
+            lastAnalysisDate: .analysisDate
             numberOfBugs: .__measures[]? | select(.metric == "bugs") | .value
             numberOfCodeSmells: .__measures[]? | select(.metric == "code_smells") | .value
             numberOfVulnerabilities: .__measures[]? | select(.metric == "vulnerabilities") | .value
             numberOfHotSpots: .__measures[]? | select(.metric == "security_hotspots") | .value
             numberOfDuplications: .__measures[]? | select(.metric == "duplicated_files") | .value
             coverage: .__measures[]? | select(.metric == "coverage") | .value
             mainBranch: .__branch.name
-            tags: .tags
```
Tankilevitch, line 38: why we remove tags?

integrations/sonarqube/.port/resources/port-app-config.yaml
```diff
@@ -1,12 +1,9 @@
 createMissingRelatedEntities: true
 deleteDependentEntities: true
 resources:
-  - kind: projects
+  - kind: ga_projects
     selector:
       query: 'true'
-      apiFilters:
-        filter:
-          qualifier: TRK
```
Tankilevitch, line 9: why we remove TRK?

integrations/sonarqube/tests/conftest.py
Tankilevitch, line 1: lets have an exmaple files for the "deprecated" project kind, so we will be able to come back to it, you can refer to datadog for it

integrations/sonarqube/tests/fixtures.py
```diff
@@ -0,0 +1,244 @@
...
+PURE_PROJECTS: list[dict[str, Any]] = [
+    {
+        "key": "project-key-1",
+        "name": "Project Name 1",
+        "qualifier": "TRK",
+        "visibility": "public",
+        "lastAnalysisDate": "2017-03-01T11:39:03+0300",
+        "revision": "cfb82f55c6ef32e61828c4cb3db2da12795fd767",
+        "managed": False,
+    },
```
Tankilevitch, line 12: ahh maybe this is the "raw" data of the "deprecated projects? if so ignore my previous comment

integrations/sonarqube/client.py
```diff
@@ -284,82 +267,77 @@ async def get_single_project(self, project: dict[str, Any]) -> dict[str, Any]:
...
+            data_key="components",
+            method="GET",
+            query_params=params,
+        ):
+            if enrich_project:
+                yield await asyncio.gather(
+                    *[self.get_single_project(project) for project in projects]
+                )
+            else:
+                yield projects
```
Tankilevitch, line 289: please add comment, what we get with enriching

integrations/sonarqube/client.py
```diff
@@ -284,82 +267,77 @@ async def get_single_project(self, project: dict[str, Any]) -> dict[str, Any]:
...
+        if self.organization_id:
+            params["organization"] = self.organization_id
+
+        async for projects in self._send_paginated_request(
+            endpoint=Endpoints.PROJECTS,
+            data_key="components",
+            method="GET",
+            query_params=params,
+        ):
+            if enrich_project:
```
Tankilevitch, line 283: also I prefer that the default will be `false` and that calls to that method that need enriching will specify it explicitly with why

integrations/sonarqube/client.py
```diff
@@ -165,56 +156,48 @@ async def send_paginated_api_request(
...
-                f"Fetched {len(response)} components {[item.get("key") for item in response]} from SonarQube"
-            )
-            return response
+            ):
+                logger.info(
+                    f"Fetched {len(components)} components {[item.get('key') for item in components]} from SonarQube"
+                )
+                yield await asyncio.gather(
+                    *[self.get_single_project(project) for project in components]
+                )
```
Tankilevitch, line 200: should we be afraid from rate limiting here?

Tankilevitch (changes requested):

integrations/sonarqube/.port/resources/port-app-config.yaml
lordsarcastic: The default value is already `TRK` so there isn't any need to pass it.

integrations/sonarqube/client.py
lordsarcastic: No, SonarQube has no rate limit in place. I have included this in the client docs.

integrations/sonarqube/examples/mappings.yaml
```diff
@@ -1,9 +1,157 @@
...
+            qualityGateStatus: .__branch.status.qualityGateStatus
+            lastAnalysisDate: .__branch.analysisDate
+            numberOfBugs: .__measures[]? | select(.metric == "bugs") | .value
+            numberOfCodeSmells: .__measures[]? | select(.metric == "code_smells") | .value
+            numberOfVulnerabilities: .__measures[]? | select(.metric == "vulnerabilities") | .value
+            numberOfHotSpots: .__measures[]? | select(.metric == "security_hotspots") | .value
+            numberOfDuplications: .__measures[]? | select(.metric == "duplicated_files") | .value
+            coverage: .__measures[]? | select(.metric == "coverage") | .value
+            mainBranch: .__branch.name
+            tags: .tags
```
Tankilevitch, line 38: project_ga should be first

Tankilevitch (changes requested):

integrations/sonarqube/main.py
```diff
@@ -18,64 +25,94 @@ def init_sonar_client() -> SonarQubeClient:
...
-    fetched_projects = False
-    async for project_list in sonar_client.get_all_projects():
-        yield project_list
-        fetched_projects = True
-
-    if not fetched_projects:
-        logger.error("No projects found in Sonarqube")
-        raise RuntimeError(
-            "No projects found in Sonarqube, failing the resync to avoid data loss"
-        )
```
Tankilevitch, line 36: does it mean that if a customer is still using this kind data can be deleted if no data is returned from the API?

lordsarcastic: Hmm yes, that's true. Do you want me to restore the behaviour where we prevent data from being deleted.

integrations/sonarqube/main.py
```diff
@@ -18,15 +25,43 @@ def init_sonar_client() -> SonarQubeClient:
...
+            component_query_params.update(initial_params)
+        elif event.resource_config:
+            # This might be called from places where event.resource_config is not set
+            # like on_start() when creating webhooks
+
+            selector = cast(CustomSelector, event.resource_config.selector)
+            component_query_params.update(selector.generate_request_params())
+    return component_query_params
+
+
```
Tankilevitch, line 47: should be in utils?

integrations/sonarqube/main.py
```diff
@@ -18,15 +25,43 @@ def init_sonar_client() -> SonarQubeClient:
     )
 
 
+def produce_component_params(
+    client: SonarQubeClient, selector: Any, initial_params: dict[str, Any] = {}
```
Tankilevitch, line 29: why selector is `Any`?

Tankilevitch: we know the types that can be here

integrations/sonarqube/tests/test_sync.py
```diff
@@ -0,0 +1,57 @@
+# from typing import Any
```
Tankilevitch, line 1: what is this for?

Tankilevitch (changes requested):

integrations/sonarqube/integration.py
```diff
@@ -79,6 +79,31 @@ def generate_request_params(self) -> dict[str, Any]:
         return value
 
 
+class SonarQubeGAProjectAPIFilter(BaseSonarQubeApiFilter):
+    analyzed_before: str | None = Field(
+        alias="analyzedBefore",
+        description="To retrieve projects analyzed before the given date",
+    )
```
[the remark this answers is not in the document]
lordsarcastic: It is used for the GA Projects, check the `api_filters` props

integrations/sonarqube/integration.py
```diff
@@ -79,6 +79,31 @@ def generate_request_params(self) -> dict[str, Any]:
...
+    on_provisioned_only: bool | None = Field(
+        alias="onProvisionedOnly",
+        description="To retrieve projects on provisioned only",
+    )
```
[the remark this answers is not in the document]
lordsarcastic: Same as above. Also in use

integrations/sonarqube/tests/test_sync.py
lordsarcastic: This is the sync code that kept failing here and on Jira that I couldn't resolve. We agreed to take it out for now.

Tankilevitch (approved):
LGTM

PR #1169, opened 2024-11-21
[Integration][AWS] | Fix an issue where the integration enters an endless loop on region permission error 

mk-armah:
# Description

**What:**

Fixed an issue where the integration would enter an endless loop upon encountering a permission error when querying resources in a particular region. The code now raises the permission error instead of skipping it, allowing downstream code to handle it appropriately.

**Why:**

Previously, when a permission error occurred, the integration skipped the error without proper handling, causing it to get stuck in an infinite loop. By raising the error, we ensure that the integration does not loop endlessly and that the error can be managed correctly by downstream processes.

**How:**

- Updated the error handling logic to raise the permission error instead of skipping it.
- Verified that downstream code is capable of handling the raised exception without adverse effects.
- Tested the integration to confirm that it no longer enters an endless loop on permission errors.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/integrations/aws/utils/resources.py
+++ b/integrations/aws/utils/resources.py
@@ -255,11 +255,5 @@ async def resync_cloudcontrol(
                 if not next_token:
                     break
             except Exception as e:
-                if is_access_denied_exception(e):
-                    logger.warning(
-                        f"Skipping resyncing {kind} in region {region} in account {account_id} due to missing access permissions"
-                    )
-                else:
-                    logger.error(f"Error resyncing {kind} in region {region}, {e}")
-
+                logger.error(f"Error resyncing {kind} in region {region}, {e}")
                 raise e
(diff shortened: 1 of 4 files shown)


PeyGis (approved):
lgtm

integrations/aws/utils/resources.py
```diff
@@ -261,4 +261,5 @@ async def resync_cloudcontrol(
                     )
                 else:
                     logger.error(f"Error resyncing {kind} in region {region}, {e}")
-                    raise e
+
+                raise e
```
shalev007, line 265: Let's also remove the warning log since it's kind of confusing

```suggestion
            except Exception as e:
                logger.error(
                    f"Error resyncing {kind} in region {region} in account {account_id}, {e}"
                )
                raise e

```

shalev007 (approved):
Let's push it after you remove the log

PR #1190, opened 2024-11-28
[Integration][AWS] | Fix InvalidToken Exceptions Due to Improper Token Refresh Calls

mk-armah:
# Description

**What:**

We've encountered timeouts and InvalidToken exceptions despite implementing a time-based token refresh strategy designed to refresh the token when it's 80% close to expiry, this occurs arbitrarily.

**Why:**

The token refresh may not be occurring at the appropriate points in the application.
It appears that the refresh is triggered while an active session is still utilizing the old token.
When the token is replaced during an ongoing session, the active session becomes obsolete, leading to InvalidToken exceptions.

The session refresh call is placed incorrectly within the application flow.
Active sessions are not updated with the new token after a refresh, causing them to fail when attempting to access AWS resources.

**How (Proposed Solution):**

Review and adjust the placement of the token refresh logic to ensure it does not interfere with active sessions.
Implement a check to verify if a session's token is valid or has a significant amount of time remaining before each resource (kind) retrieval operation.
Ensure that new tokens are only used for new sessions, or gracefully handle the transition for active sessions.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/integrations/aws/CHANGELOG.md
+++ b/integrations/aws/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
-- Fixed invalid token errors by implementing and auto-retry strategy using botocore's AioRefreshable Credentials
+- Fixed invalid token errors by implementing an auto-retry strategy using botocore's AioRefreshable Credentials
 
 
 ## 0.2.94 (2025-02-04)


shalev007:
Can you explain how will this solve the expired tokens issue?
I only saw you added cache layer to the tokens

integrations/aws/aws/session_manager.py
```diff
@@ -121,6 +123,10 @@ async def _get_organization_session(self) -> aioboto3.Session | None:
     def _get_account_read_role_name(self) -> str:
         return ocean.integration_config.get("account_read_role_name", "")
 
+    @staticmethod
+    def _assume_role_duration_seconds() -> int:
+        return int(ocean.integration_config.get("assume_role_duration", 900))
```
shalev007, line 128: shouldn't this default to ASSUME_ROLE_DURATION_SECONDS?

integrations/aws/aws/aws_credentials.py
```diff
@@ -9,18 +17,97 @@ def __init__(
...
+                    )
+                )
+                botocore_session = get_session()
+                setattr(botocore_session, "_credentials", refreshable_credentials)
+                if region:
+                    botocore_session.set_config_variable("region", region)
+                autorefresh_session = aioboto3.Session(
+                    botocore_session=botocore_session
+                )
+                return autorefresh_session
```
shalev007, line 97: this is a pretty complex part of the code what do you think about adding maybe some comments about why this code exists and what is happening baisically?

integrations/aws/aws/aws_credentials.py
```diff
@@ -33,24 +120,22 @@ async def update_enabled_regions(self) -> None:
...
-            )
-        else:
-            return aioboto3.Session(
-                aws_access_key_id=self.access_key_id,
-                aws_secret_access_key=self.secret_access_key,
-                region_name=region,
-            )
-
-    async def create_session_for_each_region(
+    async def create_refreshable_session_for_each_region(
```
shalev007, line 123: I don't see why change the name pf this function, from an API standpoint the person using this function it will not matter if it's refreshable or not, WDYT?

shalev007:
Got this error when trying to run on single account mode
image

integrations/aws/aws/aws_credentials.py
mk-armah: I agree

shalev007 (approved):

integrations/aws/aws/session_manager.py
```diff
@@ -152,36 +152,25 @@ async def _update_available_access_credentials(self) -> None:
...
-            await aws_credentials.update_enabled_regions()
-            self._aws_credentials.append(aws_credentials)
+            await credentials.update_enabled_regions()
+            self._aws_credentials.append(credentials)
             self._aws_accessible_accounts.append(account)
+
         except sts_client.exceptions.ClientError as e:
             if is_access_denied_exception(e):
                 logger.info(f"Cannot assume role in account {account['Id']}. Skipping.")
+                pass  # Skip the account if assume_role fails due to permission issues or non-existent role
```
shalev007: What is the difference between with pass and without?

PR #1226, opened 2024-12-12
[Core] Wrap smoke clean in bash

erikzaadi:
Ensure that the cleaning of the smoke test works when skipping the Port API


Tankilevitch (approved):
LGTM!

PR #1246, opened 2024-12-17
[Integration][AWS] - Remodel AWS Auth Strategy with fix for InvalidToken Exceptions Due to Improper Token Refresh Calls.

mk-armah:
# Description

**What:**

We've encountered timeouts and InvalidToken exceptions despite implementing a time-based token refresh strategy designed to refresh the token when it's 80% close to expiry, this occurs arbitrarily.

**Why:**

The token refresh may not be occurring at the appropriate points in the application.
It appears that the refresh is triggered while an active session is still utilizing the old token.
When the token is replaced during an ongoing session, the active session becomes obsolete, leading to InvalidToken exceptions.

The session refresh call is placed incorrectly within the application flow.
Active sessions are not updated with the new token after a refresh, causing them to fail when attempting to access AWS resources.

**How (Proposed Solution):**

Review and adjust the placement of the token refresh logic to ensure it does not interfere with active sessions.
Implement a check to verify if a session's token is valid or has a significant amount of time remaining before each resource (kind) retrieval operation.
Ensure that new tokens are only used for new sessions, or gracefully handle the transition for active sessions.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


phalbert (commented):
Really great work on addressing the InvalidToken exceptions issue. The refactored authentication strategy looks solid.

Given the core problem was tokens being invalidated during cloud control operations, can we add tests for these? Is it possible? There are two scenarios:

- Verify that the 80% expiry threshold works correctly.
- Verify that existing sessions remain valid when tokens refresh.

integrations/aws/utils/aws.py
```diff
@@ -4,35 +4,22 @@
...
 
-_session_manager: SessionManager = SessionManager()
-
-CACHE_DURATION_SECONDS = (
-    0.80 * ASSUME_ROLE_DURATION_SECONDS
-)  # Refresh role credentials after exhausting 80% of the session duration
 
-lock = Lock()
+provider = OrganizationCredentialsProvider()
+_session_manager: SessionManager = SessionManager(provider)
```
shalev007, line 17: So we always use the OrganizationCredentialsProvider even on a single account mode?

integrations/aws/aws/auth_provider.py
```diff
@@ -0,0 +1,187 @@
...
+            return [{"Id": caller_identity["Account"], "Name": "No name found"}]
+
+    async def get_account_credentials(
+        self, sts_client: STSClient, account: Dict[str, Any]
+    ) -> Optional[AwsCredentials]:
+        # For application-only scenario, there's no cross-account role assumption.
+        # Just return the application credentials again if needed.
+        raise NotImplementedError(
+            "No cross-account assumption for application-only scenario."
+        )
```
shalev007, line 87: I don't think this method should be abstracted if only OrganizationCredentialsProvider can use it

integrations/aws/aws/auth_provider.py
```diff
@@ -0,0 +1,187 @@
...
+from port_ocean.context.ocean import ocean
+from loguru import logger
+from types_aiobotocore_sts import STSClient
+from utils.misc import is_access_denied_exception
+from datetime import datetime, timezone, timedelta
+
+ASSUME_ROLE_DURATION_SECONDS = 3600
+
+
+class CredentialsProvider(ABC):
```
shalev007, line 14: can you also explain what is the responsibility of this class?

integrations/aws/aws/auth_provider.py
shalev007, line 1: I like the fact the you separated some of the complex logic in to classes, But it seems as though ApplicationCredentialsProvider is not really useful on it's own which makes me wonder why it deserve a class

integrations/aws/aws/aws_credentials.py
```diff
@@ -33,14 +49,62 @@ async def update_enabled_regions(self) -> None:
...
             )
+
+            botocore_session = get_session()
+            setattr(botocore_session, "_credentials", credentials)
+            if region:
+                botocore_session.set_config_variable("region", region)
+
+            autorefresh_session = aioboto3.Session(botocore_session=botocore_session)
+            return autorefresh_session
+
```
shalev007, line 107: What did we change since the prev PR?

PR #1265, opened 2024-12-24
[Core] Added logs for JQ misconfigured mappings

No0b1t0:
created a test to simulate wrong mapping and validated for single and multiple data rows

# Description

What -following a bug ticket https://getport.atlassian.net/browse/PORT-12112, I have created a test to simulate wrong mapping and validated for single and multiple data rows

Why - when the customer misconfigures a mapping we do not display the misconfigured fields and thus the customer can't understand the reason for the non working pipe

How - I've added a misconfiguration dict to map all of the misconfigured mappings

the logs shown prior to this change 
```
2024-12-24 13:00:54.025 | DEBUG    | port_ocean.utils.queue_utils:_start_processor_worker:21 - Processing async task
2024-12-24 13:00:54.036 | DEBUG    | port_ocean.core.handlers.entity_processor.jq_entity_processor:_parse_items:232 - Finished parsing raw results into entities with 0 errors. errors: []
```
after the change: 
```
2024-12-24 13:00:54.024 | INFO     | port_ocean.core.handlers.entity_processor.jq_entity_processor:_parse_items:221 - Parsing 2 raw results into entities
2024-12-24 13:00:54.025 | DEBUG    | port_ocean.utils.queue_utils:_start_processor_worker:21 - Processing async task
2024-12-24 13:00:54.036 | DEBUG    | port_ocean.core.handlers.entity_processor.jq_entity_processor:_parse_items:232 - Finished parsing raw results into entities with 0 errors. errors: []
2024-12-24 13:00:54.036 | INFO     | The mapping resulted with invalid values for identifier, blueprint, properties. Mapping result: {'identifier': '.ark', 'foo': '.bazbar', 'desc': '.foobar', 'name': '.bar.baz'}

```
note the new line for the misconfigured mappings

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


port_ocean/core/handlers/entity_processor/jq_entity_processor.py
```diff
@@ -92,21 +93,26 @@ async def _search_as_bool(self, data: dict[str, Any], pattern: str) -> bool:
         )
 
     async def _search_as_object(
-        self, data: dict[str, Any], obj: dict[str, Any]
+        self,
+        data: dict[str, Any],
+        obj: dict[str, Any],
+        misconfigurations: dict[str, str] = {},
```
matan84, line 99: https://www.pullrequest.com/blog/python-pitfalls-the-perils-of-using-lists-and-dicts-as-default-arguments/

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
```diff
@@ -230,6 +251,10 @@ async def _parse_items(
                         examples_to_send.append(result.raw_data)
                 else:
                     failed_entities.append(parsed_entity)
+        if len(entity_misconfigurations) > 0:
+            logger.info(
+                f"Entity mapping misconfigurations identified. misconfigured mapping: {entity_misconfigurations}"
```
matan84, line 256: Explain what's going on here, Don't expect anyone seeing this will remember what they put in the mapping

port_ocean/tests/core/handlers/entity_processor/test_jq_entity_processor.py
```diff
@@ -269,3 +269,35 @@ async def test_parse_items_performance_10000(
             assert len(result.entity_selector_diff.passed) == 1
             assert result.entity_selector_diff.passed[0].properties.get("foo") == "bar"
             assert not result.errors
+
+    async def test_parse_items_wrong_mapping(
+        self, mocked_processor: JQEntityProcessor
+    ) -> None:
+        mapping = Mock()
+        mapping.port.entity.mappings.dict.return_value = {
```
matan84, line 277: Let's create a helper method which creates mock entities and use it instead

port_ocean/tests/core/handlers/entity_processor/test_jq_entity_processor.py
```diff
@@ -269,3 +269,35 @@ async def test_parse_items_performance_10000(
             assert len(result.entity_selector_diff.passed) == 1
             assert result.entity_selector_diff.passed[0].properties.get("foo") == "bar"
             assert not result.errors
+
+    async def test_parse_items_wrong_mapping(
```
matan84, line 273: Let's also add tests for different kinds of mappings which are supposed to be valid (i.e. items to parse, file kinds etc..)

port_ocean/core/ocean_types.py
```diff
@@ -30,6 +37,7 @@ class EntitySelectorDiff(NamedTuple):
 class CalculationResult(NamedTuple):
     entity_selector_diff: EntitySelectorDiff
     errors: list[Exception]
+    misconfigured: dict[str, str] = {}
```
matan84, line 40: also https://www.pullrequest.com/blog/python-pitfalls-the-perils-of-using-lists-and-dicts-as-default-arguments/

matan84: perhaps you can use defaultdict here

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -184,7 +184,7 @@ async def _register_in_batches(
         send_raw_data_examples_amount = (
             SEND_RAW_DATA_EXAMPLES_AMOUNT if ocean.config.send_raw_data_examples else 0
         )
-        all_entities, register_errors = await self._register_resource_raw(
+        all_entities, register_errors,_ = await self._register_resource_raw(
```
matan84, line 187: isn't this a lint error?

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
```diff
@@ -230,6 +251,10 @@ async def _parse_items(
                         examples_to_send.append(result.raw_data)
                 else:
                     failed_entities.append(parsed_entity)
+        if len(entity_misconfigurations) > 0:
+            logger.info(
```
matan84, line 255: I think that showing it here is too verbose, let's move it to be per batch of entities instead of every entity

matan84 (changes requested):
Let's also update the pyproject.toml, and a changelog to the changelog file

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
Tankilevitch, line 255: wouldn't it be too spammy? for each entity?

Tankilevitch: I think it should be in `debug` mode and we should log it in `info` as an aggregated mode for a batch of entities

Tankilevitch (changes requested):
Nice job, overall I think that this log can be very spammy

Maybe doing something that is more aggregated can be useful, currently we have logs of how many raw entities we got and then how many were generated, so worth adding to it why some weren't created

port_ocean/core/integrations/mixins/sync_raw.py
No0b1t0: lint passes :) 
Thought so too at first ^_^

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
No0b1t0: we've discussed this 🙏

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
No0b1t0: we've discussed this 🙏

matan84 (approved):
LGTM

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
```diff
@@ -95,21 +96,26 @@ async def _search_as_bool(self, data: dict[str, Any], pattern: str) -> bool:
         )
 
     async def _search_as_object(
-        self, data: dict[str, Any], obj: dict[str, Any]
+        self,
+        data: dict[str, Any],
+        obj: dict[str, Any],
+        misconfigurations: dict[str, str] | None = None,
     ) -> dict[str, Any | None]:
         search_tasks: dict[
```
Tankilevitch, line 104: so we are not returning it? only performing a manipulation of the object if passed?

Tankilevitch: maybe worth initializing one if not passed from outside?

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
```diff
@@ -118,12 +124,20 @@ async def _search_as_object(
         for key, task in search_tasks.items():
             try:
                 if isinstance(task, list):
-                    result[key] = [await task for task in task]
+                    result_list = []
+                    for task in task:
+                        task_result = await task
+                        if task_result is None:
+                            misconfigurations[key] = obj[key]
```
Tankilevitch, line 131: wouldn't it fail? if `misconfigurations` object wasn't passed or passed as `None`? I couldn't find where you initialize it inside of the method while you are supporting `None`

No0b1t0: great catch! thank you 🙏

No0b1t0: its being initialized at line 152 and passed into the method, but in any case, this should be covered and if None it should be initialised to prevent an error.

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
```diff
@@ -95,21 +96,29 @@ async def _search_as_bool(self, data: dict[str, Any], pattern: str) -> bool:
...
 
     async def _search_as_object(
-        self, data: dict[str, Any], obj: dict[str, Any]
+        self,
+        data: dict[str, Any],
+        obj: dict[str, Any],
+        misconfigurations: dict[str, str] | None = None,
     ) -> dict[str, Any | None]:
+        if misconfigurations is None:
+            misconfigurations = {}
```
Tankilevitch, line 105: I wonder what is the value initializing it inside the method if we are never returning it?

probably worth returning it then? or not initializing it in the first place? could be reasonable use case where someone who uses the method doesn't need the misconfigurations? 

Let me know what you think about it

Tankilevitch (changes requested):

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- towncrier release notes start -->
+## 0.16.1 (2024-12-25)
+
+### Bug Fixes
+
+- Added new info log for JQ mapping per batch to notify of misconfigured JQ mappings between a property and the JQ target.
```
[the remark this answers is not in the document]
Tankilevitch: also can you add the new capabilities for the "required" properties?

port_ocean/tests/core/handlers/entity_processor/test_jq_entity_processor.py
Tankilevitch, line 273: can you add a test for logging the missing required fields?

port_ocean/core/ocean_types.py
```diff
@@ -30,6 +38,7 @@ class EntitySelectorDiff(NamedTuple):
 class CalculationResult(NamedTuple):
     entity_selector_diff: EntitySelectorDiff
     errors: list[Exception]
+    misconfigured: dict[str, str] = field(default_factory=dict)
```
Tankilevitch, line 41: I wonder if `misconfigured` is specific enough? as I am not sure I totally understand what it relates to as misconfigured

maybe `misonfigured_entity_keys` ?

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
No0b1t0: currently there is only 1 usage for this method and the method itself is being called recursively, due to this, the misconfiguration is being passed into itself and it is initialized for the case that it was not passed (just to be safe in the future) 
if in the future someone will call the method without passing the misconfiguration property, if he has misconfigured mappings, he will get it as part of the class definition itself (MappedEntity).

I think that its an over kill but its ok to leave it like this, just to be on the safe side. 
WDYT?

port_ocean/tests/core/handlers/entity_processor/test_jq_entity_processor.py
No0b1t0: I wanted to avoid monitoring the logs themselves so instead i've asserted for the assumed misconfigurations. 
in this case I assume that the log will be written properly. 
assert result.misconfigured == {
            "identifier": ".ark",
            "foo": ".bazbar",
            "desc": ".foobar",
            "name": ".bar.baz",
        }
        
@Tankilevitch, do you want to make sure that in the future the text message won't change? this is why you want the log to be tested as well?

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
```diff
@@ -95,21 +96,38 @@ async def _search_as_bool(self, data: dict[str, Any], pattern: str) -> bool:
...
     ) -> dict[str, Any | None]:
+        """
+        the _srearch_as_object function is here to allow us to identify and extract the relevant value for the chosen key and populate it into the entity
+        data: the property itself that holds the key and the value, it is being passed to the task and we get back a task item,
+            if the data is a dict, we will recursively call this function again
+        obj: the key that we want its value to be mapped into our entity
+        misconfigurations: due to the recursive nature of this function,
+            we aim to have a dict that represents all of the misconfigured properties and when used recursively,
+            we pass this value to add the relevant misconfigured keys
+        """
```
Tankilevitch, line 112: missing response type

port_ocean/core/handlers/entity_processor/jq_entity_processor.py
```diff
@@ -118,12 +136,20 @@ async def _search_as_object(
         for key, task in search_tasks.items():
             try:
                 if isinstance(task, list):
-                    result[key] = [await task for task in task]
+                    result_list = []
+                    for task in task:
+                        task_result = await task
+                        if task_result is None:
```
Tankilevitch, line 142:
```suggestion
                        if task_result is None and misconfiguration is not None:
```

Tankilevitch: same below

port_ocean/tests/core/handlers/entity_processor/test_jq_entity_processor.py
```diff
@@ -269,3 +269,35 @@ async def test_parse_items_performance_10000(
...
+        self, mocked_processor: JQEntityProcessor
+    ) -> None:
+        mapping = Mock()
+        mapping.port.entity.mappings.dict.return_value = {
+            "identifier": ".ark",
+            "blueprint": ".baz",
+            "properties": {"foo": ".bazbar"},
+            "title": ".foo",
+            "desc": ".foobar",
+            "name": ".bar.baz",
```
Tankilevitch, line 283: lets follow mapping conventions 

there is no such keys as `name` and `desc` I would rather that you add it to properties and relations

Tankilevitch: maybe worth adding how it would be with `items_to_parse` ?

No0b1t0: used the properties under the mapping property from [our docs](https://docs.getport.io/build-your-software-catalog/customize-integrations/configure-mapping/)

Tankilevitch (approved):
LGTM, great job!!

PR #1296, opened 2025-01-03
[Integration][Jira] Upgrade Jira integration to export issues by sprint

lordsarcastic:
# Description

What - Added support to export issues by sprint

Why - To allow users to aggregate issues by sprints

How - Added support for ingesting other fields apart from the default fields which allowed to ingest the sprints custom field

Note: Certain refactors expected are intentionally ignored from this PR, this includes abstracting client class initialization and fixing webhook bug. This is to ensure this PR focuses on one thing only to avoid a long review process. In addition, tests aren't added since the addition of tests didn't bode well for the original PR. Instead, tests will be included in a different PR coming with bug fixes and improvements.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/jira/.port/resources/port-app-config.yaml
```diff
@@ -47,15 +47,15 @@ resources:
...
             updated: .fields.updated
-            resolutionDate: .fields.resolutiondate
+            resolutionDate: .fields.resolutiondate // ""
           relations:
             project: .fields.project.key
-            parentIssue: .fields.parent.key
+            parentIssue: .fields.parent.key // ""
             subtasks: .fields.subtasks | map(.key)
-            assignee: .fields.assignee.accountId
+            assignee: .fields.assignee.accountId // ""
```
Tankilevitch, line 60: why?

lordsarcastic: When running the integration, I noticed certain logs showing that some fields returned `null` or were missing. I added the `// ""` part to explicitly handle null values.

integrations/jira/jira/overrides.py
```diff
@@ -5,16 +5,21 @@
...
-from pydantic import BaseModel, Field
+from pydantic import Field
 
 
-class JiraResourceConfig(ResourceConfig):
-    class Selector(BaseModel):
-        query: str
-        jql: str | None = None
+class JiraIssueSelector(Selector):
+    query: str
```
PeyGis, line 12: the `Selector` class already has the `query` prop so we can remove this

integrations/jira/main.py
```diff
@@ -61,7 +62,17 @@ async def on_resync_issues(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
         ocean.integration_config["atlassian_user_token"],
     )
 
-    async for issues in client.get_paginated_issues():
+    params = {}
+    config = typing.cast(JiraIssueConfig, event.resource_config)
+
+    if config.selector.jql:
+        params["jql"] = config.selector.jql
+        logger.info(f"Found JQL filter: {config.selector.jql}")
```
PeyGis, line 70:
```suggestion
        logger.info(f"Found JQL filter: {config.selector.jql}...Adding to request param ")
```

PeyGis (changes requested):
left some minor comment

integrations/jira/.port/resources/port-app-config.yaml
Tankilevitch: is that an issue with us relying on a key that doesn't actually exists? or is available in specific cases?

PeyGis (approved):

integrations/jira/.port/resources/port-app-config.yaml
Tankilevitch: do we document is somewhere when this data is available?

lordsarcastic: No, we aren't really relying on it per se. It is the assignee field which may not be present if an issue hasn't been assigned to anyone yet. It may be null in such cases. It doesn't need to be documented since it is common knowledge, else we'll have to document all the fields including an absent description.

lordsarcastic: The user themselves are the ones who control that field. It isn't on us.

Tankilevitch (approved):
LGTM

PR #1340, opened 2025-01-21
Fix http_client lost api_auth header

simorex80:
# Description

What - After first run http_client api_auth header is lost

Why - Preserve working after first run

How - Moved Fix http_client api_auth header set before each run

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/integrations/sonarqube/CHANGELOG.md
+++ b/integrations/sonarqube/CHANGELOG.md
@@ -7,12 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
-## 0.1.137 (2025-01-21)
+## 0.1.139 (2025-01-23)
 
 
 ### Improvements
 
 - Fix http_client lost api_auth header
+
+
+## 0.1.138 (2025-01-22)
+
+
+### Improvements
+
+- Updated mappings to have typed array items
+
+
+## 0.1.137 (2025-01-22)
+
+
+### Improvements
+
+- Bumped ocean version to ^0.18.2
 
 
 ## 0.1.136 (2025-01-21)
--- a/integrations/sonarqube/CHANGELOG.md
+++ b/integrations/sonarqube/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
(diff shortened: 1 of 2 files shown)


lordsarcastic (commented):
Thanks for taking a step to help! 

I ran some tests on the current state of the integration and this doesn't fix the issue. We are aware that SonarQube projects API (which is internal and not GA) fails intermittently which gives rise to these issues. There is the very big possibility that after you made this change, the API started working intermittently which may have led to the belief that this fixes the issue.

In addition, the current pattern of updating headers is used across several integrations. If this was a bug, it would have been discovered in several integrations also.

Thanks once again for the fix and adhering to the change log standards 🔥

simorex80:
We are using version 9.9.1 of the Community Edition of SonarQube.
With the indicated change, we no longer experienced data loss issues, and the situation has finally stabilized.
Considering that this modification does not cause regressions (setting the header for each request is a correct approach) and taking into account the resolution of the issue, can you proceed with merging the request?

simorex80:
Also, consider that I debugged the integration locally.
The fact that the header gets lost after the first run is evident, which is why it is necessary to reset it every time.

lordsarcastic:
> Also, consider that I debugged the integration locally. The fact that the header gets lost after the first run is evident, which is why it is necessary to reset it every time.

Thank you for your hard work and investigation on this. I did some further checks which yielded results consistent with your findings. The header gets lost due to the way Ocean handles `httpx` client which is used internally and managed by a context manager. This yields a wide range of errors from `401`, `403` errors (in my case) to an empty response in your case. A cleaner and consistent fix involves updating your fix to do the following:

In `main.py`,

Take out the global initialization of the client in line 28:
```python
```

And make this call in all the functions they are used which includes the resync functions, webhook handler and the `on_start` lifecycle function.

mk-armah (changes requested):
Great 👍🏾, few changes left to go

1. I know you didn't commit that but can you also remove [this line](https://github.com/port-labs/ocean/blob/bf18ea93be8d74a1485346b6a1c1cdbebd70c190/integrations/sonarqube/client.py#L158)
2. Bump the integration version in pyproject.toml to the latest... that would be `0.1.142 `now

lordsarcastic:
Closing this due to inactivity, in favour of #1359

PR #1341, opened 2025-01-22
[Core] Fixed query to make the search easier on port's side - second improvement

yaelibarg:
# Description

What - improvements to reduce port's load

Why - to make the query easier for port

How - fixed the identifier rule to be one

## Type of change

### Core testing checklist


Tankilevitch (approved):
LGTM

PR #1353, opened 2025-01-26
[Core] Raise error if integration gets empty config from port 

yaelibarg:
# Description

What - raise error if integration gets empty config from port 

Why - to avoid integration to delete entities if resync gets 0 entities 

How - raise error

## Type of change

### Core testing checklist


Tankilevitch (approved):
looks good! 🚀

PR #1364, opened 2025-01-31
[Integration][PagerDuty] Bug | Getting incident analytics in PagerDuty causes hitting of rate limits

lordsarcastic:
# Description

What - Fix 429 errors in PagerDuty incident analytics calls

Why - Users experience rate limits when incident analytics are enabled

How - Switching API to use bulk incident analytics retrieved service by service

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/integrations/pagerduty/clients/pagerduty.py
+++ b/integrations/pagerduty/clients/pagerduty.py
@@ -228,7 +228,7 @@ class PagerDutyClient:
                     },
                     extensions={
                         "retryable": True,
-                    }
+                    },
                 )
                 yield data["data"]
                 more = data["more"]
--- a/integrations/pagerduty/clients/pagerduty.py
+++ b/integrations/pagerduty/clients/pagerduty.py
@@ -226,6 +226,9 @@ class PagerDutyClient:
                         "filters": {"service_ids": service_ids},
                         "starting_after": starting_after,
                     },
+                    extensions={
+                        "retryable": True,
+                    }
                 )
                 yield data["data"]
                 more = data["more"]
--- a/integrations/pagerduty/clients/pagerduty.py
+++ b/integrations/pagerduty/clients/pagerduty.py
@@ -5,6 +5,7 @@ import httpx
 from loguru import logger
 from port_ocean.context.event import event
 from port_ocean.utils import http_async_client
+from port_ocean.utils.cache import cache_iterator_result
 
 from .utils import get_date_range_for_last_n_months
 
@@ -13,6 +14,11 @@ USER_KEY = "users"
 MAX_CONCURRENT_REQUESTS = 10
 PAGE_SIZE = 100
 OAUTH_TOKEN_PREFIX = "pd"
+
+
+class Resources:
+    INCIDENTS = "incidents"
+    SERVICES = "services"
 
 
 class PagerDutyClient:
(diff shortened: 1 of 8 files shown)


integrations/pagerduty/clients/pagerduty.py
```diff
@@ -15,6 +16,11 @@
 OAUTH_TOKEN_PREFIX = "pd"
 
 
+class Resources:
+    INCIDENTS = "incidents"
+    SERVICES = "services"
+
```
PeyGis, line 22: We already have an ObjectKind in the integration.py file. please use it

integrations/pagerduty/clients/pagerduty.py
```diff
@@ -204,23 +210,66 @@ async def update_oncall_users(
...
-            if isinstance(e, httpx.HTTPStatusError) and e.response.status_code == 404:
-                logger.debug(
-                    f"Incident {incident_id} analytics data was not found, skipping..."
+            while more:
+                data = await self.send_api_request(
+                    endpoint="analytics/raw/incidents",
+                    method="POST",
+                    json_data={
+                        "filters": {"service_ids": service_ids},
+                        "starting_after": starting_after,
```
PeyGis, line 227: the starting after needs to be configurable. In the service analytics guide, we give users the [option to tell us the starting period. ](https://docs.port.io/build-your-software-catalog/sync-data-to-catalog/incident-management/pagerduty/#ingesting-service-analytics)

integrations/pagerduty/clients/pagerduty.py
```diff
@@ -204,23 +210,66 @@ async def update_oncall_users(
...
+            yield services
+
+    async def get_incidents(
+        self, params: dict[str, Any] | None = None
+    ) -> AsyncGenerator[list[dict[str, Any]], Any]:
+        async for incidents in self.paginate_request_to_pager_duty(
+            resource=Resources.INCIDENTS,
+            params=params,
+        ):
+            yield incidents
```
PeyGis, line 272: why do we need this?

PeyGis (changes requested):

integrations/pagerduty/clients/pagerduty.py
lordsarcastic: I considered that and then thought that we do not want any explicit dependency between the pagerduty client and the integration config from Ocean. We either pass it to the respective methods as we do in other resync functions or we specify an explicit `Resources` class as I did here.

If we want a single source of truth, I suggest we move the `ObjectKind` to the `main.py` file and let it inherit values from the `Resources` class as a mixin alongside `StrEnum` since both `ObjectKind` and `Resources` will inevitably contain the same property, value key pair.

Or we shamelessly duplicate. Won't hurt much. The whole point is to ensure the `PagerDutyClient`  does not directly depend on Ocean internals save for the unavoidable `http_client`.

What do you think?

integrations/pagerduty/clients/pagerduty.py
PeyGis: Yes my bad. I meant the created_at_start and created_at_end

integrations/pagerduty/clients/pagerduty.py
PeyGis: Got it. I think it will be better if we move the ObjectKind out to a different file maybe util.py then we can reuse it in the integration.py, main.py and client.py.

integrations/pagerduty/clients/pagerduty.py
lordsarcastic: Yeah both applies to the incidents whose analytics would be returned which means that after enabling the incidents analytics, if these properties are added and made configurable, some of the incidents will not have analytics. The only advantage to this is reduced paginated API calls, since PagerDuty already returns the analytics by services rather than individual calls to the analytics API for each incident.

mk-armah:
Continued in https://github.com/port-labs/ocean/pull/1376

PR #1371, opened 2025-02-04
[Integration][Gitlab] Fix Memory Leak Identified In Live Events Sync

mk-armah:
##Description

What

Currently, when a live event is received, we pass a deep copy of the event object to each observer. This was originally implemented to ensure that every observer received its own, uniquely referenced copy of the event. However, in our current workflow, observers do not modify the event data—they only use it to fetch resources from GitLab.

Why
- Since the observers only perform read operations (i.e., fetching data) and do not modify the event object, creating deep copies is redundant.
- It has been observed that the deep‑copied objects are not being garbage collected after use, which results in increased memory usage and could eventually lead to memory exhaustion.
- The deep copy operation introduces additional processing time and memory overhead that is not needed given the current use case.

How
- Modify the event handling logic to pass the original event object (or a shallow copy if absolutely necessary for safety) to observers instead of performing a deep copy.
- Since observers only read from the event data, they can safely operate on the same shared object without risking race conditions or data corruption.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/integrations/gitlab/CHANGELOG.md
+++ b/integrations/gitlab/CHANGELOG.md
@@ -7,12 +7,21 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+0.2.32 (2025-02-04)
+===================
+
+### Improvements
+
+- Removed the need for sending a deepcopy of the event body to all listeners, which previously led to a memory leak
+
+
 0.2.31 (2025-02-04)
 ===================
 
 ### Improvements
 
 - Bumped ocean version to ^0.18.8
+- Removed the need for sending a deepcopy of the event body to all listeners, which previously led to a memory leak
 
 
 0.2.30 (2025-01-29)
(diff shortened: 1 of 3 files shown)


Tankilevitch (approved):

integrations/gitlab/CHANGELOG.md
```diff
@@ -14,12 +14,21 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+- Removed the need for sending a deepcopy of the event body to all listeners, which previously led to a memory leak
+
+
 0.2.31 (2025-02-04)
 ===================
 
 ### Improvements
 
 - Bumped ocean version to ^0.18.8
+- Removed the need for sending a deepcopy of the event body to all listeners, which previously led to a memory leak
```
Tankilevitch, line 31: remove

mk-armah: Good catch

PR #1378, opened 2025-02-07
[Core] Added flag to destroy integration config while running clean defaults cli command

oiadebayo:
# Description  

What  
- Added support for complete removal of an integration in the `ocean defaults clean` command.
- Introduced a new flag `--destroy` to enable full deletion of the integration

Why  
- Previously, the command only cleaned up the default blueprint and associated data but did not allow full removal.
- This enhancement provides users with a way to completely delete an integration during development

How  
- Updated the command logic to handle the `--destroy` flag.
- Ensured `--destroy` removes all integration-related configurations.

## Type of change

### Core testing checklist

## Screenshots

## API Documentation


Tankilevitch (approved):
LGTM

PR #1453, opened 2025-03-03
[Integration][AWS] Resolve Cloud Control Pagination Issue

mk-armah:
# Description

**What**  
- This PR fixes an issue in the AWS integration where the API returns only 1000 resources in a single batch for the SQS Kind, despite the MaxResults parameter being set, and no NextToken is provided to paginate through all available resources.
- Some resources have uncontrollable numbers of resources retrieved per paginated request.

**Why**  
The current behavior leads to incomplete ingestion (e.g., tests made it clear that, only 1000 out of 1500 SQS queues available are processed) since the MaxResults parameter is ignored and subsequent pages cannot be retrieved. This bug, observed on the AWS SQS kind and potentially affecting other resources using the cloudcontrol API, needs to be resolved to ensure that all resources are ingested as expected.

**How**  
- Introduced a custom pagination utility `AsyncPaginator` to better manage paginated API requests.  
- Modified the integration to detach SQS from the cloudcontrol API, addressing the missing `NextToken` issue that previously prevented further pagination.  
- The new utility ensures that regardless of the response from the API, the given Batch Size is respected during pagination.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/aws/utils/misc.py
```diff
@@ -104,3 +119,53 @@ def get_matching_kinds_and_blueprints_from_config(
                 allowed_kinds[kind] = [blueprint]
 
     return allowed_kinds, disallowed_kinds
+
+
+class AsyncPaginator:
```
shalev007, line 124: Can you add description to the class and properties?

integrations/aws/utils/misc.py
```diff
@@ -104,3 +119,53 @@ def get_matching_kinds_and_blueprints_from_config(
...
+
+    _RESYNC_BATCH_SIZE = 100
+    __slots__ = ("client", "method_name", "list_param")
+
+    def __init__(self, client: "AioBaseClient", method_name: str, list_param: str):
```
shalev007, line 129: Why is the `AioBaseClient` in quotes and not a type?

integrations/aws/utils/misc.py
```diff
@@ -104,3 +119,53 @@ def get_matching_kinds_and_blueprints_from_config(
...
+        self.list_param = list_param
+
+    async def paginate(self, **kwargs: Any) -> AsyncGenerator[List[Any], None]:
+        """
+        Asynchronously iterates over API pages.
+        Each iteration yields the list of items extracted from the API page.
+        """
+        paginator = self.client.get_paginator(self.method_name)
+        async for page in paginator.paginate(**kwargs):
+            yield page.get(self.list_param, [])
```
shalev007, line 141: Why no error handling?

shalev007 (approved):
Big like left a couple of questions

integrations/aws/utils/resources.py
```diff
@@ -126,6 +127,60 @@ async def describe_single_resource_cloudcontrol(
             return serialized
 
 
+async def resync_sqs_queue(
```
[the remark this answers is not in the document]
mk-armah: The bahavior is different than what is currently expected for resources fetched from `resync_custom_kind`, This requires a two step process, querying QueueUrls via the list-queue endpoint and fetching the actual attributes/properties, and to ensure backward compatibility, it was paramount to fetch properties/attributes from cloudcontrol

integrations/aws/utils/misc.py
mk-armah: Its actually a type, works too.
it's only imported during type checking (when TYPE_CHECKING is True) to avoid runtime import errors since aioboto3 doesn't expose its types directly.

integrations/aws/utils/misc.py
mk-armah: Handled in callers

integrations/aws/utils/resources.py
```diff
@@ -126,6 +127,60 @@ async def describe_single_resource_cloudcontrol(
...
+    async with session.client("sqs") as sqs_client:
+        paginator = AsyncPaginator(
+            client=sqs_client,
+            method_name="list_queues",
+            list_param="QueueUrls",
+        )
+        try:
+            async with session.client("cloudcontrol") as cloudcontrol:
+                async for page in paginator.batch_paginate(MaxResults=1000):
+                    page_resources = []
```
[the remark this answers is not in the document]
mk-armah: Right, The idea is to leverage on the capability of retrieving more per request to reduce Network calls, thereby reducing chances of hitting rate limit. We proceed to handle resending the large number in chunks of 100, and further down to 10 if get_resource_api is being used.

integrations/aws/utils/misc.py
```diff
@@ -104,3 +117,153 @@ def get_matching_kinds_and_blueprints_from_config(
                 allowed_kinds[kind] = [blueprint]
 
     return allowed_kinds, disallowed_kinds
+
+
+def chunked_list(lst: List[Any], batch_size: int) -> Iterator[List[Any]]:
+    """Yield successive batch-sized chunks from lst."""
+    for i in range(0, len(lst), batch_size):
+        yield lst[i : i + batch_size]
```
Tankilevitch, line 125:
```suggestion

def process_list_in_chunks(_list: List[Any], batch_size: int) -> Iterator[List[Any]]:
    """Yield successive batch-sized chunks from list."""
    for i in range(0, len(_list), batch_size):
        yield lst[i : i + batch_size]
```

Tankilevitch (approved):

PR #1462, opened 2025-03-07
[Integration][ADO] Bug | ADO project retrieval fails with JSON decode error

lordsarcastic:
# Description

What - Fixed a bug where retrieving items that fail with empty responses throws a JSONDecode error.

Why - In cases where responses are skipped since they are 404 errors, the response generated is `None` and other 
kinds might attempt to serialize the response which will throw an error.

How - To fix this, we explicitly ignore responses that are likely to return a 404.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/azure-devops/azure_devops/client/base_client.py
Tankilevitch, line 1: I think send_request should be written like the following and then each code handling it will either get an object or None

```
   async def send_request(
        self,
        method: str,
        url: str,
        data: Optional[Any] = None,
        params: Optional[dict[str, Any]] = None,
        headers: Optional[dict[str, Any]] = None,
        return_key: Optional[str] = None,
    ) -> dict[str, Any] | None:
        self._client.auth = BasicAuth("", self._personal_access_token)
        self._client.follow_redirects = True

        try:
            response = await self._client.request(
                method=method,
                url=url,
                data=data,
                params=params,
                headers=headers,
            )
            response.raise_for_status()
            if return_key:
                return response.json()[return_key]
            return response.json()
            if response.status_code == 404:
                logger.warning(f"Couldn't access url: {url}. Failed due to 404 error")
            else:
                if response.status_code == 401:
                    logger.error(
                        f"Couldn't access url {url} . Make sure the PAT (Personal Access Token) is valid!"
                    )
                logger.error(
                    f"Request with bad status code {response.status_code}: {method} to url {url}"
                )
                raise e
                
        except httpx.HTTPError as e:
            logger.error(f"Couldn't send request {method} to url {url}: {str(e)}")
            raise e

        return None
```

integrations/azure-devops/azure_devops/client/azure_devops_client.py
```diff
@@ -452,19 +472,19 @@ async def _get_item_content(
...
-            file_content = (
-                await self.send_request(
-                    method="GET", url=items_url, params=items_params
+
+            response = await self.send_request(
+                method="GET", url=items_url, params=items_params
+            )
+            if not response:
+                logger.warning(
+                    f"Couldn't access url {items_url}. This may be because the repo {repository_id} is disabled."
```
mk-armah, line 481:
```suggestion
                    f"Failed to access URL '{items_url}'. The repository '{repository_id}' might be disabled or inaccessible."
```

mk-armah (approved):

Tankilevitch (approved):

PR #1467, opened 2025-03-09
[Core] Wrong integration version showing in UI and API fix

yaelibarg:
# Description

What - fixed wrong integration version showing in UI and API

Why - we did not patch the integration when `OCEAN__INITIALIZE_PORT_RESOURCES=false`

How - moved the check of `OCEAN__INITIALIZE_PORT_RESOURCES=false` to after the patch

## Type of change

### Core testing checklist


Tankilevitch (approved):
lets bump core version and changelog

PR #1502, opened 2025-03-18
[Integration][Gitlab] Improve Memory Usage For Members Kind

mk-armah:
# Description

**What**
This PR optimizes our group member synchronization by ensuring that only unique group members are stored and processed. In addition, caching for groups has been removed to reduce memory overhead.

**Why**
The previous implementation accumulated all group members during pagination, which naturally led to higher memory usage for groups with many members (an O(n) space complexity). Although there was no memory leak, this design resulted in a peak usage during member syncing. With the optimization—deduplicating members in each batch—the peak memory usage was reduced from 105 MB to 84 MB when syncing 30 unique inherited members.

**How**  
- **Batch Processing:** The member enrichment now processes members in batches, filtering out duplicates on the fly before adding them to the final list.  
- **Deduplication:** For each batch, the process checks member IDs and ensures only unique members are stored, reducing unnecessary memory consumption.  
- **Caching Removal:** The caching mechanism on the `get_all_groups` method has been removed, lowering the overall memory footprint.  
- **Logging Enhancements:** Improved logging provides better visibility into the number of processed members per group and overall progress during syncing.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -753,58 +751,100 @@ async def enrich_object_with_members(
...
+        include_verbose_member_object: bool = False,
     ) -> RESTObject:
         """
-        Enriches an object (e.g., Project or Group) with its members and optionally their public emails.
+        Enriches an object (e.g., Project or Group) with its members.
         """
-        members_list = [
-            member
-            async for members in self.get_all_object_members(
+        obj_name = getattr(obj, "name", "unknown")
```
Tankilevitch, line 760: if "unkown" i guess we want to skip it

integrations/gitlab/gitlab_integration/gitlab_service.py
```diff
@@ -753,58 +751,100 @@ async def enrich_object_with_members(
...
+                                        "id": member.id,
+                                        "username": member.username,
+                                        "email": getattr(member, "email", ""),
+                                    }
+                                )
+                    total_members_processed += len(member_chunk)
+
+                logger.info(
+                    f"Processed {total_members_processed} members for {obj_name}"
+                )
```
Tankilevitch, line 800: why you are chunking the members? seems like you can iteratre over the whole member batch and just add it to the member list

integrations/gitlab/gitlab_integration/gitlab_service.py
mk-armah: Lets be explicit instead

mk-armah: I'm sure name should be there in both project and groups

Tankilevitch (approved):

PR #1641, opened 2025-05-19
[Integration][Bitbucket Server] Add Bitbucket Server integration

lordsarcastic:
# Description

What - Add Bitbucket Server integration

Why - To provide Bitbucket Server as an Ocean integration

How -

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+    "pr:reviewer:approved",
+    "pr:comment:added",
+]
+
+
+WEBHOOK_EVENTS = [
+    *PROJECT_WEBHOOK_EVENTS,
+    *REPO_WEBHOOK_EVENTS,
+    *PR_WEBHOOK_EVENTS,
+]
```
mk-armah, line 50: move this to a `webhook/events.py`

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+            webhook_secret: Optional secret for webhook signature verification
+            app_host: Optional host URL for webhook callbacks
+        """
+        self.username = username
+        self.password = password
+        self.base_url = base_url
+        self.bitbucket_auth = BasicAuth(username=username, password=password)
+        self.client = httpx.AsyncClient(
+            timeout=httpx.Timeout(60), auth=self.bitbucket_auth
+        )
```
mk-armah, line 83: is there a reason why we are not using `httpx_async_client` from ocean

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+        self.app_host = app_host
+        self.webhook_secret = webhook_secret
+        # Despite this, being the rate limits, we do not reduce to the lowest common factor because we want to allow as much
+        # concurrency as possible. This is because we expect most users to have resources
+        # synced under one hour.
+        self.rate_limiter = AsyncLimiter(
+            BITBUCKET_RATE_LIMIT, BITBUCKET_RATE_LIMIT_WINDOW
+        )
+
+    async def send_port_request(
```
mk-armah, line 93:
```suggestion
    async def send_api_request(
```

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+        Yields:
+            Batches of project data
+        """
+        logger.info(f"Getting projects with filter: {projects_filter}")
+        if projects_filter:
+            project_batch = await self._get_projects_with_filter(projects_filter)
+            yield project_batch
+        else:
+            async for project_batch in self._get_all_projects():
+                yield project_batch
```
mk-armah, line 221: Looks non-static, how dynamic is it, and when does the cache invalidates

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+            f"projects/{project_key}/repos"
+        ):
+            repositories = []
+            for repository in repo_batch:
+                repositories.append(
+                    await self._enrich_repository_with_readme_and_latest_commit(
+                        repository
+                    )
+                )
+            yield repositories
```
mk-armah, line 265: gather results and yield

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+        Args:
+            project_key: Key of the project
+
+        Returns:
+            List of webhook configurations
+        """
+        webhooks = await self.send_port_request(
+            "GET", f"projects/{project_key}/webhooks"
+        )
+        return webhooks.get("values", [])
```
mk-armah, line 520: This is not a paginated endpoint ?

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+            Created webhook configuration
+        """
+        existing_webhooks = await self.get_repository_webhooks(project_key, repo_slug)
+        existing_webhook_names = set(webhook["name"] for webhook in existing_webhooks)
+        if (
+            self._get_webhook_name(f"{project_key}-{repo_slug}")
+            in existing_webhook_names
+        ):
+            logger.info(
+                f"Webhook for repository {repo_slug} already exists, skipping creation"
```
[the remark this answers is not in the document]
mk-armah: async def _webhook_exists(project) ->  bool: ...

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+class BitbucketClient:
+    """
+    A client for interacting with the Bitbucket Server API.
+    Handles authentication, rate limiting, and provides methods for accessing various Bitbucket resources.
+    """
+
+    def __init__(
+        self,
+        username: str,
+        password: str,
```
mk-armah, line 62: Do we have the option of authenticating via tokens ?

mk-armah (changes requested):
Add test for webhook processors

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,752 @@
...
+                response.raise_for_status()
+                return response.json()
+            except httpx.HTTPStatusError as e:
+                logger.error(
+                    f"HTTP error with status code: {e.response.status_code} and response text: {e.response.text}"
+                )
+                raise
+            except httpx.HTTPError as e:
+                logger.error(f"Failed to send {method} request to url {url}: {str(e)}")
+                raise
```
mk-armah, line 123: handle 404 errors here

integrations/bitbucket-server/webhook_processors/_base.py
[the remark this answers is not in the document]
lordsarcastic: Any reason for the name change specifically? I find this name too verbose without adding any extra information than is already being communicated by `_base`.

integrations/bitbucket-server/client.py
lordsarcastic: They are needed in the `client.py` file. Since the client is a standalone file without dependencies, I wouldn't want to move it under `webhook` which relates only to webhooks

mk-armah (changes requested):

integrations/bitbucket-server/client.py
lordsarcastic: The cache invalidates based on the argument passed to the function itself. So if a different args is passed, it stores that separately with its own key. All items in the cache are invalidated at the end of the resync event

integrations/bitbucket-server/client.py
lordsarcastic: You mean of all the repositories in the project?

integrations/bitbucket-server/webhook_processors/_base.py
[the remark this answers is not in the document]
mk-armah: We are keeping the naming convention consistent in the codebase actually

integrations/bitbucket-server/client.py
lordsarcastic: No, it isn't. [Project webhooks](https://developer.atlassian.com/server/bitbucket/rest/v906/api-group-project/#api-api-latest-projects-projectkey-webhooks-get)

integrations/bitbucket-server/client.py
lordsarcastic: No, [we don't](https://developer.atlassian.com/server/bitbucket/how-tos/example-basic-authentication/#bitbucket-data-center-rest-api-example---basic-authentication)

integrations/bitbucket-server/client.py
mk-armah: Do we need a cache on it if it possibly highly dynamic ? If two kinds possibly do not return the same results I'm not sure they will be a need to cache as the cache results will not be used, can we optimize ?

integrations/bitbucket-server/client.py
mk-armah: Can you move all affairs of webhook creation and management under `webhook`,  so we get webhook client and rest client independent, it's maintainable that way and we prefer to keep the codebase aligned as such, refer to the bitbucket-cloud 🙏🏽

integrations/bitbucket-server/client.py
mk-armah: for repository in repo_batch

lordsarcastic: If we do that, the endpoint is made much slower rather than returning results on demand which is the whole point of using an async generator in the first place

integrations/bitbucket-server/client.py
lordsarcastic: Yeah, you are absolutely correct. However, it isn't highly dynamic since it is at project level which is the highest level. It's based on the projects specified by the user or all the projects in their installation. Pretty much static.

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,768 @@
...
+        async def find_readme_file() -> str | None:
+            file_listing_path = f"projects/{project_key}/repos/{repo_slug}/files"
+            async for batch in self.get_paginated_resource(
+                file_listing_path, page_size=500
+            ):
+                for file_path in batch:
+                    if "/" not in file_path and README_PATTERN.match(file_path):
+                        return file_path
+
+            return None
```
oiadebayo, line 360: Rather than retrieving the entire files in path and checking with a static regex. If you use the `file://` prefix, then you can just retrieve the specific file. This will optimise performance, as you won't need to iterate through the entire files in the path for each repository.

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,768 @@
...
+            True if version is 8.7 or older, False otherwise
+        """
+        application_properties = await self._get_application_properties()
+        # we intentionally do not use get to ensure the integration
+        # fails early if there is a problem
+        version: str = application_properties["version"]
+        # keep only the first two numbers in the version leaving the rest
+        # and converting the result to a float for easy comparison
+        float_version = float(".".join(version.split(".")[:2]))
+        return float_version <= 8.7
```
oiadebayo, line 712: Are we very confident the `application-properties` path is actually available for all versions?

oiadebayo (changes requested):
Great work! I left a few comments

integrations/bitbucket-server/client.py
lordsarcastic: Is there an API that automatically supports the file prefix?

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,768 @@
...
+
+        Args:
+            repository: Repository data to enrich
+
+        Returns:
+            Enriched repository data
+        """
+        repository["__readme"] = await self.get_repository_readme(
+            repository["project"]["key"], repository["slug"]
+        )
```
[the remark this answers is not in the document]
lordsarcastic: Could you explain how the `file://` prefix works? Although, the task was to basically port the bitbucket server script to an Ocean integration keeping the same features and everything in place for backwards compatibility.

Could you point to an existing implementation I can look at?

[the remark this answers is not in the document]
oiadebayo: https://github.com/port-labs/ocean/blob/fdb896f6e9bee4f3f33825bcbdffbdced6754375/integrations/bitbucket-cloud/integration.py#L110-L117

https://github.com/port-labs/ocean/blob/fdb896f6e9bee4f3f33825bcbdffbdced6754375/integrations/gitlab/gitlab_integration/git_integration.py#L22-L40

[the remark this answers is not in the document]
oiadebayo: It works on the JQ level such that you can specify filename (readme or otherwise) of files to import and it will only be fetched if it is added to the mapping and the user can set specific name of file so the match of the entire file in the path to the file name pattern will not be necessary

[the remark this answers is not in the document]
oiadebayo: Running through the entire path for every repo is not performance friendly for the integration, especially considering it's already looping through projects to get the repo and more dependent loop still follows for pull request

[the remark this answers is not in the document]
oiadebayo: Also the script being moved to ocean does not do this loop over all files as well; we just get the specific readme file directly

integrations/bitbucket-server/client.py
oiadebayo: No special api, the file prefix will give you the file name through the mapping and you can retrieve just that

integrations/bitbucket-server/client.py
oiadebayo: It is, the whole reason why this flag was added to the script is because some api endpoints are not available for older versions

integrations/bitbucket-server/pyproject.toml
```diff
@@ -0,0 +1,114 @@
+[tool.poetry]
+name = "bitbucket-server"
+version = "0.1.0"
```
mk-armah, line 3:
```suggestion
version = "0.1.0-beta"
```

integrations/bitbucket-server/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
+# Changelog - Ocean - bitbucket-server
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- towncrier release notes start -->
+
+## 0.1.0 (2025-05-20)
```
mk-armah, line 10:
```suggestion
## 0.1.0-beta (2025-05-20)
```

integrations/bitbucket-server/.port/spec.yaml
```diff
@@ -0,0 +1,34 @@
...
+    description: Bitbucket Server API URL
+  - name: bitbucketWebhookSecret
+    type: string
+    description: Optional secret used to verify incoming webhook requests. Ensures that only legitimate events from Bitbucket are accepted.
+    sensitive: true
+    required: false
+  - name: bitbucketIsVersion8Point7OrOlder
+    type: boolean
+    description: Whether the Bitbucket Server version is 8.7 or older
+    required: false
```
mk-armah, line 34: why not take the version ? any discrepancies to take not of ?

integrations/bitbucket-server/integration.py
```diff
@@ -0,0 +1,57 @@
...
+        default=None,
+        description="List of project keys to filter. If empty, all projects will be synced",
+    )
+
+
+class BitbucketPullRequestSelector(BitbucketGenericSelector):
+    state: Literal["ALL", "OPEN", "MERGED", "DECLINED"] = Field(
+        default="OPEN",
+        description="State of pull requests to sync (ALL, OPEN, MERGED, DECLINED)",
+    )
```
mk-armah, line 32: does the api support passing more than one state . e.g ["OPEN", "MERGED"]

mk-armah: This can allow us to support querying more than one states

mk-armah (changes requested):
Last few comments

integrations/bitbucket-server/.port/spec.yaml
lordsarcastic: The only purpose of this is to know how to setup webhooks. It only has to be provided as a fallback for the endpoint that retrieves the application information, `_get_application_properties`. Not all versions provide that endpoint, hence the need for this.

integrations/bitbucket-server/integration.py
lordsarcastic: Unfortunately, it doesn't

mk-armah (approved):
LGTM

oiadebayo (approved):

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,472 @@
...
+        repository["__latestCommit"] = await self.get_latest_commit(
+            repository["project"]["key"], repository["slug"]
+        )
+        return repository
```
mk-armah, line 213: We agreed to not support this for beta, No ?

integrations/bitbucket-server/client.py
```diff
@@ -0,0 +1,472 @@
...
+        async def find_readme_file() -> str | None:
+            file_listing_path = f"projects/{project_key}/repos/{repo_slug}/files"
+            async for batch in self.get_paginated_resource(
+                file_listing_path, page_size=500
+            ):
+                for file_path in batch:
+                    file_path = cast(str, file_path)
+                    if "/" not in file_path and README_PATTERN.match(file_path):
+                        return file_path
+
```
mk-armah, line 331: Can we support this in later versions ?

mk-armah (changes requested):

Tankilevitch (changes requested):

PR #1670, opened 2025-05-20
[Jira Server] Added token support for the Jira server integration

matarpeles:
Added support for PAT authentication, now it supports both username-password or token authentication


Tankilevitch (approved):
Thanks @matarpeles for your nice contribution 👍

PR #1671, opened 2025-05-20
[Integration][Github] - Implement support for Github App

melodyogonna:
# Description
This add support for Github apps to our Ocean integration for Github.

What - Settings and configurations to authenticate as a Github app.

Why - We want to make the experience of using the Ocean integration as similar as the existing app, this means giving the integration the capability of being installed with our existing app flow. This is the first step to that goal. At the moment this works more like the "self-hosted app" process.

How - This works by subclassing the existing GithubRestClient and adding App-specific functionalities on top. The existing abstraction means the support can be added with minimal changes.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/github/github/clients/auth/abstract_authenticator.py
```diff
@@ -0,0 +1,45 @@
...
+class GitHubToken(BaseModel):
+    token: str
+    expires_at: Optional[datetime] = None
+    _time_buffer: timedelta = PrivateAttr(default_factory=lambda: timedelta(minutes=5))
+
+    @property
+    def is_expired(self) -> bool:
+        if not self.expires_at:
+            return False
+        return datetime.now(timezone.utc) >= (self.expires_at - self._time_buffer)
```
mk-armah, line 18: smoooooth 🙂‍↕️🤝

integrations/github/github/clients/http/graphql_client.py
mk-armah, line 1: move GraphQLClientError to`integrations/github/github/helpers/exceptions.py`

mk-armah (changes requested):
bump integration version and add a changelog

PR #1674, opened 2025-05-21
[Core] Implement caching providers- Introduced disk caching option

matan84:
# Description

What:
- Added base class `CacheProvider` for cache implementations.
- Introduced `DiskCacheProvider` for file-based caching with pickle.
- Introduced `InMemoryCacheProvider` for in-memory caching.
- Integrated caching into the `Ocean` class to select the appropriate provider based on multiprocessing configuration.
- Updated `SyncRawMixin` to clear the cache after execution.
- Added tests for both disk and memory cache providers to ensure functionality and error handling.

Why - Adding caching to disk option with later allowing to add more complex caching options

How - Making the caching providers generic

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


port_ocean/ocean.py
```diff
@@ -86,6 +89,12 @@ def __init__(
         )
 
         self.app_initialized = False
+        self.cache_provider: CacheProvider = self._get_caching_provider()
+
+    def _get_caching_provider(self) -> CacheProvider:
+        if self.config.multiprocessing_enabled:
```
nivm-port, line 95: Is it the only condition that we would like to cache on the disk? maybe allowing to configure it would be more flexible for future needs

port_ocean/cache/disk.py
```diff
@@ -0,0 +1,45 @@
...
+from pathlib import Path
+from typing import Any, Optional
+
+from port_ocean.cache.base import CacheProvider
+
+
+class DiskCacheProvider(CacheProvider):
+    def __init__(self, cache_dir: str | None = None) -> None:
+        if cache_dir is None:
+            cache_dir = ".cache"
```
Tankilevitch, line 11: lets understand what are the conventions for using /tmp, whether we should rely on tmp directory and maybe prefix with `ocean` or `integration` / `processId`?

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -737,3 +737,5 @@ async def sync_raw_all(
                         logger.info("Finished executing resync_complete hooks")
 
                     return True
+            finally:
+                await ocean.app.cache_provider.clear()
```
Tankilevitch, line 741: lets add that to listen to "sigterm" and on initialization?

port_ocean/ocean.py
Tankilevitch: agree with this point, I think we should be able to turn this on even when not using `multiprocessing`. 

maybe `storage_type` = `memory`/ `local` / `postgres` ?

port_ocean/cache/disk.py
```diff
@@ -0,0 +1,45 @@
...
+    async def get(self, key: str) -> Optional[Any]:
+        cache_path = self._get_cache_path(key)
+        if not cache_path.exists():
+            return None
+
+        try:
+            with open(cache_path, "rb") as f:
+                return pickle.load(f)
+        except (pickle.PickleError, EOFError):
+            raise RuntimeError(f"Failed to read cache file: {cache_path}")
```
Tankilevitch, line 27: its doesn't seems actually async, can we do it async?

port_ocean/cache/disk.py
matan84: Regarding tmpfile,the files are mainly used with a context manager, which we shouldn't really create here (will increase complexity IMO)
Using the `finally` should serve the same purpose. However, I renamed the caching dir from `.cache` to `.ocean_cache`

port_ocean/cache/disk.py
matan84: There's no way to pickle using asyncio

port_ocean/cache/disk.py
```diff
@@ -0,0 +1,47 @@
...
+
+    async def set(self, key: str, value: Any) -> None:
+        cache_path = self._get_cache_path(key)
+        try:
+            with open(cache_path, "wb") as f:
+                pickle.dump(value, f)
+        except (pickle.PickleError, IOError) as e:
+            raise RuntimeError(f"Failed to write cache file: {e}")
+
```
Tankilevitch, line 38: same here, I am not sure if we want to just fail the whole resync process due to some issue in caching ( I am okay with it being raised here, but we need to make sure we are handling it in the `utils/cache.py` handler

Tankilevitch (approved):
LGTM

port_ocean/config/settings.py
```diff
@@ -93,6 +93,7 @@ class IntegrationConfiguration(BaseOceanSettings, extra=Extra.allow):
     )
     max_event_processing_seconds: float = 90.0
     max_wait_seconds_before_shutdown: float = 5.0
+    caching_storage_type: Optional[str] = Field(default=None)
```
Tankilevitch, line 96: this needs to be enum or referenced to a type

PR #1799, opened 2025-06-18
[Core] Added `on_start` for specific event listeners

Tankilevitch:
# Description

What - Added `on_start` for specific event listeners

Why - currently integrations are specifying inside of the code what to do in case of each event listener, this causes quite a few problems as we might want to have different on_start capabilities for different eventListener types.

How - added a configurable parameter to the `on_start` method that allows the user to specify the event listener type

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/port_ocean/core/integrations/base.py
+++ b/port_ocean/core/integrations/base.py
@@ -72,7 +72,24 @@ class BaseIntegration(SyncRawMixin, SyncMixin):
 
         self.started = True
 
-        if self.event_strategy["start"]:
+        # Collect start functions to execute
+        start_functions = []
+        current_listener_type = self.context.config.event_listener.type
+
+        # Get event listener-specific start functions (preferred)
+        listener_specific_functions = self.event_strategy["start_for_listener"].get(current_listener_type, [])
+
+        if listener_specific_functions:
+            # Use event listener-specific functions if available
+            logger.debug(f"Using {len(listener_specific_functions)} event listener-specific start functions for {current_listener_type}")
+            start_functions.extend(listener_specific_functions)
+        else:
+            # Fall back to general start functions for backwards compatibility
+            if self.event_strategy["start"]:
+                logger.debug(f"Using {len(self.event_strategy['start'])} general start functions (backwards compatibility)")
+                start_functions.extend(self.event_strategy["start"])
+
+        if start_functions:
             async def run_on_start_tasks() -> None:
                 try:
                     async with event_context(
@@ -80,7 +97,7 @@ class BaseIntegration(SyncRawMixin, SyncMixin):
                         trigger_type="machine",
                     ):
                         await asyncio.gather(
-                            *(listener() for listener in self.event_strategy["start"])
+                            *(listener() for listener in start_functions)
                         )
                 except Exception as e:
                     logger.exception("Error in start event listeners: %s", str(e))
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+## 0.24.12 (2025-06-18)
```
matan84, line 56:
```suggestion
## 0.25.0 (2025-06-18)
```

port_ocean/core/integrations/mixins/events.py
```diff
@@ -34,11 +36,58 @@ def __init__(self) -> None:
...
 
-    def on_start(self, function: START_EVENT_LISTENER) -> START_EVENT_LISTENER:
-        """Register a function as a listener for the "start" event."""
-        logger.debug(f"Registering {function} as a start event listener")
-        self.event_strategy["start"].append(function)
-        return function
+    def on_start(
+        self,
+        function: START_EVENT_LISTENER | None = None,
+        event_listener: EventListenerType | None = None
```
matan84, line 42: It should also be a list 
For example, on start which starts webhooks should be both for POLLING and WEBHOOKS_ONLY

port_ocean/core/integrations/base.py
```diff
@@ -72,15 +72,32 @@ async def start(self) -> None:
...
+                start_functions.extend(self.event_strategy["start"])
+
+        if start_functions:
             async def run_on_start_tasks() -> None:
                 try:
                     async with event_context(
                         EventType.START,
                         trigger_type="machine",
                     ):
                         await asyncio.gather(
```
matan84, line 99: Maybe we can also add here the "Don't fail if on_start fails"? or do you want it to be per event listener (for example, POLLING can ignore on_start errors but WEBHOOKS_ONLY should fail)

port_ocean/core/integrations/base.py
```diff
@@ -72,15 +72,32 @@ async def start(self) -> None:
...
-        if self.event_strategy["start"]:
+        # Collect start functions to execute
+        start_functions = []
+        current_listener_type = self.context.config.event_listener.type
+
+        # Get event listener-specific start functions (preferred)
+        listener_specific_functions = self.event_strategy["start_for_listener"].get(current_listener_type, [])
+
+        if listener_specific_functions:
+            # Use event listener-specific functions if available
```
matan84, line 83: Please remove the intermediate comments

matan84 (changes requested):
Left a few comments

port_ocean/core/integrations/mixins/events.py
Tankilevitch: see tests we have an example of a method being decorated a few times.

yaelibarg (commented):
A general question - don't you think its over using decorators? 
They allows you to add new behaviors to objects dynamically without **altering their structure** but if I still need 2 different functions for different types of on_start , i would solve it by applying strategy design pattern, not NOT decorators

PR #1841, opened 2025-06-26
[Integration][Gitlab-V2] Implementation of File Deletion Support in GitLab v2 Integration

dev-habib-nuhu:
# Description

What -

- Added support for handling deleted files in GitLab webhook events for the File kind. The implementation properly processes file deletions from push events and reports them to Port's platform.

Why -

- Previously, the integration was only handling added and modified files, but wasn't properly tracking deleted files. This meant that when files were deleted in GitLab, their state wasn't being accurately reflected in Port. This implementation ensures data consistency between GitLab and Port by properly tracking file deletions.

How -

- Enhanced the file push webhook processor to track deleted files
- Collect deleted files from commit data using `commit.get("removed", [])`
- Use the appropriate commit reference (before hash for deleted files)

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/integrations/gitlab-v2/gitlab/webhook/webhook_processors/file_push_webhook_processor.py
+++ b/integrations/gitlab-v2/gitlab/webhook/webhook_processors/file_push_webhook_processor.py
@@ -43,21 +43,33 @@ class FilePushWebhookProcessor(_GitlabAbstractWebhookProcessor):
                 updated_raw_results=[], deleted_raw_results=[]
             )
 
-        changed_files = set()
-        for commit in payload.get("commits", []):
-            changed_files.update(commit.get("modified", []))
-            changed_files.update(commit.get("added", []))
+        added_files = set()
+        modified_files = set()
+        removed_files = set()
 
+        for commit in payload.get("commits", []):
+            added_files.update(commit.get("added", []))
+            modified_files.update(commit.get("modified", []))
+            removed_files.update(commit.get("removed", []))
+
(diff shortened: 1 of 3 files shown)


integrations/gitlab-v2/gitlab/webhook/webhook_processors/file_push_webhook_processor.py
```diff
@@ -43,21 +43,33 @@ async def handle_event(
...
-        changed_files = set()
+        added_files = set()
+        modified_files = set()
+        removed_files = set()
+
         for commit in payload.get("commits", []):
-            changed_files.update(commit.get("modified", []))
-            changed_files.update(commit.get("added", []))
+            added_files.update(commit.get("added", []))
+            modified_files.update(commit.get("modified", []))
```
shariff-6: I noticed we’re collecting added_files and modified_files separately, but it looks like we always process them together as changed_files. Is there a reason we might want to keep them, or could we simplify by just using one set, downstream we only use the changed files set.

shariff-6 (approved):
LGTM

Tankilevitch (commented):
you are missing tests to actually validate the new added capability

Tankilevitch (approved):

PR #1930, opened 2025-07-28
[Bug] [Integration] [Bitbucket] Fetch paginated API not working

matiri132:
# Description

**What** -  
Fixes a bug in the Bitbucket Cloud integration where only the first page of API results was fetched due to incorrect handling of pagination.

**Why** -  
Bitbucket’s paginated API returns a `next` field with the full URL to the next page. The current implementation does not parse this properly, which prevents full ingestion when more than one page of data is available (e.g., workspaces with >100 repositories).

**How** -  
Instead of overriding `url` with `next`, the function now parses the query string in `next` and merges its parameters into the current `params` dict. This preserves existing request parameters (like `fields`) while updating only the page-related values.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [ ] Support Multi account

## Screenshots

N/A – No visual changes.

## API Documentation

https://developer.atlassian.com/bitbucket/api/2/reference/meta/pagination

### DIFF
--- a/integrations/bitbucket-cloud/bitbucket_cloud/client.py
+++ b/integrations/bitbucket-cloud/bitbucket_cloud/client.py
@@ -1,5 +1,6 @@
 from typing import Any, AsyncGenerator, Optional
 from httpx import HTTPError, HTTPStatusError
+from urllib.parse import urlparse, parse_qs
 from loguru import logger
 from port_ocean.utils import http_async_client
 from bitbucket_cloud.helpers.exceptions import MissingIntegrationCredentialException
@@ -109,9 +110,18 @@ class BitbucketClient:
                 )
                 if values := response.get(data_key, []):
                     yield values
-                url = response.get("next")
-                if not url:
+                next_url = response.get("next")
+                if not next_url:
                     break
+
+                # Get params from next url
+                parsed = urlparse(next_url)
+                query = parse_qs(parsed.query)
+                for key, value in query.items():
+                    if len(value) == 1:
+                        params[key] = int(value[0]) if value[0].isdigit() else value[0]
+                    else:
+                        params[key] = value
 
     async def _send_paginated_api_request(
         self,
@@ -137,9 +147,18 @@ class BitbucketClient:
             response = await self._send_api_request(url, params=params, method=method)
             if values := response.get(data_key, []):
                 yield values
-            url = response.get("next")
-            if not url:
+            next_url = response.get("next")
+            if not next_url:
                 break
+
+            # Get params from next url
+            parsed = urlparse(next_url)
+            query = parse_qs(parsed.query)
+            for key, value in query.items():
+                if len(value) == 1:
+                    params[key] = int(value[0]) if value[0].isdigit() else value[0]
+                else:
+                    params[key] = value
 
     async def get_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
         """Get all projects in the workspace."""
--- a/changelog/1929.bugfix.md
+++ b/changelog/1929.bugfix.md
@@ -1,0 +1,1 @@
+Fix Bitbucket Cloud pagination using merged params from next URL


avalojuan:
This integration is currently missing error handling. 
When the rate limiter allows a request through, if that request results in a 429 error, the entire sync fails.
[_send_api_request error](https://github.com/matiri132/ocean/blob/c3bd965de43598cf60658e21e83884f86f54d1bb/integrations/bitbucket-cloud/bitbucket_cloud/client.py#L82)

avalojuan:
To add more context to my previous message, the `client.py` file implements a function that utilizes a rate limiter decorator to fetch entities from Bitbucket
This decorator tracks when a Sync started and, if the configured rate limit is exceeded, it awaits a specified period before attempting to fetch data again.

The problem occurs when this rate limiter loses track of the actual limits (e.g., by running multiple Syncs in quick succession). 
In such cases, it may allow a request to the Bitbucket API, which then returns a 429 (Too Many Requests) error. Then, the Sync process fails at this point without any built-in retry mechanism.

To fix this, I propose adding a retry logic for 429 errors inside the [_send_api_request](https://github.com/port-labs/ocean/blob/137562cb58a949ee2693d6096b4990fca72ea010/integrations/bitbucket-cloud/bitbucket_cloud/client.py#L66) function.

We deployed an image with the following implementation of the retry mechanism, that starts with an await of 30 seconds and duplicates this number on each retry:
```python
    async def _send_api_request(
        self,
        url: str,
        params: Optional[dict[str, Any]] = None,
        json_data: Optional[dict[str, Any]] = None,
        method: str = "GET",
        return_full_response: bool = False,
    ) -> Any:
        max_retries = 10  # Maximum number of retries
        base_delay = 3.5  # Base delay for exponential backoff in seconds
        retries = 0
        while retries <= max_retries:
            try:
                """Send request to Bitbucket API with error handling."""
                response = await self.client.request(
                    method=method, url=url, params=params, json=json_data
                )
                response.raise_for_status()
                return response if return_full_response else response.json()
            except HTTPStatusError as e:
                if e.response.status_code == 429:
                    retries += 1
                    if retries > max_retries:
                        logger.error(f"Bitbucket API error: {str(e)} - Max retries exceeded.")
                        raise e # Re-raise if max retries reached
                    
                    # Exponential backoff: 1s, 2s, 4s, 8s, 16s...
                    wait_time = base_delay * (2 ** (retries - 1))
                    logger.warning(
                        f"Rate limited (429) for {url}. Waiting {wait_time:.2f} seconds. (Retry {retries}/{max_retries})"
                    )

                    jitter = time.time() % 0.5
                    await asyncio.sleep(wait_time + jitter)

                elif e.response.status_code == 404:
                    logger.warning(
                        f"Requested resource not found: {url}; message: {str(e)}"
                    )
                    return {}
                else:
                    logger.error(f"Bitbucket API error: {str(e)}")
                    raise e
```

I can add this modification to the current PR if it's ok.
cc: @oiadebayo

integrations/bitbucket-cloud/bitbucket_cloud/client.py
```diff
@@ -109,10 +128,13 @@ async def _fetch_paginated_api_with_rate_limiter(
...
                 if values := response.get(data_key, []):
                     yield values
-                url = response.get("next")
-                if not url:
+                next_url = response.get("next")
+                if not next_url:
                     break
 
+                next_params = self._parse_next_url_params(next_url)
+                params.update(next_params)
```
oiadebayo, line 136: Thank you for the proposed solution. Here is my thought about the proposed solution:

Rather than parsing the url, can we simply `None` the `query_param` in subsequent call as that will prevent the need to transform, validate or sanitize the data in the query param which bitbucket strongly advice against in the [docs](https://developer.atlassian.com/cloud/bitbucket/rest/intro/#pagination). I also believe that url already contains the entire query parameter needed to fetch the next page so the merge/update might not be necessary. Kindly let me know what you think.

oiadebayo:
> To add more context to my previous message, the `client.py` file implements a function that utilizes a rate limiter decorator to fetch entities from Bitbucket This decorator tracks when a Sync started and, if the configured rate limit is exceeded, it awaits a specified period before attempting to fetch data again.
> 
> The problem occurs when this rate limiter loses track of the actual limits (e.g., by running multiple Syncs in quick succession). In such cases, it may allow a request to the Bitbucket API, which then returns a 429 (Too Many Requests) error. Then, the Sync process fails at this point without any built-in retry mechanism.
> 
> To fix this, I propose adding a retry logic for 429 errors inside the [_send_api_request](https://github.com/port-labs/ocean/blob/137562cb58a949ee2693d6096b4990fca72ea010/integrations/bitbucket-cloud/bitbucket_cloud/client.py#L66) function.
> 
> We deployed an image with the following implementation of the retry mechanism, that starts with an await of 30 seconds and duplicates this number on each retry:
> 
> ```python
>     async def _send_api_request(
>         self,
>         url: str,
>         params: Optional[dict[str, Any]] = None,
>         json_data: Optional[dict[str, Any]] = None,
>         method: str = "GET",
>         return_full_response: bool = False,
>     ) -> Any:
>         """Send request to Bitbucket API with error handling."""
>         response = await self.client.request(
>             method=method, url=url, params=params, json=json_data
>         )
>         max_retries = 10  # Maximum number of retries
>         base_delay = 30  # Base delay for exponential backoff in seconds
>         retries = 0
>         while retries <= max_retries:
>             try:
>                 response.raise_for_status()
>                 return response if return_full_response else response.json()
>             except HTTPStatusError as e:
>                 if e.response.status_code == 429:
>                     retries += 1
>                     if retries > max_retries:
>                         logger.error(f"Bitbucket API error: {str(e)} - Max retries exceeded.")
>                         raise e # Re-raise if max retries reached
>                     
>                     # Exponential backoff: 1s, 2s, 4s, 8s, 16s...
>                     wait_time = base_delay * (2 ** (retries - 1))
>                     logger.warning(
>                         f"Rate limited (429) for {url}. Waiting {wait_time:.2f} seconds. (Retry {retries}/{max_retries})"
>                     )
> 
>                     jitter = time.time() % 0.5
>                     await asyncio.sleep(wait_time + jitter)
> 
>                 elif e.response.status_code == 404:
>                     logger.warning(
>                         f"Requested resource not found: {url}; message: {str(e)}"
>                     )
>                     return {}
>                 else:
>                     logger.error(f"Bitbucket API error: {str(e)}")
>                     raise e
> ```
> 
> I can add this modification to the current PR if it's ok. cc: @oiadebayo

Hi @avalojuan, thank you for mentioning this. The fact that the integration sometimes loose track of the current rate limit of the api and assume full quota on start is a limitation of the bitbucket API itself which lacks transparency around the specific quota available at any point in time. This could be problematic in cases where resyncs are done in quick successions as you have rightly pointed out.

We have a retry mechanism already implemented in the ocean core for all 429 errors which uses exponential backoff algorithm to retry the request until a reasonable deadline is met and then raise the error if it fails ([reference](https://github.com/port-labs/ocean/blob/eb6e6aa656896078ccdba20465ab47712e1fd34e/port_ocean/helpers/retry.py#L23)) the multiple retry. This is the reason why we are not doing the retry at the integration level. The final propagation of the error is the reason why the sync of the kind fails.

I however agree with you that the rate limiter and by extension the integration could benefit from being 429 aware. Also I will suggest we keep the concerns separate so another PR and issue can track this.

PR #1938, opened 2025-07-31
[Integration][Github] Fetch folders with search API

melodyogonna:
# Description

Rather than making individual requests to fetch the contents of specified repositories, use the Github search API to fetch all repository contents at once.

Why - We need to make as few requests as we can

How -

## Type of change

- [ x ] Non-breaking change (fix of existing functionality that will not change current behavior)

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/github/github/core/exporters/folder_exporter.py
```diff
@@ -9,15 +12,78 @@
...
+                return
+
+            yield " OR ".join([f"{r} in:name" for r in chunk])
+            chunk = [repo]
+        else:
+            chunk = new_chunk
+
+    if chunk:
+        yield " OR ".join([f"{r} in:name" for r in chunk])
+
```
mk-armah: It's a string, do we need to yield, why not return ?, the function can take a repo at a time and return its search param

melodyogonna: @mk-armah then I would have had to implement the logic that ensures the final search string fits into GitHub's limit somewhere else. Here it takes a list of repository names and returns the search string for them, it is yielding because it may potentially have to return multiple search strings (when you factor Github's search query limitation)

melodyogonna:
Yeah, just did a simple check with a mapping that should be in the best case of "without search" option. 
The "With search API" option is approximately 46.3% faster than the "Without search API" option.

I imagine this gap will increase in more extreme cases (e.g the same repository but searching different paths or just ingesting from a lot of repositories)

melodyogonna:
- With search api - 14.43s
- Without search api - 26.93s

Mapping:
```yaml
deleteDependentEntities: true
createMissingRelatedEntities: true
enableMergeEntity: true
resources:
  - kind: folder
    selector:
      query: 'true'
      folders:
        - path: '*'
          repos:
            - name: jobman
              branch: main
            - name: medium-repo
        - path: pkg/*
          repos:
            - name: jobman
              branch: main
        - path: extensions/*
          repos:
            - name: vscode
              branch: main
    port:
      entity:
        mappings:
          identifier: .folder.path | split("/") | last
          title: .folder.path | split("/") | last
          blueprint: '"folders"'
          properties:
            path: .folder.path

```

integrations/github/github/core/exporters/folder_exporter.py
mk-armah: Then function has too many exit points, it makes it difficult to follow through, enforce single responsibility

mk-armah (changes requested):

integrations/github/github/core/exporters/folder_exporter.py
melodyogonna: @mk-armah I've updated it to use a list rather than a generator. It is basically a function that tries to build valid Github search string, if the string gets too long it is chunked into multiple valid search strings.

integrations/github/github/core/exporters/folder_exporter.py
```diff
@@ -83,15 +184,19 @@ def _filter_folder_contents(
...
-        folders = self._filter_folder_contents(tree, options["path"])
-        logger.info(f"fetched {len(folders)} folders from {options['repo']['name']}")
-        if folders:
-            formatted = self._enrich_folder_with_repository(
-                folders, repo=options["repo"]
-            )
-            return formatted
-        else:
-            return []
+    async def _search_for_repositories(
```
matan84: Do we want to cache this? Using it for each folder kind call seems kind of expensive in api calls

matan84: + Why is this method specifically in `folder_exporter.py` Can we put it in a utils folder?

integrations/github/github/core/exporters/folder_exporter.py
```diff
@@ -9,6 +12,75 @@
 from github.helpers.utils import IgnoredError
 from wcmatch import glob
 
+from integration import FolderSelector
+
+_DEFAULT_BRANCH = "hard_to_replicate_name"
```
matan84, line 16: What is this?

melodyogonna: It is used as a key when branches are not passed. The idea is to use a name with low chances of colliding with a real branch name.

matan84: Cool, Lets make sure we leave a comment explaining this

matan84 (approved):
LGTM

Issue #1929, opened 2025-07-28
[Bug] [Integration] [Bitbucket] Fetch paginated API not working


matiri132:
### Ocean Framework Version

v0.26.1

### Integration Version

v0.1.53

### Steps to reproduce

1. Have the Bitbucket Cloud integration installed and configured for a workspace with more than 100 repositories.
2. Let the integration fetch data from `https://api.bitbucket.org/2.0/repositories/<workspace>?pagelen=100`.
3. Observe the ingestion result.

### What did you expect to see?

The integration should fetch all pages of results using the Bitbucket paginated API and ingest all repositories from the workspace.

### What did you see instead?

Only the first page of results is fetched. The exporter does not proceed to the next pages returned in the `"next"` field of the Bitbucket API response. This leads to partial ingestion and missing repositories.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

matiri132:
PR: https://github.com/port-labs/ocean/pull/1930

50-Course:
hi there, shouldn't this be marked as closed/resolved since there's been a patch to resolve the issue with no newer complaints? we could always reopen this if something pops-up

PR #1989, opened 2025-08-13
[Integration][Bitbucket-cloud] Fixed issue with pagination in Bitbucket Cloud

oiadebayo:
# Description

What - Fixed issue with pagination in bitbucket cloud

Why - 

How - Reset params in the pagination loop to ensure the param in the url did not get overwritten 

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

This PR fixes issue raised [here](https://github.com/port-labs/ocean/issues/1929) and builds on the work done in this [PR](https://github.com/port-labs/ocean/pull/1930)

### DIFF
--- a/integrations/bitbucket-cloud/bitbucket_cloud/client.py
+++ b/integrations/bitbucket-cloud/bitbucket_cloud/client.py
@@ -133,6 +133,7 @@ class BitbucketClient:
             if values := response.get(data_key, []):
                 yield values
             url = response.get("next")
+            params = None
             if not url:
                 break
 
@@ -191,6 +192,7 @@ class BitbucketClient:
             url = response.get("next")
             if not url:
                 break
+            params = None
 
     async def get_projects(self) -> AsyncGenerator[list[dict[str, Any]], None]:
         """Get all projects in the workspace."""
--- a/integrations/bitbucket-cloud/CHANGELOG.md
+++ b/integrations/bitbucket-cloud/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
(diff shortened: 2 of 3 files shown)


integrations/bitbucket-cloud/bitbucket_cloud/client.py
```diff
@@ -191,6 +192,7 @@ async def _send_paginated_api_request(
             url = response.get("next")
             if not url:
                 break
+            params = None
```
yaelibarg, line 195: after the first loop, params will always be None, is this ok?

integrations/bitbucket-cloud/bitbucket_cloud/client.py
```diff
@@ -133,6 +133,7 @@ async def _fetch_paginated_api_with_rate_limiter(
             if values := response.get(data_key, []):
                 yield values
             url = response.get("next")
+            params = None
```
yaelibarg, line 136: after the first loop, params will always be Nonw, is that  ok?

integrations/bitbucket-cloud/bitbucket_cloud/client.py
oiadebayo: Yes, it's ok. This [documentation](https://developer.atlassian.com/cloud/bitbucket/rest/intro/#pagination) recommends that the next `link should be treated as an opaque location that is not to be constructed by clients or even assumed to be predictable` therefore all require parameter are in the link and it is better handled this way than parsing the url

integrations/bitbucket-cloud/bitbucket_cloud/client.py
oiadebayo: Same response as above

PR #2351, opened 2025-10-28
[Integration][Jira] Fixed JIRA integration failure when users remove the JQL query

dev-habib-nuhu:
# Description

**What** - Fixed JIRA integration failure when users remove the JQL query from the integration mapping configuration.

**Why** - When customers completely removed the JQL filter from their JIRA integration configuration, the integration would break because:
1. The JIRA `/rest/api/3/search` endpoint has been discontinued by Atlassian (returns 410 Gone)
2. The `/rest/api/3/search/jql` endpoint requires a valid JQL query and does not support empty or whitespace-only values

**How** - 
- Added a default JQL value (`project is not EMPTY ORDER BY created DESC`) to the `JiraIssueSelector` in `overrides.py` to ensure there's always a valid query that returns all issues
- Implemented validation in `main.py` to detect empty/whitespace-only JQL strings from user configuration and replace them with the default JQL
- Always use `/search/jql` endpoint since `/search` has been discontinued
- Updated test coverage to verify the default JQL is properly applied when users don't provide a custom query

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/integrations/jira/pyproject.toml
+++ b/integrations/jira/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "jira"
-version = "0.5.17"
+version = "0.5.18"
 description = "Integration to bring information from Jira into Port"
 authors = ["Mor Paz <mor@getport.io>"]
 


integrations/jira/jira/overrides.py
```diff
@@ -22,7 +22,10 @@ class TeamResourceConfig(ResourceConfig):
 
 
 class JiraIssueSelector(Selector):
-    jql: str | None = None
+    jql: str = Field(
+        default="project is not EMPTY ORDER BY created DESC",
+        description="JQL query to filter issues. Defaults to fetching all issues across all projects.",
+    )
```
mk-armah, line 28: does this match the default in the `port app config` ?

dev-habib-nuhu: We  don't have defaults in the `port app config`

mk-armah: From the config we have this ` jql: "(statusCategory != Done) OR (created >= -1w) OR (updated >= -1w)"`

mk-armah (changes requested):

mk-armah (approved):
LGTM

PR #2503, opened 2025-12-02
[Core] fix unhashable dict error in webhook processor

melodyogonna:
# Description

What - Fix crash in webhook processor when entity identifier is a dict.

Why - This is preventing some users from receiving live events.

How - First, add proper type to the Entity model. then add an utility method to return a string version of the identifier encoded as json when it's a dict. I also had to update a few places where entity identifier being dict was not factored (due to mypy errors)

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "port-ocean"
-version = "0.31.2"
+version = "0.31.3"
 description = "Port Ocean is a CLI tool for managing your Port projects."
 readme = "README.md"
 homepage = "https://app.getport.io"


port_ocean/core/models.py
```diff
@@ -55,7 +56,7 @@ class PortAPIErrorMessage(Enum):
 
 
 class Entity(BaseModel):
-    identifier: Any
+    identifier: str | dict[str, Any]
```
yaelibarg: I would leave it as Any for edge cases

port_ocean/core/handlers/entities_state_applier/port/order_by_entities_dependencies.py
```diff
@@ -8,7 +8,7 @@
 
 
 def node(entity: Entity) -> Node:
-    return entity.identifier, entity.blueprint
+    return entity.get_identifier_as_string(), entity.blueprint
```
yaelibarg: how can you be sure we need it as string at this point and not as `dict` for example?

port_ocean/core/utils/entity_topological_sorter.py
```diff
@@ -45,7 +45,7 @@ def get_entities(self, sorted: bool = True) -> Generator[Entity, Any, None]:
 
     @staticmethod
     def node(entity: Entity) -> Node:
-        return entity.identifier, entity.blueprint
+        return entity.get_identifier_as_string(), entity.blueprint
```
yaelibarg: how can you be sure we need it as string at this point and not as `dict` for example?

port_ocean/clients/port/mixins/entities.py
```diff
@@ -456,7 +458,7 @@ async def delete_entity(
                 f"Delete entity: {entity.identifier} of blueprint: {entity.blueprint}"
             )
             response = await self.client.delete(
-                f"{self.auth.api_url}/blueprints/{entity.blueprint}/entities/{quote_plus(entity.identifier)}",
+                f"{self.auth.api_url}/blueprints/{entity.blueprint}/entities/{quote_plus(str(entity.identifier))}",
```
yaelibarg: this change is unnecessary because we cannot delete using search identifier

port_ocean/core/models.py
melodyogonna: This imprecise type is causing bugs, this is the second time I've had to fix a type-related bug caused by having `Any` on the entity identifier. Port's API says the identifier can be one of string or dict, I think this should be enforced.

port_ocean/core/handlers/entities_state_applier/port/order_by_entities_dependencies.py
melodyogonna: `Node` is a type alias of `tuple[str, str]`

port_ocean/core/utils/entity_topological_sorter.py
melodyogonna: Node is a type alias of `tuple[str, str]`

port_ocean/clients/port/mixins/entities.py
melodyogonna: Indeed, I had to explicitly cast to str to please the type checker.

PR #2902, opened 2026-03-09
[Integration][GitLab V2] Add support for project members (direct + inherited)

edenhermelin:
Add a new `project-with-members` kind that fetches project members via GitLab's /projects/:id/members and /projects/:id/members/all endpoints, with support for includeInheritedMembers and includeBotMembers selectors.

# Description

What - Add support for fetching project members (direct and inherited) in the GitLab V2 integration via a new `project-with-members` resource kind.

Why - Users need visibility into who has access to each GitLab project, including inherited members from parent groups. This was requested in the [Port roadmap](https://roadmap.port.io/ideas/p/support-for-inherited-group-members-and-projects-members-including-inherited-in).

How -
- New `PROJECT_WITH_MEMBERS` ObjectKind and `GitlabProjectMemberSelector` config model (extends `Selector` with `includeBotMembers`, `includeInheritedMembers`, `includeOnlyActiveProjects`)
- `get_project_members()` and `enrich_project_with_members()` client methods using GitLab's `/projects/:id/members` and `/projects/:id/members/all` endpoints
- Resync handler that fetches projects, enriches each with `__members`, and yields them in batches
- `ProjectWithMemberWebhookProcessor` for live events (project create/destroy + project member add/remove/update)
- `gitlabMembers` relation added to the `service` blueprint targeting `gitlabMember`
- The kind is **opt-in** (not in default `port-app-config.yml`) to avoid duplicate service upserts with the existing `project` kind

## Type of change

- [x] Docs PR link [here](https://github.com/port-labs/port-docs/pull/3633)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [x] Support Multi account

## API Documentation

- [GitLab Project Members API](https://docs.gitlab.com/ee/api/members.html)
- [GitLab Member Events (webhooks)](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#member-events)


mk-armah (approved):
LGTM

integrations/gitlab-v2/main.py
```diff
@@ -360,6 +364,39 @@ async def on_resync_members(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+
+    async for projects_batch in client.get_projects(
+        params=build_project_params(
+            include_only_active_projects=include_only_active_projects
+        ),
+        max_concurrent=DEFAULT_MAX_CONCURRENT,
+        include_languages=False,
+    ):
+        for i in range(0, len(projects_batch), RESYNC_MEMBERS_BATCH_SIZE):
+            current_batch = projects_batch[i : i + RESYNC_MEMBERS_BATCH_SIZE]
```
nivm-port: use batched from itertools it is more convenient

integrations/gitlab-v2/main.py
```diff
@@ -360,6 +364,39 @@ async def on_resync_members(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+            )
+
+            tasks = [
+                client.enrich_project_with_members(
+                    project, include_bot_members, include_inherited_members
+                )
+                for project in current_batch
+            ]
+            results = await asyncio.gather(*tasks)
+            yield results
```
nivm-port: I don't like this section... we are keeping the generator of the project open while doing some heavy tasks in concurrent I think it would be better to collect all the projects into an array and then iterating over them to enrich them

integrations/gitlab-v2/gitlab/clients/gitlab_client.py
```diff
@@ -633,6 +633,56 @@ async def enrich_group_with_members(
...
+        project_id: str,
+        include_bot_members: bool,
+        include_inherited_members: bool,
+    ) -> AsyncIterator[list[dict[str, Any]]]:
+        members_api = "members/all" if include_inherited_members else "members"
+        logger.info(f"Fetching members for project {project_id} with {members_api} API")
+        async for batch in self.rest.get_paginated_project_resource(
+            str(project_id), members_api
+        ):
+            if batch:
```
nivm-port: invert if to reduce nesting

integrations/gitlab-v2/gitlab/clients/gitlab_client.py
```diff
@@ -633,6 +633,56 @@ async def enrich_group_with_members(
...
+                filtered_batch = batch
+                if not include_bot_members:
+                    filtered_batch = [
+                        member
+                        for member in batch
+                        if "bot" not in member["username"].lower()
+                    ]
```
nivm-port: I think it is a better way to do so 
```suggestion
                    filtered_batch = [
                        member
                        for member in batch
                        if "bot" not in member["username"].lower() or include_bot_members
                    ]
```
but I see a potential bug here how do we prevent it from filter users like 'botner-elay'?

nivm-port (changes requested):
See comments :)

CHANGELOG.md
```diff
@@ -6,12 +6,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- towncrier release notes start -->
```
edenhermelin, line 8:
```suggestion
<!-- towncrier release notes start -->
## 0.38.27 (2026-03-30)

### Improvements

- Updated GH token used in CI

```

melodyogonna (approved):
LGTM

nivm-port (approved):
LGTM

integrations/gitlab-v2/integration.py
```diff
@@ -139,6 +139,29 @@ class GitlabMemberResourceConfig(ResourceConfig):
...
+    include_inherited_members: bool = Field(
+        alias="includeInheritedMembers",
+        default=False,
+        description="If set to true, the integration will include inherited members in the project members list. Default value is false",
+    )
+
+
+class GitlabProjectWithMembersResourceConfig(ResourceConfig):
+    kind: Literal["project-with-members"]
+    selector: GitlabProjectMemberSelector
```
melodyogonna: Due to the recent mapping form in the UI, we now require all kinds and properties to have a title and description.

integrations/gitlab-v2/integration.py
```diff
@@ -139,6 +139,29 @@ class GitlabMemberResourceConfig(ResourceConfig):
     )
 
 
+class GitlabProjectMemberSelector(Selector):
+    include_only_active_projects: Optional[bool] = Field(
```
melodyogonna, line 143: These selector properties are missing titles

melodyogonna (changes requested):
We have some missing titles and properties, take a look at `integration.py` in the Github integration to see some examples

melodyogonna (approved):
LGTM

integrations/gitlab-v2/gitlab/webhook/webhook_processors/project_with_member_webhook_processor.py
```diff
@@ -0,0 +1,114 @@
+import re
+import typing
+from typing import Union, cast
```
raskinmaya: Let's consolidate these

integrations/gitlab-v2/gitlab/clients/gitlab_client.py
```diff
@@ -651,6 +672,55 @@ async def enrich_group_with_members(
...
+            for member in members_batch:
+                members.append(
+                    {
+                        "email": member.get("email"),
+                        "username": member["username"],
+                        "name": member["name"],
+                        "access_level": member["access_level"],
+                        "id": member["id"],
+                    }
+                )
```
raskinmaya: We rely on `member["username"]`, `member["name"]`, and `member["access_level"]`, which GitLab typically provides, but invited/pending or partially materialized members can be a source of API issues. Would be better to use `.get()` with a clear skip/log, or validate before appending, so a single malformed entry doesn’t break the entire resync/webhook enrichment

integrations/gitlab-v2/gitlab/webhook/webhook_processors/project_with_member_webhook_processor.py
```diff
@@ -0,0 +1,114 @@
...
+        return True
+
+    async def get_matching_kinds(self, event: WebhookEvent) -> list[str]:
+        return [ObjectKind.PROJECT_WITH_MEMBERS]
+
+    async def handle_event(
+        self, payload: EventPayload, resource_config: ResourceConfig
+    ) -> WebhookEventRawResults:
+        # project_id may be absent for group-hook member events; fall back to path.
+        # validate_payload ensures at least one of these is present for non-destroy events.
```
raskinmaya: These lines say `validate_payload` guarantees a `ref` for “non-destroy” events, but `project_destroy` is not in `_MEMBER_EVENTS` and still goes through the `project_ref` path with `project_id` required elsewhere. Can we tighten the comment to match the real branches? (`_MEMBER_EVENTS` vs create/destroy)

integrations/gitlab-v2/gitlab/webhook/webhook_processors/project_with_member_webhook_processor.py
```diff
@@ -0,0 +1,114 @@
...
+            return WebhookEventRawResults(
+                updated_raw_results=[project],
+                deleted_raw_results=[],
+            )
+
+        logger.warning(f"Project '{project_ref}' not found")
+        return WebhookEventRawResults(
+            updated_raw_results=[],
+            deleted_raw_results=[],
+        )
```
raskinmaya, line 100: Returning empty lists here is the right contract. Worth aligning `group-with-members` the same way as it still upserts `[{}]` on missing group

raskinmaya (approved):

integrations/gitlab-v2/gitlab/helpers/utils.py
```diff
@@ -1,12 +1,37 @@
...
-import json
 
-# import strictyaml as syaml
-from yaml import safe_load, YAMLError
+from loguru import logger
+from yaml import YAMLError, safe_load
+
+# GitLab project/group access-token bots have system-generated usernames like
+# "project_<id>_bot_<hex>" or "group_<id>_bot_<hex>".  The members/all API
+# does NOT return a `bot` field, so we fall back to pattern detection.
```
mk-armah, line 13: For both cases, usernames are still generally generated with the bots, No ? can we solely depend on that if we decide not to add an extra logic to check the payload . ... ?
This route will save you an API request for webhook events that should be skipped

integrations/gitlab-v2/gitlab/webhook/webhook_processors/member_webhook_processor.py
```diff
@@ -71,6 +63,14 @@ async def handle_event(
...
             group_member = {}
 
+        if not selector.include_bot_members and is_bot_member(group_member):
+            logger.info(
+                f"Excluding bot member '{user_name}' from group '{group_id}' because include_bot_members is false"
+            )
+            return WebhookEventRawResults(
+                updated_raw_results=[], deleted_raw_results=[payload]
+            )
+
```
mk-armah, line 87: I understand the extra payload check is the reason this had to be moved down. But the initial approach strategically figured this out prior to making a request which saves some us some rate limits. Is the payload check very necessary for some specific scenarios ?

mk-armah (changes requested):

integrations/gitlab-v2/gitlab/helpers/utils.py
edenhermelin: I kept `is_bot_member` as-is because resync and `/members` responses often omit `bot` for project/group access-token users, so we still need the regex (and explicit bot when the API sends it). For **webhooks**, we now short-circuit **before** `get_group_member` using the same helper on `user_username` (and `user_bot` when present), so we might handle better the ratelimit where we can understand from the payload alone, without going down the "API-aware" path for enrichment and ambiguous cases.

WDYT?

integrations/gitlab-v2/gitlab/webhook/webhook_processors/member_webhook_processor.py
edenhermelin: Restored a **pre-`get_group_member`** check when `include_bot_members` is false: we call `is_bot_member` on `user_username` (and `user_bot` when GitLab sends it), skip the API if that’s already a bot, and only fetch when we still need a full member row or the webhook isn’t enough (e.g. `bot: true` only from the API, or usernames like `"botner-elay"` where we must trust `bot: false` from the member response).

mk-armah (approved):
LGTM

PR #2954, opened 2026-03-17
[Integration][Azure] Fix Literal kind definitions and enable custom kinds for schema validation compliance

eriport:
# Description

What -

Added explicit kind definitions for known Azure resource kinds, added dedicated resource config models per kind, enabled `allow_custom_kinds = True` on `AzurePortAppConfig` to support arbitrary Azure resource kinds, and added `title` and `description` fields to all selector fields.

Why -

Ocean framework requires explicit `Literal` kinds for schema extraction and the Mapping UI. Azure's `AzureSpecificKindsResourceConfig` needs `kind: str` to support arbitrary Azure resource kinds defined by users; so `allow_custom_kinds = True` is the correct opt-in approach here. The known special kinds, however needed their own dedicated models for schema generation not to fail.

How -

Split out `AzureResourceGroupResourceConfig` and `AzureSubscriptionResourceConfig` from the generic `AzureSpecificKindsResourceConfig`, set `allow_custom_kinds = True` on `AzurePortAppConfig`, updated the `get_resource_configs_with_resource_kind` signature to reflect the expanded union, and added missing field metadata to selectors.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


melodyogonna (changes requested):
Add kind examples for kinds with special handling

mk-armah (approved):
LGTM

melodyogonna (approved):
lgtm

integrations/azure/azure_integration/overrides.py
```diff
@@ -25,15 +31,57 @@ def validate_resource_kinds_min_size(cls, v: dict[str, str]) -> dict[str, str]:
 
 
 class AzureSpecificKindsResourceConfig(ResourceConfig):
```
omby8888:
```suggestion
class AzureCustomKindResourceConfig(ResourceConfig):
```

integrations/azure/azure_integration/overrides.py
```diff
@@ -25,15 +31,57 @@ def validate_resource_kinds_min_size(cls, v: dict[str, str]) -> dict[str, str]:
...
-    selector: AzureSpecificKindSelector
+    kind: str = Field(
+        title="Kind",
+        description="Azure resource kind.",
```
omby8888: this missing the description of azure custom kind we have in the list

integrations/azure/azure_integration/overrides.py
omby8888, line 31: this should be replaced with an attr that will reflect in the output json schema, i'll check and respond here later for that

omby8888: pls add the field extra attribute `minProperties=1` while keeping the root validator as pydantic seem to not validate `minProperties` for some reason while json schema supports it OOTB

omby8888 (approved):

PR #2973, opened 2026-03-18
[Integration][Datadog] Fix Literal kind definitions for schema validation compliance

eriport:
# Description

What -

Added `Literal` kind definitions to all existing resource config models, introduced `HostResourceConfig`, `MonitorResourceConfig`, `SLOResourceConfig`, `ServiceResourceConfig`, and `UserResourceConfig` for kinds that had resync handlers but no config model, and removed the bare `ResourceConfig` from the union.

Why -

Ocean framework requires explicit `Literal` kinds per resource config model for schema extraction and the Mapping UI. Datadog was missing this for several kinds, causing schema generation to fail.

How -

One model per kind.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/datadog/overrides.py
```diff
@@ -12,10 +12,30 @@
...
+    )
+    concurrency: int = Field(
+        alias="concurrency",
+        default=2,
+        title="Concurrency",
+        description="Number of concurrent requests to make to Datadog.",
+    )
 
     @validator("timeframe")
     def validate_timeframe_field(cls, v: int) -> int:
```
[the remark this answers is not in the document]
mk-armah: This is valid

mk-armah (changes requested):

integrations/datadog/overrides.py
```diff
@@ -12,10 +12,30 @@
...
+        default=7,
+        title="Timeframe",
+        description="Time window in days for each SLO history data point.",
+    )
+    period_of_time_in_months: int = Field(
+        alias="periodOfTimeInMonths",
+        default=6,
+        title="Period of Time in Months",
+        description="How far back in time to fetch SLO history (1-12 months).",
+    )
```
mk-armah, line 28: Why 6 ? The existing is 12

mk-armah: all the AI is right here

mk-armah (changes requested):

integrations/datadog/overrides.py
eriport: Hi @mk-armah, the original code had an inconsistency, the field defaulted to 12 but the validator's fallback was 6. I had to change the values to align; both to 6 to match the validator's intent, since the validator is the expected behaviour. Happy to go with 12 consistently if that's the preferred default.

integrations/datadog/CHANGELOG.md
```diff
@@ -7,11 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
+<<<<<<< task_tboe6s/datadog_selector_propagation
```
omby8888: fix merge econflicts

integrations/datadog/overrides.py
```diff
@@ -12,10 +12,30 @@
...
+        description="How far back in time to fetch SLO history in days (1-365).",
+    )
+    concurrency: int = Field(
+        alias="concurrency",
+        default=2,
+        title="Concurrency",
+        description="Number of concurrent requests to make to Datadog.",
+    )
 
     @validator("timeframe")
```
omby8888: pls replace all this validators with pydantic native ones (min/max)

integrations/datadog/overrides.py
mk-armah: change to 12

integrations/datadog/overrides.py
```diff
@@ -1,123 +1,195 @@
...
+    timeframe: int = Field(
+        alias="timeframe",
+        default=7,
+        title="Timeframe",
+        description="Time window in days for each SLO history data point.",
+        ge=1,
+    )
+    period_of_time_in_months: int = Field(
+        alias="periodOfTimeInMonths",
+        default=6,
```
mk-armah:
```suggestion
        default=12,
```

mk-armah (changes requested):
Left comments here

integrations/datadog/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
+## 0.3.144 (2026-05-20)
+
+
+### Improvements
+
+- Add missing field metadata for schema validation compliance
+- Reduced SLOHistorySelector `period_of_time_in_months` default from 12 to 6 for improved performance, and fixed `period_of_time_in_days` validator to safely handle None values
```
mk-armah: update changelog

mk-armah (changes requested):

mk-armah (approved):
LGTM

integrations/datadog/overrides.py
```diff
@@ -1,123 +1,195 @@
...
+        title="Period of Time in Days",
+        description="How far back in time to fetch SLO history in days (1-365).",
+        ge=1,
+        le=365,
+    )
+    concurrency: int = Field(
+        alias="concurrency",
+        default=2,
+        title="Concurrency",
+        description="Number of concurrent requests to make to Datadog.",
```
omby8888: pls specify the considerations of playing with this parameter, e.g why not making it 100

integrations/datadog/overrides.py
```diff
@@ -1,123 +1,195 @@
...
+        alias="env",
+        title="Environment",
+        description="Environment tag filter for the metric.",
+    )
+    service: DatadogMetricSelector = Field(
+        alias="service",
+        title="Service",
+        description="Service tag filter for the metric.",
+    )
     timeframe: int = Field(
```
omby8888, line 90: missing ge=1

integrations/datadog/overrides.py
```diff
@@ -1,123 +1,195 @@
...
 class DatadogResourceSelector(Selector):
-    datadog_selector: DatadogSelector = Field(alias="datadogSelector")
+    datadog_selector: DatadogSelector = Field(
+        alias="datadogSelector",
+        title="Datadog Selector",
+        description="Datadog metric query configuration.",
+    )
 
 
 class DatadogResourceConfig(ResourceConfig):
```
omby8888: `DatadogSelector`, `DatadogResourceSelector`, `DatadogResourceConfig` are super generic names, if im not mistaken all are being used solely in serviceMetric kind no? if so lets make those classes names specific

integrations/datadog/overrides.py
eriport: `SLOHistorySelector.concurrency` controls how this resource is collected asynchronously. Internally, it acts as a bound parameter that defines how many `get_slos` requests can run concurrently when fetching `slo_histories`. this is handled through the [`process_in_queue`](https://github.com/port-labs/ocean/blob/main/port_ocean/utils/queue_utils.py#L30) utility in ocean core.

from the implementation, the default concurrency is set to 50 to help avoid rate limiting. I think we should align this value here with the underlying default for consistency. I am open to what you think about this value

integrations/datadog/overrides.py
```diff
@@ -1,137 +1,218 @@
...
+        title="Timeframe",
+        description="Time frame in minutes for the metric query.",
+        ge=1,
     )
 
 
-class DatadogResourceSelector(Selector):
-    datadog_selector: DatadogSelector = Field(alias="datadogSelector")
+class ServiceMetricResourceSelector(Selector):
+    metric_selector: ServiceMetricSelector = Field(
```
omby8888, line 100: unrelated to your PR as i know it's currently existing behavior - this is really weird that this kind has a generic selector that s an object of the actually relevant selectors, can u pls open a task in Port to deprecate this one and instead spread the selectors on the root level?
i.e before:
```
- kind: serviceMetric
   selector:
     query: 'true'
     datadogSelector:
       metric: 'x'
       env:
         tag: 'y'
         value: '*'
       service:
         tag: 'y'
         value: '*'
       timeframe: 2
```

after:
```
- kind: serviceMetric
   selector:
     query: 'true'
     metric: 'x'
     env:
       tag: 'y'
       value: '*'
     service:
       tag: 'y'
       value: '*'
     timeframe: 2
```

omby8888 (approved):

PR #3063, opened 2026-04-08
[Integration][Azure DevOps] Enhance repository-policy and test-run fetching with concurrency support

dennis-quabynah-bilson:
# Description

This update optimizes the Azure DevOps client for better efficiency in handling multiple projects simultaneously.

## Type of change

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

### BEFORE
image

### AFTER
image

## API Documentation


integrations/azure-devops/azure_devops/client/azure_devops_client.py
```diff
@@ -447,25 +449,47 @@ async def generate_group_members(
                 )
                 yield members
 
+    async def _generate_repositories_for_project(
```
danamsalem-ux: I feel this is a big change and it might affect customers with heavy loads, which will cause their resyncs to take way longer then it already is.
The original issue were with repository-policy and test-run kinds, which were with smaller loads and were with a little latency then expected. Lets start with a change to this 2 resources and see how it functions. 
If we want to check heavier resources like work-items, i suggest we should check it as part of the stress tests Melody is currently doing for TR, and see if it affects it with heavy loads

danamsalem-ux (approved):

mk-armah (approved):
LGTM, please update the dates in the changelog

PR #3073, opened 2026-04-09
[Integration][Snyk] Allow project filtering for vulnerability kind and project kind

melodyogonna:
# Description

What - Allow vulnerabilities and projects to be filtered using API query params, this provides more fine-grained control of the ingested data. Also added missing vulnerability webhook processor.

Why - Allow fine-grained control of ingested data.

How - Add query params class to handle API query params

## Type of change

- [ x ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Core testing checklist

- [ x ] Integration able to create all default resources from scratch
- [ x ] Resync finishes successfully

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/snyk/snyk/client.py
```diff
@@ -146,16 +148,60 @@ async def get_issues(self, org_id: str, project_id: str) -> list[dict[str, Any]]
...
+                ]
+
+                async for issues in stream_async_iterators_tasks(*tasks):
+                    yield enrich_batch_with_data(issues, options.org)
+        else:
+            url = f"/orgs/{options.org['id']}/issues"
+            async for issues in self._get_paginated_resources(
+                url, query_params=query_params
+            ):
+                yield enrich_batch_with_data(issues, options.org)
```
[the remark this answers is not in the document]
mk-armah: @melodyogonna

mk-armah (approved):
approved with a minor comment

integrations/snyk/snyk/client.py
```diff
@@ -146,16 +148,60 @@ async def get_issues(self, org_id: str, project_id: str) -> list[dict[str, Any]]
...
+            url,
+            query_params={
+                **query_params,
+                "scan_item.id": project["id"],
+                "scan_item.type": project["type"],
+                "version": self.snyk_api_version,
+            },
         ):
-            yield enrich_batch_with_org(issues, org)
+            yield enrich_batch_with_data(issues, project, enricment_key="__project")
```
raskinmaya: `enricment_key` is a typo. should be `enrichment_key`. Same typo in `utils.py` line 7. Fix the parameter name at the source before it proliferates

integrations/snyk/snyk/client.py
```diff
@@ -174,20 +220,30 @@ def _get_projects_by_target(
...
 
         async for projects in self._get_paginated_resources(
-            url_path=url, query_params=query_params
+            url_path=url, query_params=final_query_params
         ):
-            yield enrich_batch_with_org(projects, org)
+            if options.enrich_with_org:
+                yield enrich_batch_with_data(projects, options.org)
+            else:
+                yield projects
```
raskinmaya: The cache key is derived from `repr(options)`. The `enrich_with_org` flag is part of the options, so: if `get_paginated_issues` calls this first with `enrich_with_org=False` (line 187), the unenriched result gets cached. Then `on_projects_resync` calls it for the same org with `enrich_with_org=True` - it hits the cache and gets data without `__organization`. The inverse is equally broken. Enrichment logic must live outside the cached function. Return raw data from the cache, enrich at the call site

integrations/snyk/snyk/client.py
```diff
@@ -146,16 +148,60 @@ async def get_issues(self, org_id: str, project_id: str) -> list[dict[str, Any]]
...
+        url = f"/orgs/{org_id}/issues"
         async for issues in self._get_paginated_resources(
-            url_path=f"/orgs/{org['id']}/issues", query_params=query_params
+            url,
+            query_params={
+                **query_params,
+                "scan_item.id": project["id"],
+                "scan_item.type": project["type"],
+                "version": self.snyk_api_version,
+            },
```
raskinmaya, line 161: `get_paginated_issues` already puts `version` into the `query_params` dict it passes here (line 170). Then this method slaps another version on top. It works by accident (dict merge order), but it's fragile and confusing. Set version in exactly one place

integrations/snyk/snyk/client.py
```diff
@@ -146,16 +148,60 @@ async def get_issues(self, org_id: str, project_id: str) -> list[dict[str, Any]]
...
+            ):
+                tasks = [
+                    self.get_project_vulnerabilities(
+                        options.org["id"], project, query_params
+                    )
+                    for project in projects
+                ]
+
+                async for issues in stream_async_iterators_tasks(*tasks):
+                    yield enrich_batch_with_data(issues, options.org)
```
raskinmaya: When `attach_project=True` and `project_params=None`, this triggers an N+1 pattern - one issues call per project instead of a single org-level call. For large orgs, this can be very expensive. Consider logging a warning or documenting that `enrichWithProject: true` is `O(projects * issues)`.

Also, when both `attach_project=True` and `project_params` are set, you're effectively reimplementing server-side filtering. The Snyk org issues endpoint already supports filtering by `scan_item.id`, so fetching projects first can lead to hundreds of unnecessary API calls and potential rate limiting

integrations/snyk/snyk/options.py
raskinmaya, line 1: I think the entire file is unnecessary. 
These are parameter objects passed to exactly one method each. The codebase convention (see `get_paginated_targets`, `get_single_target_by_project_id`, `process_project_issues`) is to pass arguments directly. Creating a new file and Pydantic models just to bundle arguments is over-engineering and breaks the established pattern. Use keyword arguments on the methods directly, like every other method in this client does

integrations/snyk/snyk/overrides.py
```diff
@@ -4,18 +4,184 @@
...
 
 
+class GenerateQueryParamMixin(BaseModel):
+    def generate_query_params(self) -> dict[str, Any]:
+        params = self.dict(exclude_none=True, exclude_unset=True)
+        return {k: ",".join(v) if isinstance(v, list) else v for k, v in params.items()}
+
+    def merge_with(self, other: dict[str, Any]) -> dict[str, Any]:
+        query_params = self.generate_query_params()
+        return {**query_params, **other}
```
raskinmaya: `merge_with` lets other override the model's own values. This means system-critical params like `version` silently stomp user-provided params. The name `merge_with` doesn't convey that other wins. Either rename to something explicit like `with_defaults_overriding`, or flip the merge order so user params win (and protect system params separately)

integrations/snyk/snyk/overrides.py
```diff
@@ -4,18 +4,184 @@
...
+        default=None,
+        title="Lifecycle",
+        description="Return projects with the provided lifecycle",
+    )
+    ids: Optional[list[str]] = Field(
+        default=None,
+        title="Project IDs",
+        description="Return only projects with the provided IDs",
+    )
+
```
raskinmaya, line 105: How were these tested?
Doesn't the Snyk API expect `target.id`? Same for `names_start_with` - doesn't Snyk expect `names[start_with]` or similar bracket notation? Let's verify each of these against the Snyk docs and use `Field(alias=...)` to map Python field names to the correct API parameter names. If these don't match the Snyk API exactly, filtering will silently do nothing (AFAIK Snyk ignores unknown params)

integrations/snyk/webhook_processors/vulnerability_webhook_processor.py
```diff
@@ -0,0 +1,54 @@
...
+        ):
+            issues.extend(issue_batch)
+        if organization:
+            item_to_upsert = enrich_batch_with_data(issues, organization)
+        else:
+            item_to_upsert = issues
+
+        return WebhookEventRawResults(
+            updated_raw_results=item_to_upsert,
+            deleted_raw_results=payload.get("removedIssues", []),
```
raskinmaya: The comment on line 22 says "no deletion in snyk events" (which is consistent with the Snyk webhook docs and what all other processors do - they all pass `deleted_raw_results=[]`). But then you pass `payload.get("removedIssues", [])` as deleted results. This is contradictory. Either the comment is wrong, or the code is. Looking at Snyk's webhook payload docs, `removedIssues` uses v1 issue IDs which are not the same as the REST API issue IDs you're ingesting - so even if the field exists, the IDs won't match any entities and deletions will silently fail or delete the wrong things. This should be `deleted_raw_results=[]` to match the other processors, or you need an explicit ID translation layer

integrations/snyk/webhook_processors/vulnerability_webhook_processor.py
```diff
@@ -0,0 +1,54 @@
...
+            await snyk_client.get_all_organizations(), organization_id
+        )
+        project_details = await snyk_client.get_single_project(
+            organization_id, project_id
+        )
+        issues = []
+        async for issue_batch in snyk_client.get_project_vulnerabilities(
+            organization_id, project_details, query_params
+        ):
+            issues.extend(issue_batch)
```
raskinmaya, line 45: Every other webhook processor returns data directly without accumulating into a list. For a project with thousands of vulnerabilities, this loads everything into memory. The other processors (`project`, `issue`) fetch a single resource or a bounded list. This one paginates through potentially unbounded results. Consider yielding batches or at least adding a size guard

integrations/snyk/webhook_processors/vulnerability_webhook_processor.py
raskinmaya, line 1: The resync path in `main.py` supports `selector.enrich_with_project` which attaches `__project` data to each vulnerability. The webhook processor completely ignores this flag. A user who configured `enrichWithProject: true` will get project-enriched data from resyncs but not from webhooks, causing inconsistent entity stat

integrations/snyk/tests/test_processors.py
```diff
@@ -382,3 +415,121 @@ async def test_target_webhook_processor_handle_event_should_pass_attach_project_
...
+        mock_client.get_project_vulnerabilities = mock_get_project_vulnerabilities
+
+        result = await vulnerabilityWebhookProcessor.handle_event(
+            payload, vulnerability_resource_config
+        )
+
+        assert len(result.updated_raw_results) == 1
+        assert len(result.deleted_raw_results) == 0
+        assert result.updated_raw_results[0]["id"] == "vuln-1"
+        assert result.updated_raw_results[0]["__organization"] == mock_org
```
raskinmaya, line 503: Missing test coverage for a few edge cases:
1. `enrichWithProject: true` on the webhook path
2. `empty project_details` (e.g. project deleted → Snyk 404)
3. projects without a type field in the payload
4. the `removedIssues` deletion flow

integrations/snyk/tests/snyk/test_overrides.py
raskinmaya, line 1: The tests verify `generate_query_params` joins lists, but no test verifies that the generated keys actually match Snyk's expected API parameter names (e.g. `target_id` vs `target.id`). If there's a mismatch, all filtering silently does nothing

integrations/snyk/snyk/client.py
```diff
@@ -146,16 +148,60 @@ async def get_issues(self, org_id: str, project_id: str) -> list[dict[str, Any]]
...
+            yield enrich_batch_with_data(issues, project, enricment_key="__project")
+
+    async def get_paginated_issues(
+        self, options: IssueOptions
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        logger.info(f"Fetching paginated issues for organization: {options.org['id']}")
+        base_params = {"version": self.snyk_api_version}
+        query_params = (
+            options.api_params.merge_with(base_params)
+            if options.api_params
```
raskinmaya: Here you check `if options.api_params` (truthy), but on line 236 you check if `options.api_params is not None` (identity). Pick one style. Since an empty `SnykProjectAPIQueryParams()` is truthy but semantically means "no filters", `is not None` is more correct everywhere

integrations/snyk/webhook_processors/vulnerability_webhook_processor.py
```diff
@@ -0,0 +1,54 @@
...
+
+class VulnerabilityWebhookProcessor(SnykBaseWebhookProcessor):
+    async def get_matching_kinds(self, event: WebhookEvent) -> list[str]:
+        return [IntegrationKind.VULNERABILITY]
+
+    async def handle_event(
+        self, payload: EventPayload, resource_config: ResourceConfig
+    ) -> WebhookEventRawResults:
+        # no deletion in snyk events https://docs.snyk.io/snyk-api/how-to-use-snyk-webhooks-apis/webhooks
+        selector = cast(VulnerabilityResourceConfig, resource_config).selector
```
raskinmaya, line 23: The cast is consistent with other processors (`target_webhook_processor` does the same), but note that `get_matching_kinds` returns `["vulnerability"]` unconditionally, meaning this processor fires for every webhook event that passes `should_process_event`. If a non-vulnerability `resource_config` is passed, this cast will silently produce wrong behavior. Not a new problem (same in target processor), but worth noting

integrations/snyk/main.py
```diff
@@ -74,7 +82,8 @@ async def on_issues_resync(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
 
     all_organizations = await snyk_client.get_organizations_in_groups()
     project_tasks = (
-        snyk_client.get_paginated_projects(org) for org in all_organizations
+        snyk_client.get_paginated_projects(ProjectOptions(org=org))
```
raskinmaya: The issue resync path creates `ProjectOptions(org=org)` without forwarding any project filter params. This is intentional (issues are deprecated), but if a user has project filters configured and expects them to also limit the issues kind, they won't. Worth a comment

raskinmaya (changes requested):

integrations/snyk/webhook_processors/vulnerability_webhook_processor.py
melodyogonna: ha, good catch. Added that after the fact, now seeing it was thought out properly. Snyk doesn't really send deletion events

integrations/snyk/webhook_processors/vulnerability_webhook_processor.py
melodyogonna: Vulnerabilities are paginated, accumulation is a limitation of the webhook processor, I don't think we can yield incrementally.

integrations/snyk/webhook_processors/vulnerability_webhook_processor.py
melodyogonna: Vulnerabilities webhook processor uses `get_project_vulnerabilities` which always enrich the result.

integrations/snyk/tests/snyk/test_overrides.py
melodyogonna: the keys match exactly what is expected by the Snyk api. And it is `target_id` not `target.id`, ref: https://docs.snyk.io/snyk-api/reference/projects

integrations/snyk/webhook_processors/vulnerability_webhook_processor.py
melodyogonna: I'm not sure I understand. get_matching_kinds will ensure this processor is always called with resource config for vulnerability kind.

integrations/snyk/main.py
melodyogonna: I generally try to avoid adding new behaviour to the deprecated issue kind. It is here as a legacy kind for customers who were using it before the new vulnerability kind, for those customers, not filtering projects is the expected and only behaviour.

integrations/snyk/snyk/overrides.py
melodyogonna: Nah, all the keys here are exactly the same as the keys expected by Snyk API. Surprising to me too but I won't complain, made things a lot easier for me.

integrations/snyk/snyk/client.py
melodyogonna: It is expensive API wise, but note that it is much faster. In my testing it improves performance by as much as 50%.

raskinmaya: The speed win is fair. My concern is the failure mode: with `enrichWithProject: true` and no `project_params`, an org with 500 projects fires hundreds of concurrent calls. If Snyk starts 429-ing, the retry storm makes it slower than the single endpoint

integrations/snyk/tests/test_processors.py
raskinmaya: Was this handled?

integrations/snyk/snyk/client.py
melodyogonna: There is a semaphore on the client.

raskinmaya (approved):

PR #3112, opened 2026-04-19
[Integration][Jira] Added support for Board Kind

eriport:
# Description

What - Adds `board` as a new kind to the Jira integration, sourced from the Jira Software Cloud Agile REST API (`/rest/agile/1.0/board`).

Why - Part of the Jira Agile kinds feature request to support Boards, Sprints, Epics, and Backlog in Port's Jira integration.

How - 

- Added `_get_agile_api_url()` with lazy cloud ID resolution for OAuth - resolves once via `accessible-resources` endpoint and caches on the instance. Basic auth resolves immediately at construction time.
- Added `_get_agile_paginated_data()` - a dedicated paginator for Agile REST API endpoints that terminates on `isLast: true` rather than comparing `startAt` against `total`, keeping agile pagination logic separate from the existing `/rest/api/3` paginator.
- Added `JiraBoardSelector` with `boardType` (`scrum` | `kanban` | `simple`) and `projectKey` optional filters.
- Added `JiraBoardResourceConfig` and registered it in `JiraPortAppConfig`.
- Added `on_resync_boards` handler and `BoardWebhookProcessor` handling `board_created`, `board_updated`, `board_deleted` events.
- Added `jiraBoard` blueprint with relation to `jiraProject`.
- Full test coverage: selector parsing, paginator edge cases, cloud ID resolution, webhook processor behavior.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

image

## API Documentation

[Board - Jira Software Cloud API Reference](https://developer.atlassian.com/cloud/jira/software/rest/api-group-board/#api-group-board)


integrations/jira/.port/resources/blueprints.json
```diff
@@ -220,5 +217,80 @@
...
+    "identifier": "jiraBoard",
+    "title": "Jira Board",
+    "icon": "Jira",
+    "schema": {
+      "properties": {
+        "url": {
+          "title": "Board API URL",
+          "type": "string",
+          "format": "url",
+          "description": "REST API URL to the board in Jira"
```
Hadar-Vaknin: Do we really need the url in the blueprint?

integrations/jira/.port/resources/blueprints.json
```diff
@@ -220,5 +217,80 @@
...
+        },
+        "type": {
+          "title": "Board Type",
+          "type": "string",
+          "description": "The type of the board",
+          "enum": ["scrum", "kanban", "simple"],
+          "enumColors": {
+            "scrum": "blue",
+            "kanban": "green",
+            "simple": "lightGray"
```
Hadar-Vaknin: I don't see reason for the colors?

integrations/jira/.port/resources/blueprints.json
```diff
@@ -220,5 +217,80 @@
...
+        },
+        "projectKey": {
+          "title": "Project Key",
+          "type": "string",
+          "description": "The key of the project this board belongs to"
+        },
+        "projectName": {
+          "title": "Project Name",
+          "type": "string",
+          "description": "The name of the project this board belongs to"
```
Hadar-Vaknin: In Jira, a board is usually connected to one or more projects
So I think a board can include multiple projects

integrations/jira/.port/resources/blueprints.json
```diff
@@ -220,5 +217,80 @@
...
+    "mirrorProperties": {},
+    "calculationProperties": {},
+    "aggregationProperties": {},
+    "relations": {
+      "project": {
+        "title": "Project",
+        "description": "The Jira project this board belongs to",
+        "target": "jiraProject",
+        "required": false,
+        "many": false
```
Hadar-Vaknin: Related to my comments, I think it's `many: true`

integrations/jira/jira/client.py
```diff
@@ -104,6 +117,56 @@ def _get_bearer(self) -> BearerAuth:
...
+        For basic auth this is known at construction time.
+        For OAuth, the cloud ID must be resolved once via the accessible-resources
+        endpoint and the result is cached on the instance.
+        """
+        if self._agile_api_url is not None:
+            return self._agile_api_url
+
+        cloud_id = await self._get_cloud_id()
+        self._agile_api_url = (
+            f"https://api.atlassian.com/ex/jira/{cloud_id}/rest/agile/1.0"
```
Hadar-Vaknin: From the Notion docs u specified we can set the latest instead the 1.0 for resolving the most recent deployed api version of the collection, don't we prefer latest then?

integrations/jira/jira/client.py
```diff
@@ -104,6 +117,56 @@ def _get_bearer(self) -> BearerAuth:
...
+        """Return the Agile REST API base URL for the current auth scheme.
+
+        For basic auth this is known at construction time.
+        For OAuth, the cloud ID must be resolved once via the accessible-resources
+        endpoint and the result is cached on the instance.
+        """
+        if self._agile_api_url is not None:
+            return self._agile_api_url
+
+        cloud_id = await self._get_cloud_id()
```
Hadar-Vaknin, line 130: I think we need the check if the integration is running as OAuth before - recording to the docstring.
Because if we have `_agile_api_url` is already set, the code will not take the cloud id

integrations/jira/jira/client.py
```diff
@@ -104,6 +117,56 @@ def _get_bearer(self) -> BearerAuth:
             )
             return BearerAuth(self.jira_token)
 
+    async def _get_agile_api_url(self) -> str:
+        """Return the Agile REST API base URL for the current auth scheme.
+
+        For basic auth this is known at construction time.
+        For OAuth, the cloud ID must be resolved once via the accessible-resources
```
Hadar-Vaknin, line 124: According to the code in `_get_cloud_id`  the cloud id can be extracted from the code, and the docstring says it resolved via the accessible-resources.
Let's change it to be less vague.

integrations/jira/jira/client.py
```diff
@@ -104,6 +117,56 @@ def _get_bearer(self) -> BearerAuth:
...
+            return cloud_id
+
+        resources = await self._send_api_request(
+            "GET",
+            "https://api.atlassian.com/oauth/token/accessible-resources",
+        )
+        normalized_jira_url = self.jira_url.rstrip("/")
+        for resource in resources:
+            if resource.get("url", "").rstrip("/") == normalized_jira_url:
+                resolved_id: str = resource["id"]
```
Hadar-Vaknin: `resolved_id: str = resource["id"] ` will raise KeyError if a resource entry is malformed
use .get instead.

integrations/jira/jira/client.py
```diff
@@ -104,6 +117,56 @@ def _get_bearer(self) -> BearerAuth:
...
+                logger.debug(f"Resolved cloud ID {resolved_id} for {self.jira_url}")
+                return resolved_id
+
+        raise ValueError(
+            f"Could not resolve cloud ID for Jira site '{self.jira_url}'. "
+            f"Ensure the configured jiraHost matches one of the accessible sites "
```
Hadar-Vaknin:
```suggestion
            f"Ensure the configured jira url matches one of the accessible sites "
```

integrations/jira/jira/client.py
```diff
@@ -177,6 +240,45 @@ async def _get_paginated_data(
...
                 break
 
+    async def _get_agile_paginated_data(
+        self,
+        url: str,
+        extract_key: str = "values",
+        initial_params: dict[str, Any] | None = None,
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        """Paginate Jira Agile REST API endpoints using offset-based pagination."""
+        params: dict[str, Any] = initial_params or {}
```
Hadar-Vaknin: You mutating the `initial_params` here.
If u rely on the initial_params somewhere else in the code that can be problematic
avoid mutating parameters unless there’s a clear, intentional reason.

integrations/jira/jira/overrides.py
```diff
@@ -90,13 +90,49 @@ class JiraReleaseResourceConfig(ResourceConfig):
...
 
 
+class JiraBoardSelector(Selector):
+    board_type: Literal["scrum", "kanban", "simple"] | None = Field(
+        alias="boardType",
+        default=None,
+        title="Board Type",
+        description=("Filter boards by type. Omit to fetch all board types."),
+    )
+    project_key: str | None = Field(
```
Hadar-Vaknin, line 100: Same here for the project_key about the many projects

integrations/jira/tests/webhook_processors/__init__.py
Hadar-Vaknin, line 1: There are some whitespace changes.

integrations/jira/.port/resources/blueprints.json
mk-armah, line 1: remove from default blueprints

integrations/jira/.port/resources/port-app-config.yaml
mk-armah, line 1: remove from default mapping

integrations/jira/jira/client.py
```diff
@@ -104,6 +117,56 @@ def _get_bearer(self) -> BearerAuth:
...
+        """
+        if self._agile_api_url is not None:
+            return self._agile_api_url
+
+        cloud_id = await self._get_cloud_id()
+        self._agile_api_url = (
+            f"https://api.atlassian.com/ex/jira/{cloud_id}/rest/agile/1.0"
+        )
+        logger.debug(f"Resolved agile API URL: {self._agile_api_url}")
+        return self._agile_api_url
```
mk-armah, line 135: Is cloud_id dynamic or does is it the same regardless of how many times we call it ?
If its static, i want this function to only evaluate once, one the url is obtained the first time, cache

integrations/jira/jira/client.py
```diff
@@ -177,6 +240,45 @@ async def _get_paginated_data(
...
+        initial_params: dict[str, Any] | None = None,
+    ) -> AsyncGenerator[list[dict[str, Any]], None]:
+        """Paginate Jira Agile REST API endpoints using offset-based pagination."""
+        params: dict[str, Any] = initial_params or {}
+        params.setdefault("maxResults", PAGE_SIZE)
+        start_at = 0
+
+        while True:
+            params["startAt"] = start_at
+            response_data = await self._send_api_request("GET", url, params=params)
```
mk-armah, line 260: why is this not using on of our pagination functions ?

integrations/jira/webhook_processors/board_webhook_processor.py
```diff
@@ -0,0 +1,65 @@
...
+        return [Kinds.BOARD]
+
+    async def authenticate(self, payload: EventPayload, headers: EventHeaders) -> bool:
+        return True
+
+    async def validate_payload(self, payload: EventPayload) -> bool:
+        board = payload.get("board")
+        if not isinstance(board, dict):
+            logger.warning("Invalid payload: missing board information")
+            return False
```
mk-armah, line 31: is it likely to receive any other type apart from a dict ?

integrations/jira/jira/overrides.py
```diff
@@ -90,13 +90,49 @@ class JiraReleaseResourceConfig(ResourceConfig):
     )
 
 
+class JiraBoardSelector(Selector):
+    board_type: Literal["scrum", "kanban", "simple"] | None = Field(
+        alias="boardType",
+        default=None,
+        title="Board Type",
```
mk-armah, line 97: are we sure these are the only possible values there is ?

mk-armah (changes requested):
left some comments

integrations/jira/jira/overrides.py
eriport: yes please, according to their docs:

> type
object

> Filters results to boards of the specified types. Valid values: scrum, kanban, simple.

integrations/jira/webhook_processors/board_webhook_processor.py
eriport: no "single board" object is what this webhook listens to.

integrations/jira/jira/client.py
eriport: the consensus here (within my pod), was its better to maintain a new clean and maintainable pagination function since they fundamentally different, in-spite of the fact that they are both cursor navigation under the hood. Adding this functionality within existing pagination function would meaning knowing "when to paginate agile resources", whereas with this approach, its clean and single responsibility.

integrations/jira/jira/client.py
eriport: its the same regardless, its tied to an Atlassian organization and yes please, it's cached _agile_api_url is set on first resolution and returned directly on all subsequent calls without re-evaluating. _get_cloud_id is only ever called once per client instance.

integrations/jira/jira/overrides.py
eriport: `project_key` here in the selector is a filter parameter, that scopes the board list query to boards relevant to a specific project, it maps to `projectKeyOrId` on `GET /rest/agile/1.0/board`. It's not a representation of how many projects a board belongs to. For the multi-project relation on the board entity itself, I've added full support via `GET /rest/agile/1.0/board/{boardId}/project` and the project relation is now `many: true` with all associated project keys injected at resync time.

From the docs:
> projectKeyOrId
string
> Filters results to boards that are relevant to a project. 
Relevance means that the jql filter defined in board contains 
a reference to a project.

integrations/jira/jira/client.py
```diff
@@ -509,3 +611,59 @@ async def get_single_version(self, version_id: str) -> dict[str, Any]:
...
+    async def enrich_board_with_projects(self, board: dict[str, Any]) -> dict[str, Any]:
+        """Enrich a board with all associated project keys.
+
+        Fetches all projects for the board and injects __projectKeys
+        as a list for relation mapping.
+
+        Args:
+            board: The raw board object from the list endpoint.
+        """
+        board_id: int = board["id"]
```
Hadar-Vaknin: same here - if the id is not set it will raise an error

integrations/jira/tests/test_overrides.py
```diff
@@ -0,0 +1,170 @@
+import pytest
```
Hadar-Vaknin, line 1: There are no tests for projectKey: "" or whitespace-only values. Adding two small tests (or one parametrized case) would lock that contract and document the error message.

integrations/jira/webhook_processors/board_webhook_processor.py
```diff
@@ -0,0 +1,65 @@
...
+
+        client = create_jira_client()
+        logger.debug(f"Fetching board with id: {board_id}")
+        item = await client.get_single_board(board_id)
+
+        if not item:
+            logger.warning(
+                f"Board {board_id} could not be retrieved after {webhook_event} event"
+            )
+            return WebhookEventRawResults(
```
Hadar-Vaknin, line 56: The board mapping expects .__projectKeys, but the webhook path returns the raw board from get_single_board() without enrichment. For board_created / board_updated, this can produce board upserts with project: null and potentially drop/override previously synced relations.

Hadar-Vaknin (approved):

integrations/jira/.port/resources/blueprints.json
mk-armah: revert changes made here

integrations/jira/.port/resources/port-app-config.yaml
mk-armah: revert this change

mk-armah (changes requested):

integrations/jira/jira/client.py
mk-armah: No its not called once per client i see its called in `_get_agile_api_url` which means its called on before every sync function

mk-armah (changes requested):

mk-armah (approved):
LGTM

mk-armah:
update the date in the changelog and ping me

PR #3118, opened 2026-04-19
[Integration][GitHub] Add includeSAMLEmail Selector for SAML Email Enrichment

emekanwaoma:
# Description

What -
Add an `includeSAMLEmail` selector for GitHub Ocean user/team exports that enriches exported members with a stable `__SAMLEmail` field sourced from the org SAML identity provider.

Why -
Enterprise orgs see inconsistent GitHub `email` values (often personal/public). `__SAMLEmail` provides a reliable corporate identity signal when SAML external identities exist, without depending on GitHub profile email behavior.

How -
- Extend GitHub Ocean selectors/options to propagate `includeSAMLEmail` through resync and webhook paths.
- Enrich exported users and team members with `__SAMLEmail` when enabled.
- Keep existing SAML behavior of filling `email` only when missing.
- Update/expand tests and changelog.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)


integrations/github/github/helpers/utils.py
```diff
@@ -274,25 +274,35 @@ async def enrich_members_with_saml_email(
...
-            member["email"] = saml_users[member["login"]]
+    for member in members:
+        login = member.get("login")
+        saml_email = saml_map.get(login) if login else None
+
+        if include_saml_email:
+            member["__SAMLEmail"] = saml_email
+
+        if not member.get("email") and saml_email:
+            member["email"] = saml_email
```
omby8888:
```suggestion
        should_fallback_to_saml_email = not member.get("email") and saml_email
        if include_saml_email or should_fallback_to_saml_email:
            member["__SAMLEmail"] = saml_email

        # This is deprecated attribute when using SAML email event as fallback
        if should_fallback_to_saml_email:
            member["email"] = saml_email
```
wdyt? I think we'll want to move customers who uses the email key for SAML email as those are 2 different things, so in any case we need the SAML email we'll populate it into the new key name

omby8888 (approved):
Left one suggestion
Very nice!!!

mk-armah (approved):
LGTM

PR #3123, opened 2026-04-20
[Integration] Cloud Function / HTTP Sync Protocol

thomasrichner-oviva:
# Description

What - A simple connector to sync data over HTTP.

Why - The generic HTTP integration is often not powerful enough while a full custom integration is too heavyweight.
A great use-case is plugging in an existing data-warehouse or writing an adapter for a custom API.

How - Simple but well-defined HTTP protocol specifically for syncing data. Inspired by existing similar protocols.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


cloud_function/client.py
```diff
@@ -0,0 +1,86 @@
...
+            {"id":101, "name": "Christmas"},
+            {"id":102, "name": "New Year"}
+        ],
+    "hasMore" : true
+}
+```
+"""
+
+
+class CloudFunctionClient:
```
thomasrichner-oviva, line 35: This is the core of the entire integration.

omby8888:
After reviewing the approach, I have a few concerns that lead me to think we shouldn't merge it in its current form:

1. **Duplicate source of truth** - This creates a second source for cloud provider data, which already exists in our dedicated cloud integrations. Maintaining two parallel paths for the same data tends to create inconsistencies over time.
2. **Security & secrets handling** - Cloud provider integrations require careful handling of credentials and secrets. We've already built that logic into the existing integrations, and replicating it here would introduce surface area for security issues.

**Our suggestion:** Rather than a standalone implementation, we think the right approach is to add this as a new *kind* within the existing cloud integrations. That way it benefits from the existing secrets management, avoids duplication, and stays consistent with how we model cloud resources across the board.

That said, we really appreciate the initiative - the use case itself is valid and worth tracking! We'd love it if you opened a **Feature Request** so we can prioritize it on our roadmap and keep you in the loop as it moves forward.

Happy to discuss further - let us know what you think!

cloud_function/.port/spec.yaml
```diff
@@ -0,0 +1,15 @@
+description: Cloud Function integration for Port Ocean
+icon: Cookiecutter
```
matan84, line 2: I think we can use the existing GCP icon here, WDYT?

cloud_function/.port/spec.yaml
```diff
@@ -0,0 +1,15 @@
+description: Cloud Function integration for Port Ocean
```
matan84, line 1:
```suggestion
description: Google Cloud (GCP) Cloud Function integration
```

cloud_function/pyproject.toml
```diff
@@ -0,0 +1,114 @@
+[tool.poetry]
+name = "cloud_function"
+version = "1.0.0"
+description = "Sync entities from upstream APIs via Cloud Functions into Port"
+authors = ["Thomas Richner <thomas.richner@oviva.com>"]
+
+[tool.poetry.dependencies]
+python = "^3.12"
+port_ocean = { version = "^0.33.0", extras = ["cli"] }
```
matan84, line 9: Please update to the latest Ocean version

cloud_function/.port/resources/.gitignore
matan84, line 1: Please place all this integration's files under integrations/cloud_function

cloud_function/CONTRIBUTING.md
matan84, line 1: Can you please provide instructions on how can we also debug this integration's behavior? What resources do we need to create etc..?

cloud_function/main.py
```diff
@@ -0,0 +1,34 @@
...
+        yield data
+
+
+# Optional
+# Listen to the start event of the integration. Called once when the integration starts.
+@ocean.on_start()
+async def on_start() -> None:
+    # Something to do when the integration starts
+    # For example create a client to query 3rd party services - GitHub, Jira, etc...
+    print("Starting cloud_function integration")
```
matan84, line 34: Please delete this if possible

cloud_function/main.py
```diff
@@ -0,0 +1,34 @@
...
+from port_ocean.utils import http_async_client
+
+
+# Required
+# Listen to the resync event of all the kinds specified in the mapping inside port.
+# Called each time with a different kind that should be returned from the source system.
+@ocean.on_resync()
+async def on_resync(kind: str) -> AsyncGenerator[list[dict[str, Any]], None]:
+    port_client = ocean.integration.context.app.port_client
+    agent = f"{port_client.integration_identifier}/{port_client.integration_version}"
```
matan84, line 14: We bump integration versions for each change in Ocean core, Can we maybe use the integration id alone?

cloud_function/client.py
```diff
@@ -0,0 +1,86 @@
...
+                "agent": self.agent,
+                "secrets": self.secrets,
+                "kind": kind,
+                "state": state,
+            }
+            response = await self.http_client.request(
+                method="POST",
+                url=self.function_url,
+                json=json_data,
+                # headers=headers,
```
matan84, line 71: Can we remove this?

cloud_function/CONTRIBUTING.md
matan84: + how should we configure the Cloud fucntion?

matan84 (changes requested):
Left comments, We're getting closer!

PR #3163, opened 2026-04-29
[Core] add lifecycle notifications and raw data batching

ivankalinovski:
# Description

What - Add LifecycleClient for the ingest lifecycle API and notify aborted resyncs on graceful shutdown when transform is enabled. Buffer the data sent to lakehouse.

Why - DSP rollout.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


port_ocean/core/integrations/mixins/live_events.py
```diff
@@ -23,6 +23,10 @@ async def sync_raw_results(self, webhook_events_raw_result: list[WebhookEventRaw
         if await is_lakehouse_data_enabled():
             await self._send_webhook_raw_data_to_lakehouse(webhook_events_raw_result)
 
+        if await is_dsp_mode_enabled():
+            logger.info("DSP mode active: skipping transform and load for live events")
+            return
```
raskinmaya, line 30: `sync_raw_results` awaits `is_lakehouse_data_enabled()` then `is_dsp_mode_enabled()`. Each calls `ocean.port_client.get_organization_feature_flags()` (see [here](https://github.com/port-labs/ocean/pull/3163/changes#diff-7459ee67d5bdb7c3d4f122b015af9402d100aa32eee18c957c195867caae82d0R56) & [here](https://github.com/port-labs/ocean/pull/3163/changes#diff-7459ee67d5bdb7c3d4f122b015af9402d100aa32eee18c957c195867caae82d0L30)). Under webhook load this doubles Port traffic for the same payload. We could have a single helper that returns `(lakehouse, dsp)` or maybe cache flags for the request

port_ocean/core/integrations/mixins/sync_raw.py
raskinmaya, line 1: I'm going to leave a single comment for this, but you snapshot `dsp_mode` once at the start of `sync_raw_all`, but later gates call await is_dsp_mode_enabled() again.
This means that if those flags change mid-resync, you can mix behaviors (e.g. extract path through DSP, lifecycle notifications skipped or vice verca). Would be better to treat DSP as fixed for the whole resync (use only `dsp_mode`)

port_ocean/ocean.py
```diff
@@ -106,6 +108,10 @@ def __init__(
         self.resync_state_updater = ResyncStateUpdater(
             self.port_client, self.config.scheduled_resync_interval
         )
+        self.lifecycle_client: LifecycleClient = LifecycleClient(
+            base_url=str(self.config.port.ingest_url),
+            auth=self.port_client.auth,
+        )
```
raskinmaya, line 115: This is slightly related to qudo's comment here, but `LifecycleClient` is constructed with `self.config.port.ingest_url`. The client docstring says “integration-life-cycle”/“external transform service”. If the lifecycle API is not actually served from ingest_url, every notification is a silent failure (only warnings)

port_ocean/clients/lifecycle.py
```diff
@@ -0,0 +1,110 @@
...
+        status = body.get("status", "unknown")
+        try:
+            headers = await self.auth.headers()
+            async with httpx.AsyncClient(timeout=httpx.Timeout(10)) as client:
+                response = await client.post(url, headers=headers, json=body)
+            if response.is_error:
+                logger.warning(
+                    f"Lifecycle API returned an error for status={status}",
+                    status_code=response.status_code,
+                    response_body=response.text,
```
raskinmaya: Here and on [line 42](https://github.com/port-labs/ocean/pull/3163/changes#diff-a1798129497af9a7954eef78fd0d0cf89517afe9a5567baa20288a51cc01a059R42) you return `response.text`. I think a buggy/chatty upstream can bloat logs. Maybe a bounded truncate would be better here

port_ocean/ocean.py
ivankalinovski: the ingest endpoint is the one serving those calls, do you think it'll better fit port api?

port_ocean/clients/lifecycle.py
```diff
@@ -0,0 +1,110 @@
...
+    async def _notify(
+        self, resync_id: str, integration_id: str, body: dict[str, Any]
+    ) -> None:
+        """POST /v1/lifecycle/{resync_id}/{integration_id} -- best-effort, never raises."""
+        url = f"{self.base_url}/v1/lifecycle/{resync_id}/{integration_id}"
+        status = body.get("status", "unknown")
+        try:
+            headers = await self.auth.headers()
+            async with httpx.AsyncClient(timeout=httpx.Timeout(10)) as client:
+                response = await client.post(url, headers=headers, json=body)
```
raskinmaya: Each `_notify` builds a new `httpx.AsyncClient`. For bursts this is extra TLS setup and is atypical. Why not use a shared client?

port_ocean/core/integrations/mixins/sync_raw.py
ivankalinovski: we will remove this check for the feature flag once we release it publicly.
for now we dont want noise from curious minds.

port_ocean/core/integrations/mixins/live_events.py
ivankalinovski: iv'e added cache with ttl.

port_ocean/clients/lifecycle.py
raskinmaya, line 1: I'm going to leave a single comment for this, but there are four neat-identical `notify_*` methods. A private `_build_body(status, **fields)` would shrink the surface

port_ocean/tests/clients/test_lifecycle_client.py
raskinmaya, line 1: I'm going to leave a single comment for this, but the same `patch` + `AsyncMock` + `__aenter__`/`__aexit__` pattern in almost every test. Can't we extract a small fixture or helper for this?

port_ocean/ocean.py
raskinmaya: Not exactly what I meant. If we ever split lifecycle onto another host, we could add a dedicated `lifecycle_base_url` (defaulting to `ingest_url`) so operators don’t overload `ingest` semantically

raskinmaya (approved):

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -1134,6 +1152,12 @@ async def sync_raw_all(
             )
             logger.info(f"Resync will use the following mappings: {json.loads(app_config.json())}")
 
+            dsp_mode = await is_dsp_mode_enabled()
+            if dsp_mode:
+                logger.info(
+                    "DSP mode active: ocean-core will skip transform, load and reconciliation"
+                )
```
raskinmaya: Not sure if I left a similar comment for this already, but you snapshot `dsp_mode` once at the start of `sync_raw_all` but gate `notify_started` on a new call, skip reconciliation on the snapshot, and gate `notify_finished` after the non-DSP path on another fresh call. Cached flags + TTL make "stale vs fresh" mostly aligned, but the model is still a bit sloppy: toggles or cache expiry mid-resync can yield lifecycle started/not started/finished out of sync with what ocean actually did

port_ocean/core/integrations/mixins/lakehouse_buffer.py
```diff
@@ -0,0 +1,92 @@
...
+
+    async def add(self, batch_items: LakehouseDataEntry) -> None:
+        """Append items to the buffer, flushing if the interval or size cap is reached."""
+        self._buffer.append(batch_items)
+        self._current_size_bytes += len(json.dumps(batch_items).encode("utf-8"))
+
+        now = time.monotonic()
+        if self._last_flush_at is None:
+            self._last_flush_at = now
+            return
```
raskinmaya: On the first add, this sets `_last_flush_at` and returns before checking `size_exceeded`. A single enormous batch can blow past `_DEFAULT_MAX_SIZE_BYTES` and still not flush until a second add. For huge third-party payloads, that defeats the cap and spikes memory/HTTP size

port_ocean/clients/port/mixins/integrations.py
```diff
@@ -335,27 +328,61 @@ async def post_integration_raw_data(
...
+            "kind": event["kind"],
             "eventType": (
-                event_type.value if event_type else LakehouseEventType.LIVE_EVENT.value
+                event["event_type"].value
+                if event["event_type"]
+                else LakehouseEventType.LIVE_EVENT.value
             ),
+            "extractionTimestamp": event["extraction_timestamp"],
+            "data": data,
         }
```
raskinmaya, line 375: Isn't this a breaking change?

ivankalinovski: we are alpha, that's fine. ocean is the only source sending data.

port_ocean/clients/lifecycle.py
```diff
@@ -0,0 +1,113 @@
...
+                    f"Lifecycle API returned an error for status={status} {escaped_response_text}",
+                    status_code=response.status_code,
+                    response_body=_truncate(response.text),
+                )
+            else:
+                logger.info(
+                    f"Lifecycle API notified successfully for status={status}",
+                    status_code=response.status_code,
+                    response_body=_truncate(response.text),
+                )
```
raskinmaya, line 55: You build `escaped_response_text` for loguru brace safety, then also pass `response_body=_truncate(response.text)` - I think the main message can still embed the full escaped body, no? This would duplicate content and can flood logs on large error pages

port_ocean/core/integrations/mixins/lakehouse_buffer.py
ivankalinovski: nice catch, removed the return.

PR #3175, opened 2026-05-03
[Integration][PagerDuty] Handled daily rate limits for `analytics` endpoints

dennis-quabynah-bilson:
- Added support for respecting PagerDuty's `daily-ratelimit-*` headers on analytics endpoints, preventing excessive 429 errors and dropped enrichment.
- Introduced `PagerDutyDailyRateLimitExceededError` to manage daily quota exhaustion.
- Enhanced logging for skipped analytics enrichment due to rate limit issues.
- Updated version to 0.5.50 in `pyproject.toml` and documented changes in `CHANGELOG.md`.
- Added tests to ensure correct behavior under rate limit conditions.

# Description

What - Teach the PagerDuty integration to honor the analytics-only daily-ratelimit-* headers so resyncs stop hammering 429s once the daily quota is exhausted.

Why - Analytics endpoints (/analytics/metrics/..., /analytics/raw/...) enforce a separate, much smaller daily quota (e.g. 10/day) than the per-minute one (~960/min). The current rate limiter only parses ratelimit-* headers, so it never noticed daily exhaustion. Worse, OceanAsyncClient's default retry transport silently retries 429s using ratelimit-reset (~56s), burning ~10 attempts × ~56s ≈ 9 minutes per call before our code even sees the failure — and every concurrent in-flight analytics call enters the same doomed loop. Net effect: long resyncs, dropped analytics enrichment, no clear signal in logs.

How  
- Extended `RateLimiterRequiredHeaders` with the three `daily-ratelimit-*` aliases and added `daily_rate_limit_info` tracking on `PagerDutyRateLimiter`. Daily state is updated only when those headers are present so REST responses don't clobber it.
- Added PagerDutyRateLimiter.check_daily_budget(endpoint) and a PagerDutyDailyRateLimitExceededError. send_api_request calls it before acquiring the semaphore so subsequent analytics calls short-circuit without HTTP.
- Introduced PagerDutyRetryTransport(RetryTransport) (wired via OceanAsyncClient(transport_class=...)). It feeds the limiter from every response (including intermediate 429s) via after_retry_async, and overrides _should_retry_async to return False for any 429 with daily-ratelimit-remaining ≤ 0 — so the very first daily-quota 429 surfaces in seconds and concurrent retry loops also bail out on their next iteration.
- main.py catches PagerDutyDailyRateLimitExceededError narrowly in service and incident analytics enrichment paths, logs at WARNING, and returns un-enriched entities with __analytics: None instead of erroring.
- Added unit tests for the new headers, check_daily_budget, the transport hooks, and the end-to-end retry loop short-circuit.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination

## Screenshots

Screenshot 2026-05-03 at 11 15 42 AM
image
Screenshot 2026-05-03 at 12 23 34 PM

## API Documentation

### DIFF
--- a/integrations/pagerduty/clients/pagerduty.py
+++ b/integrations/pagerduty/clients/pagerduty.py
@@ -337,6 +337,7 @@ class PagerDutyClient(OAuthClient):
 
                 if (
                     status_code == HTTPStatus.TOO_MANY_REQUESTS
+                    and endpoint.startswith("analytics/")
                     and daily_quota_exhausted(e.response.headers)
                 ):
                     raise PagerDutyDailyRateLimitExceededError(
--- a/integrations/pagerduty/clients/pagerduty.py
+++ b/integrations/pagerduty/clients/pagerduty.py
@@ -12,7 +12,12 @@ from port_ocean.helpers.async_client import OceanAsyncClient
 from port_ocean.helpers.retry import RetryConfig
 
 from clients.utils import get_date_range_for_last_n_months
-from clients.rate_limiter import PagerDutyRateLimiter
+from clients.rate_limiter import (
+    PagerDutyDailyRateLimitExceededError,
+    PagerDutyRateLimiter,
+    daily_quota_exhausted,
+)
+from clients.retry_transport import PagerDutyRetryTransport
 
 USER_KEY = "users"
 
@@ -29,7 +34,12 @@ class PagerDutyClient(OAuthClient):
         self.token = token
         self.api_url = api_url
         self.app_host = app_host
+        self._rate_limiter = PagerDutyRateLimiter(
+            max_concurrent=MAX_CONCURRENT_REQUESTS,
+        )
         self.http_client = OceanAsyncClient(
+            transport_class=PagerDutyRetryTransport,
+            transport_kwargs={"rate_limiter": self._rate_limiter},
             retry_config=RetryConfig(
                 additional_retry_status_codes=[HTTPStatus.INTERNAL_SERVER_ERROR],
                 retry_after_headers=[
@@ -39,9 +49,6 @@ class PagerDutyClient(OAuthClient):
             timeout=ocean.config.client_timeout,
         )
         self.http_client.headers.update(self.headers)
-        self._rate_limiter = PagerDutyRateLimiter(
-            max_concurrent=MAX_CONCURRENT_REQUESTS,
-        )
 
     @classmethod
     def from_ocean_configuration(cls) -> "PagerDutyClient":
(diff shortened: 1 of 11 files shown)


integrations/pagerduty/main.py
```diff
@@ -43,6 +44,11 @@ async def fetch_service_analytics(
                 for service in services
             ]
             return enriched_services
+        except PagerDutyDailyRateLimitExceededError as e:
```
oiadebayo: Let's move this function into a utils file and test it. Mock possible errors and ascertain that it elicit expected behavior on the error

integrations/pagerduty/clients/retry_transport.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    def __init__(self, *, rate_limiter: PagerDutyRateLimiter, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._rate_limiter = rate_limiter
+
+    async def after_retry_async(
+        self, request: httpx.Request, response: httpx.Response, attempt: int
+    ) -> None:
+        if await self._should_retry_async(response):
```
oiadebayo: Why this condition? Should we not update limiter with headers regardless?

integrations/pagerduty/clients/pagerduty.py
```diff
@@ -320,12 +329,20 @@ async def send_api_request(
...
                     return {}
 
+                if (
+                    status_code == HTTPStatus.TOO_MANY_REQUESTS
+                    and daily_quota_exhausted(e.response.headers)
+                ):
+                    raise PagerDutyDailyRateLimitExceededError(
+                        f"PagerDuty analytics daily quota exhausted on {endpoint}."
+                    ) from e
+
```
oiadebayo, line 346: If we are confident this applies to only the analytics endpoint then let's explicitly scope it to that

oiadebayo (changes requested):
I left some comments

integrations/pagerduty/clients/rate_limiter.py
```diff
@@ -84,22 +96,60 @@ def _parse_rate_limit_headers(
...
+        info = self._parse_per_minute_headers(parsed)
         if info:
             self.rate_limit_info = info
             logger.debug(
-                f"Rate limit hit on {resource}: {info.remaining}/{info.limit} remaining "
+                f"Rate limit on {resource}: {info.remaining}/{info.limit} remaining "
                 f"(resets in {info.seconds_until_reset}s)"
             )
+
+        daily_info = self._parse_daily_headers(parsed)
```
danamsalem-ux: daily_rate_limit_info

danamsalem-ux (approved):

oiadebayo (approved):
Lgtm

PR #3183, opened 2026-05-07
[Core] Fix Resync Mappings Logging to be JSON Safe

emekanwaoma:
# Description

What -
- Updated resync mappings logging to use JSON-safe serialization (`json.loads(app_config.json())`) instead of `app_config.dict()`.
- Added `port_ocean/core/integrations/mixins/changelog.MD` entry for the change.

Why -
- Prevent issues with non-JSON-serializable values/types in logged mappings and keep logs consistently JSON-compatible.

How -
- Convert the Port app config model to JSON (`.json()`), then back to plain Python types (`json.loads`) for logging.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)


omby8888 (approved):
LGTM

PR #3193, opened 2026-05-10
[Integration][Github] Prevent IncludedFiles README Fetch from Crashing on Missing Content

emekanwaoma:
# Description

What -

Fix GitHub `includedFiles` enrichment so repository resync doesn’t crash when the Contents API returns a response without `content`/`encoding` or when the requested path is not a regular file (e.g., symlink/submodule).

Why -

Large org syncs can consistently fail once they hit a repository whose README (or other included file) can’t be decoded, causing a `KeyError: 'content'` and aborting ingestion.

How -

- Added response validation before decoding file contents (size/type/content/encoding).
- Return `content: null` for invalid/unsupported responses instead of throwing.
- Added unit tests for missing `content`, missing `encoding`, non-file `type`, and missing `type` behavior.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)


oiadebayo (approved):
LGTM

omby8888 (approved):
Nice work

PR #3195, opened 2026-05-11
[Integration][Jira] Added support for `backlog` kind 

eriport:
# Description

What -

Adds `backlog` as a new kind with a graceful fall-through to the legacy `agile/1.0` endpoint via the `useSoftwareApi` selector flag. Backlog issues are upserted as `jiraIssue` entities enriched with a `board` relation, so existing `jiraIssue` reconciliation continues to work end-to-end alongside the existing `issue` kind.

Why -

Part of the Jira Agile kinds feature request to support Boards, Sprints, Epics, and Backlog in Port's Jira integration. Backlog completes the per-board issue grouping for customers building catalog views around sprint planning and work-in-progress visualization. 

How -

## Type of change

### Core testing checklist

- [NA] Resync able to detect and delete entities

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

Screenshot 2026-05-11 at 15 32 53
Screenshot 2026-05-29 at 12 49 36
Screenshot 2026-05-29 at 12 49 45
Screenshot 2026-05-31 at 10 49 53

## API Documentation


integrations/jira/jira/overrides.py
```diff
@@ -125,6 +125,62 @@ class JiraBoardResourceConfig(ResourceConfig):
...
+    )
+    use_software_api: bool = Field(
+        alias="useSoftwareApi",
+        default=True,
+        title="Use Software API",
+        description=(
+            "When true, attempts the newer rest/software/1.0 backlog endpoint first, "
+            "falling back to the legacy rest/agile/1.0 endpoint on auth failure. "
+            "The agile/1.0 endpoint is deprecated and scheduled for removal on "
+            "November 1, 2026. Set to false to force legacy behavior."
```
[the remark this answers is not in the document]
oiadebayo: Consider this comment, the description does not seem to represent what is happening

integrations/jira/jira/client.py
```diff
@@ -676,3 +766,53 @@ async def enrich_board_with_projects(self, board: dict[str, Any]) -> dict[str, A
...
+            async for backlogs in self._fetch_backlog_from_software_api(
+                board_id, query_params
+            ):
+                yield backlogs
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == 400:
+                logger.warning(
+                    f"Board {board_id} returned 400 — board type does not support backlog "
+                    f"(e.g. Kanban or backlog disabled), skipping"
+                )
```
oiadebayo: Are we absolutely sure this is the only reason a user will get 400 with this call? Can other issues like having a malformed request jql get a 400 as well? If other cases exists then we should not mislead the users in the log. Are we sure we should skip in all cases as well

integrations/jira/jira/client.py
```diff
@@ -417,6 +476,37 @@ async def _create_events_webhook_oauth(self, app_host: str) -> None:
...
+        """Fetch backlog via the ``software`` endpoint."""
+        software_url = await self._get_software_api_url()
+        url = f"{software_url}/board/{board_id}/backlog"
+        async for issue_batch in self._get_token_paginated_data(url, "issues", params):
+            yield self._enrich_with_board_id(issue_batch, board_id)
+
+    @deprecated(
+        "_fetch_backlog_from_agile_api is deprecated and will be removed in favor of _fetch_backlog_from_software_api by November 1, 2026. The software API endpoint provides more efficient pagination and is the current standard for backlog retrieval. Please migrate to the software API endpoint as soon as possible."
+    )
+    async def _fetch_backlog_from_agile_api(
```
oiadebayo: I think we should add a warning log in this function rather than using this decorator here. Also we need to reword the message so it does not mention private methods. Let's make it straightforward for users who will see this to understand

integrations/jira/main.py
```diff
@@ -124,6 +125,32 @@ async def on_resync_boards(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+
+    logger.info(f"Starting backlog resync with JQL filter: '{selector.jql or 'none'}'")
+
+    async for board_batch in client.get_paginated_boards():
+        boards_with_ids = [board for board in board_batch if board.get("id")]
+
+        backlog_streams = [
+            client.get_paginated_backlog_for_board(
+                board_id=board["id"],
+                jql=selector.jql or None,
```
oiadebayo: The description of the jql in the backlog selector says you should add empty string to fetch all backlogs but empty string is falsy which means it will effectively become None here. Is this the intended behavior?

oiadebayo (changes requested):
Left some comments

integrations/jira/jira/overrides.py
```diff
@@ -125,6 +125,62 @@ class JiraBoardResourceConfig(ResourceConfig):
...
+    fields: list[str] | None = Field(
+        default=None,
+        title="Fields",
+        description=(
+            "Specific issue fields to return. Omit to return all fields. "
+            "Use field projection to reduce payload size on large boards, "
+            "e.g. ['id', 'key', 'summary', 'status', 'assignee', 'priority']."
+        ),
+    )
+    max_results: int = Field(
```
omby8888: Why this is configurable?

omby8888 (changes requested):

integrations/jira/jira/client.py
eriport: yes this is intentional, although this also could happen for invalid jql. i would update this to be more explicit and yes, skipping here is intentional, the goal of the log here is to give context on why that is neccessary.

Screenshot 2026-05-29 at 12 53 04

eriport: > update: updated

oiadebayo: I still believe we should explicitly handle for the case where the 400 is caused by boards not supporting backlog. User induced error like incorrect jql should still break the integration and skip reconciliation so users won't lose data when they accidentally enter an incorrect jql and the integration complete resync without any indications of failure as reconciliation will then happen. You can skip for 400 when error message matches the one shown in the screenshot above for instance

omby8888 (approved):

oiadebayo (approved):
LGTM

PR #3196, opened 2026-05-11
[Core]: update ocean polling event listener to query from resync requests as well

Hadar-Vaknin:
# Description

What -
 Polling event listener: when the organization feature flag `OCEAN_POLLING_INTEGRATION_RESYNC_REQUESTS_ENABLED` is present, Ocean polls the integration resync-request endpoint when the integration document’s `updatedAt` is unchanged, and runs a resync when the request’s `updatedAt` is newer than the last stored watermark.

Why -
As a part of the manual resync refactor.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


port_ocean/core/event_listener/polling.py
```diff
@@ -77,12 +111,32 @@ async def resync() -> None:
...
+            is_resync_request_triggered = False
+
+            if not should_resync and await self.is_resync_requests_polling_enabled():
+                resync_request = (
+                    await ocean.app.port_client.get_integration_resync_request()
+                )
+                resync_request_updated_at = resync_request.get("updatedAt", "")
+                should_resync = self.should_resync_from_resync_request(
+                    resync_request_updated_at
+                )
```
raskinmaya: If `get_integration_resync_request()` throws a network error or returns a non-2xx status, the exception will bubble up and potentially crash the polling loop. The existing `get_current_integration` call on line 110 also doesn't handle this, but adding a second unhandled API call doubles the failure surface.

Either wrap in try/except with proper logging, or use `should_raise=False`

port_ocean/core/event_listener/polling.py
```diff
@@ -58,6 +60,38 @@ def should_resync(self, last_updated_at: str) -> bool:
...
+    def should_resync_from_resync_request(self, last_updated_at: str | None) -> bool:
+        _last_updated_at = ocean.app.resync_state_updater.last_resync_request_updated_at
+
+        if not last_updated_at:
+            return False
+
+        try:
+            resync_request_updated_at = convert_str_to_utc_datetime(last_updated_at)
+        except (TypeError, ValueError):
+            return False
```
raskinmaya, line 85: `convert_str_to_utc_datetime` never actually returns `None` - it raises on invalid input. The return type is misleading (I know it's not your code but FYI). Also, catching `TypeError` is unnecessary since the function only calls `fromisoformat()` which raises `ValueError` for invalid formats

port_ocean/core/event_listener/polling.py
```diff
@@ -77,12 +111,32 @@ async def resync() -> None:
...
-                logger.info("Detected change in integration, resyncing")
+            if should_resync:
+                logger.info(resync_reason)
                 ocean.app.resync_state_updater.last_integration_state_updated_at = (
                     last_updated_at
                 )
+                if is_resync_request_triggered:
+                    ocean.app.resync_state_updater.last_resync_request_updated_at = (
+                        resync_request_updated_at
+                    )
```
raskinmaya, line 172: When resync is triggered by integration change, `last_resync_request_updated_at` is NOT reset. This means if:
1. Resync request triggers (sets `last_resync_request_updated_at`)
2. Integration changes later (updates only `last_integration_state_updated_at`)
3. Same resync request timestamp returns from API

Then step 3 won't trigger a redundant resync, which is correct. But consider: what if the API returns a resync request with a timestamp older than our stored one? The comparison `current_state_updated_at < resync_request_updated_at` (line 82) would return `False`, which is correct. Whether it was an accident or intentional in this case, this asymmetry is something that I feel that we should be documenting

port_ocean/clients/port/mixins/integrations.py
```diff
@@ -100,6 +100,22 @@ async def get_current_integration(
...
         return response.json().get("integration", {})
 
+    async def _get_integration_resync_request(self) -> httpx.Response:
+        logger.info(
+            f"Fetching integration resync request for id: {self.integration_identifier}"
+        )
+        return await self.client.get(
+            f"{self.auth.api_url}/integration/{self.integration_identifier}/resync-request",
+            headers=await self.auth.headers(),
+        )
```
raskinmaya, line 107: Private methods generally don't log; the public wrappers handle logging via `handle_port_status_code`. This is inconsistent with codebase conventions. Would be better to remove the info logs

port_ocean/core/event_listener/polling.py
```diff
@@ -77,12 +111,32 @@ async def resync() -> None:
...
+                resync_request = (
+                    await ocean.app.port_client.get_integration_resync_request()
+                )
+                resync_request_updated_at = resync_request.get("updatedAt", "")
+                should_resync = self.should_resync_from_resync_request(
+                    resync_request_updated_at
+                )
+                if should_resync:
+                    resync_reason = "Detected integration resync request"
+                    is_resync_request_triggered = True
```
raskinmaya: `is_resync_request_triggered` tracks whether we entered the resync-request branch and `should_resync` became `True`. But `resync_request_updated_at` already serves as this indicator - it's only set to a non-empty value in that branch, so I think there's a redundancy here. We can remove `is_resync_request_triggered` and use `resync_request_updated_at` instead

port_ocean/core/event_listener/polling.py
raskinmaya, line 1: Can we add docstrings both for `is_resync_requests_polling_enabled` and `should_resync_from_resync_request`?

port_ocean/tests/core/event_listener/test_polling.py
```diff
@@ -0,0 +1,255 @@
...
+
+def _run_twice_repeat_every(*_args, **_kwargs):
+    def decorator(func):
+        async def wrapped():
+            await func()
+            await func()
+
+        return wrapped
+
+    return decorator
```
raskinmaya: Duplicate test helpers could be consolidated.
You could use a factory, something like:

```
def _make_repeat_every(times: int):
    def repeat_every_mock(*_args, **_kwargs):
        def decorator(func):
            async def wrapped():
                for _ in range(times):
                    await func()
            return wrapped
        return decorator
    return repeat_every_mock
```

port_ocean/tests/core/event_listener/test_polling.py
raskinmaya, line 1: There's a test for empty string and invalid string ("not-a-timestamp"), but no test for `None`. This matters as long as `should_resync_from_resync_request` accepts `str | None`

port_ocean/core/event_listener/polling.py
```diff
@@ -58,6 +60,38 @@ def should_resync(self, last_updated_at: str) -> bool:
 
         return _last_updated_at != last_updated_at
 
+    def should_resync_from_resync_request(self, last_updated_at: str | None) -> bool:
+        _last_updated_at = ocean.app.resync_state_updater.last_resync_request_updated_at
```
raskinmaya: Using underscore as a prefix typically indicates a private/internal variable, but it's a local variable

raskinmaya:
Nice, left some comments. Could you please also indicate Copilot's handled comments?

port_ocean/core/event_listener/polling.py
Hadar-Vaknin: I wrapped the `get_current_integration` as well, let me know what u think

port_ocean/core/event_listener/polling.py
Hadar-Vaknin: This is intentional behavior: last_resync_request_updated_at is treated as a monotonic watermark for request-driven resyncs, not a generic resync marker. We only advance it when a resync is actually caused by a resync request. Integration-change resyncs should not reset or alter that watermark. As a result, stale or replayed request timestamps are safely ignored (stored >= incoming => no trigger).
I will document it.

raskinmaya (approved):

PR #3203, opened 2026-05-13
[Integration][GitHub] Cache Recursive Tree Fetch In File Exporter

emekanwaoma:
# Description

What -
- Cache `RestFileExporter.get_tree_recursive()` results per repo/branch to avoid re-fetching the same git tree during file matching.

Why -
- Reduces GitHub API calls and speeds up file resync + webhook processing; lowers risk of rate-limit pressure.

How -
- Added `@cache_coroutine_result()` to `get_tree_recursive` so repeated calls reuse the same response.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### DIFF
--- a/integrations/github/github/webhook/clients/base_webhook_client.py
+++ b/integrations/github/github/webhook/clients/base_webhook_client.py
@@ -77,7 +77,6 @@ class BaseGithubWebhookClient(GithubRestClient):
         self, webhook_id: str, config_data: dict[str, str], target: HookTarget
     ) -> None:
         webhook_data = {"config": config_data}
-        logger.info(f"Patching webhook {webhook_id} with data {webhook_data}")
 
         await self.send_api_request(
             target.hook_url(webhook_id),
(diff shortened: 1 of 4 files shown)


emekanwaoma:
The log was removed because we noticed for long secrets, it redact only a small portion of it.

oiadebayo (approved):
LGTM

integrations/github/github/webhook/clients/base_webhook_client.py
```diff
@@ -77,7 +77,6 @@ async def _patch_webhook(
         self, webhook_id: str, config_data: dict[str, str], target: HookTarget
     ) -> None:
         webhook_data = {"config": config_data}
-        logger.info(f"Patching webhook {webhook_id} with data {webhook_data}")
```
oiadebayo, line 80: Why did you remove this log?

omby8888 (approved):
LGTM

PR #3209, opened 2026-05-13
[Integration][Azure DevOps] Fixed retry transient 503 on `itemsbatch`

dennis-quabynah-bilson:
# Description

**What**
- **Azure DevOps**: opt the `itemsbatch` POST in `_get_files_by_descriptors` into the shared retry transport by passing `extensions={"retryable": True}` through a new `extensions` parameter on `HTTPBaseClient.send_request`. Transient 503 responses from Azure DevOps no longer take down the file kind resync.

**Why**
A customer's resync was aborting with a `503 Service Unavailable` from `dev.azure.com/.../itemsbatch`. The 503 is transient and recoverable, but today it's fatal because the shared `RetryConfig` skips POSTs by default, so the `itemsbatch` lookup never gets a second chance and the whole file kind resync dies.

**How**
- `integrations/azure-devops/azure_devops/client/base_client.py`: `send_request` now accepts and forwards an `extensions` dict.
- `integrations/azure-devops/azure_devops/client/azure_devops_client.py`: `_get_files_by_descriptors` passes `extensions={"retryable": True}`. Scope is intentionally narrow — only this known-idempotent POST opts in; webhook creation and other non-idempotent POSTs are unaffected.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/azure-devops/azure_devops/client/azure_devops_client.py
```diff
@@ -1642,6 +1642,7 @@ async def _get_files_by_descriptors(
                     params=API_PARAMS,
                     data=json.dumps(request_data),
                     headers={"Content-Type": "application/json"},
+                    extensions={"retryable": True},
```
danamsalem-ux: how do we defer here from other errors besides 503s?

port_ocean/core/handlers/entity_processor/jq_entity_processor_sync.py
```diff
@@ -94,11 +94,15 @@ def _search(data: dict[str, Any], pattern: str, field: str | None = None) -> Any
...
-        if isinstance(value, bool):
-            return value
+            try:
+                value = compile_jq(compiled_pattern, data).first()
+            except JQInputNotJsonSerializableError:
+                value = compile_jq(compiled_pattern, make_json_compatible(data)).first()
+            if isinstance(value, bool):
+                return value
+        except ValueError as exc:
+            JQEntityProcessorSync._log_search_failure(pattern, exc)
```
danamsalem-ux: This will affect all integrations, are there other use cases we might be missing where we dont want to return false here

dennis-quabynah-bilson: This change was made to conform with the other `search_*` operations when handling JQ compilation. I noticed that the other operations handled exceptions properly but `_search_as_bool` re-raises it. Is there a reason why we had this in the first place please?

integrations/azure-devops/azure_devops/client/azure_devops_client.py
dennis-quabynah-bilson: Thanks @danamsalem-ux . I am refactoring to scope to only 503s

oiadebayo (commented):
Let's separate concerns, move ocean core changes into a different PR

dennis-quabynah-bilson:
> Let's separate concerns, move ocean core changes into a different PR

@oiadebayo  Done

oiadebayo (approved):
LGTM

danamsalem-ux (approved):

PR #3212, opened 2026-05-14
[Integration][Jira] Add support for Worklogs

eriport:
# Description

What - Add `worklog` kind to the Jira integration to sync time tracking entries into Port as `jiraWorklog` entities, with a relation back to the parent `jiraIssue`.

Why - Worklog is a Canny-requested feature. Time tracking is a core Jira workflow and customers need visibility into time logged per issue directly in Port.

How -

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

Screenshot 2026-05-14 at 16 51 29
Screenshot 2026-05-14 at 16 51 56

## API Documentation


integrations/jira/jira/client.py
```diff
@@ -676,3 +681,27 @@ async def enrich_board_with_projects(self, board: dict[str, Any]) -> dict[str, A
...
         board["__projectKeys"] = project_keys
         return board
+
+    async def get_paginated_worklogs_for_issue(
+        self,
+        issue_key: str,
+        max_results: int = 5000,
+        started_after: int | None = None,
+        started_before: int | None = None,
+        expand: str | None = None,
```
omby8888: Can we consolidate those into single type?

integrations/jira/webhook_processors/worklog_webhook_processor.py
```diff
@@ -0,0 +1,79 @@
...
+    WebhookEventRawResults,
+)
+from port_ocean.core.handlers.webhook.abstract_webhook_processor import (
+    AbstractWebhookProcessor,
+)
+
+from initialize_client import get_or_create_jira_client
+
+
+WORKLOG_WEBHOOK_EVENTS = frozenset(
```
omby8888: Can we take those from the predefined array in the client?

integrations/jira/jira/overrides.py
```diff
@@ -125,6 +125,61 @@ class JiraBoardResourceConfig(ResourceConfig):
...
+    started_before: int | None = Field(
+        alias="startedBefore",
+        default=None,
+        title="Started Before",
+        description=(
+            "UNIX timestamp in milliseconds. "
+            "Only return worklogs started before this time."
+        ),
+    )
+    expand: str | None = Field(
```
omby8888, line 144: Can we allaborate the description here? Its not clear what extra data it gives, and no indication if thats affecting performance

integrations/jira/jira/overrides.py
```diff
@@ -125,6 +125,61 @@ class JiraBoardResourceConfig(ResourceConfig):
...
+    jql: str = Field(
+        default="updated >= -1w",
+        title="JQL Filter",
+        description=(
+            "JQL to scope which issues' worklogs are fetched. "
+            "Defaults to issues updated in the last week to avoid "
+            "re-fetching all worklogs on every resync."
+        ),
+    )
+    max_results: int = Field(
```
omby8888: Why this is configurable? What benefit it gives to customers to tune this differently?

omby8888 (changes requested):

integrations/jira/webhook_processors/worklog_webhook_processor.py
omby8888: u can instead expose there dedicated worklog variable and use it here and inject it in the global webhook events var instead of using startswith

omby8888 (approved):
Nice!

PR #3219, opened 2026-05-14
[Core] Make JQEntityProcessor._search_as_bool resilient to runtime ValueErrors

dennis-quabynah-bilson:
# Description

What -
- **Core** (`port_ocean.core.handlers.entity_processor`): in `JQEntityProcessor._search_as_bool` (and the sync mirror `JQEntityProcessorSync._search_as_bool`), wrap the JQ runtime evaluation in `try/except ValueError`. A failing selector query now logs a structured WARNING with the offending pattern and returns `False` (entity is filtered out) instead of bubbling a fatal `ValueError` that aborts the whole batch.

Why -
A customer's resync was aborting with an `ExceptionGroup` of four `ValueError: null (null) cannot be matched, as it is not a string` — their mapping uses `test()` / `match()` against a nullable upstream field, and one null value was poisoning the entire batch. The JQ error is per-entity and recoverable: today it's fatal only because `_search_as_bool` doesn't guard `func.first()` against runtime errors. Skipping the offending entity with a warning preserves the rest of the resync.

How -
- `port_ocean/core/handlers/entity_processor/jq_entity_processor.py`: minimal `try/except ValueError` around the JQ `.first()` call inside `_search_as_bool`; log a WARNING and return `False`. Compile errors and "expected boolean" errors remain loud.
- `port_ocean/core/handlers/entity_processor/jq_entity_processor_sync.py`: same change applied to the sync implementation so behavior stays consistent across both processors.
- Tests added in `port_ocean/tests/core/handlers/entity_processor/test_jq_entity_processor.py` and `test_jq_entity_processor_sync.py` covering the `ValueError → False + warning` path.
- `pyproject.toml` / `CHANGELOG.md`: bump to `0.42.2`, bug-fix entry under "Bug Fixes".

## Type of change

### Core testing checklist

- [ ] Specifically validated: a mapping whose selector calls `test()` / `match()` against a nullable field now logs a WARNING and skips the offending entity instead of aborting the batch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- towncrier release notes start -->
+## 0.42.2 (2026-05-14)
+
+### Bug Fixes
+
+- Fixed JQ selector queries crashing the batch when a regex operator receives null input now skipping with a warning.
+
+
 ## 0.42.1 (2026-05-14)
 
 ### Improvements
--- a/port_ocean/core/handlers/entity_processor/jq_entity_processor.py
+++ b/port_ocean/core/handlers/entity_processor/jq_entity_processor.py
@@ -123,9 +123,13 @@ class JQEntityProcessor(BaseEntityProcessor):
         except JQInputNotJsonSerializableError:
             func = compile_jq(compiled_pattern, make_json_compatible(data))
 
-        value = func.first()
-        if isinstance(value, bool):
-            return value
+        try:
+            value = func.first()
+            if isinstance(value, bool):
+                return value
+        except ValueError as exc:
+            self._log_search_failure(pattern, exc)
+            return False
         raise EntityProcessorException(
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- towncrier release notes start -->
+## 0.42.2 (2026-05-14)
+
+### Bug Fixes
+
+- Fixed JQ selector queries crashing the batch when a regex operator receives null input now skipping with a warning.
```
danamsalem-ux, line 13: Do we want this to happen for every field? From what i understood if its required fields in the mapping that has null we should and do fail the batch processing. Do we want to change that logic?

oiadebayo: I think the required fields are processed through `_search_as_object` which is not affected by the changes in this PR. A null on the required fields should still go to the misconfiguration flow as always

oiadebayo (approved):
LGTM

dennis-quabynah-bilson:
Closing this ticket as the change requires a comprehensive assessment on how it affects other integrations.

PR #3231, opened 2026-05-17
[Integration][Azure DevOps] Reduce resync TSTU footprint to prevent webhook drops

dennis-quabynah-bilson:
# Description

**What** — Reduce the Azure DevOps API spend during webhook processing so service hooks stop being dropped when resync and live events overlap under rate-limit pressure.

**Why** — Webhook deliveries were exceeding Ocean's 90-second `asyncio.wait_for` deadline (`port_ocean/core/handlers/webhook/processor_manager.py`) when resync was actively consuming the Azure DevOps TSTU (Team Services Time Units) budget. Because each webhook event runs in its own `event_context` and constructs a fresh `AzureDevopsClient` + `AzureDevOpsRateLimiter` (per-instance state), a higher `minimum_limit_remaining` buffer is **not** a viable fix — the live event limiter has zero visibility into what resync just consumed. The only levers available without a framework change are: (1) make resync spend less, and (2) make each webhook event do less work.

**How** —

1. **Lower resync concurrency caps** in `azure_devops/client/azure_devops_client.py` to leave more TSTU headroom for live events:
   - `MAX_CONCURRENT_PROJECTS`: 5 → 3
   - `MAX_CONCURRENT_REPOS_FOR_PULL_REQUESTS`: 25 → 10
   - `MAX_CONCURRENT_REPOS_FOR_FILE_PROCESSING`: 25 → 10
   - `MAX_CONCURRENT_FILE_DOWNLOADS` (50) and `MAX_SUBJECTS_PER_LOOKUP` (500) left untouched.

2. **Drop the redundant `get_single_project` HTTP call** in both pipeline webhook processors. The ADO webhook payload already carries the project id under `resourceContainers.project`; refetching the project just to enrich a live event was burning TSTUs and adding a serial round-trip inside the 90s deadline.
   - `pipeline_run_webhook_processor.py`: pass `payload["resourceContainers"]["project"]` straight into `annotate_runs(..., project=...)`. Verified `annotate_runs` only stamps the dict onto each run as `__project`; default mapping reads `.__project.id`.
   - `pipeline_stage_webhook_processor.py`: same treatment for `get_pipeline_stage`. Verified the client only uses `project["id"]` and stamps the dict as `__project` on each stage record.

3. **Drop `LIMIT_RESET_HEADER` from `HTTPBaseClient`'s `retry_after_headers`** so `X-RateLimit-Reset` is no longer treated as a retry-after directive by Ocean's retry helper (it's a counterfactual telemetry timestamp, not a wait deadline).

### Out of scope (separate PRs)
- The 90s `asyncio.wait_for` deadline behavior in `processor_manager.py` (deadline-aware re-queue is a framework change).
- Identity isolation (second PAT / Entra service principal) — depends on a spec update.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

N/A — backend TSTU reduction.

## API Documentation

https://learn.microsoft.com/en-us/azure/devops/integrate/concepts/rate-limits?view=azure-devops#api-client-experience


integrations/azure-devops/azure_devops/client/azure_devops_client.py
```diff
@@ -66,10 +66,10 @@
 MAX_WORK_ITEMS_RESULTS_PER_PROJECT = 19999
 MAX_ALLOWED_FILE_SIZE_IN_BYTES = 1 * 1024 * 1024
 MAX_CONCURRENT_FILE_DOWNLOADS = 50
-MAX_CONCURRENT_REPOS_FOR_FILE_PROCESSING = 25
-MAX_CONCURRENT_REPOS_FOR_PULL_REQUESTS = 25
+MAX_CONCURRENT_REPOS_FOR_FILE_PROCESSING = 10
+MAX_CONCURRENT_REPOS_FOR_PULL_REQUESTS = 10
 MAX_SUBJECTS_PER_LOOKUP = 500
-MAX_CONCURRENT_PROJECTS = 5
+MAX_CONCURRENT_PROJECTS = 3
```
matan84, line 75: This seems REALLY restrictive and it will affect performance drastically, no?

integrations/azure-devops/azure_devops/webhooks/webhook_processors/pipeline_stage_webhook_processor.py
```diff
@@ -49,20 +49,10 @@ async def handle_event(
     ) -> WebhookEventRawResults:
         client = AzureDevopsClient.create_from_ocean_config()
         pipeline_id = payload["resource"]["pipeline"]["id"]
-        project_id = payload["resourceContainers"]["project"]["id"]
+        project = payload["resourceContainers"]["project"]
```
matan84, line 51: Does this mean the object in the `payload["resourceContainers"]` contains the project's data in the same schema?

integrations/azure-devops/azure_devops/webhooks/webhook_processors/pipeline_run_webhook_processor.py
```diff
@@ -43,7 +43,8 @@ async def handle_event(
     ) -> WebhookEventRawResults:
         client = AzureDevopsClient.create_from_ocean_config()
         pipeline_id = payload["resource"]["pipeline"]["id"]
-        project_id = payload["resourceContainers"]["project"]["id"]
+        project = payload["resourceContainers"]["project"]
```
matan84, line 45: Does this mean the object in the payload["resourceContainers"] contains the project's data in the same schema?

matan84 (changes requested):
left comments

integrations/azure-devops/azure_devops/webhooks/webhook_processors/pipeline_stage_webhook_processor.py
dennis-quabynah-bilson: Yes, the project object in the webhook event already includes both the id and link. In most cases, the id alone is sufficient to map relations from these kinds back to the project, which is why the additional enrichment API call was removed.

integrations/azure-devops/azure_devops/webhooks/webhook_processors/pipeline_run_webhook_processor.py
dennis-quabynah-bilson: Yes, the project object in the webhook event already includes both the id and link. In most cases, the id alone is sufficient to map relations from these kinds back to the project, which is why the additional enrichment API call was removed.

integrations/azure-devops/azure_devops/client/azure_devops_client.py
dennis-quabynah-bilson: I agree with you. I’ll bump them up to 20 instead of 25 to reduce the performance impact while still leaving enough headroom for webhook events when TSTUs are close to the rate limit.

integrations/azure-devops/azure_devops/webhooks/webhook_processors/pipeline_stage_webhook_processor.py
mk-armah: Just a deep thought.. 
is it likely that customers aren't using link or id to map ? in that case this will be a breaking changing
can you confirm ?

mk-armah (approved):
LGTM

integrations/azure-devops/azure_devops/client/rate_limiter.py
```diff
@@ -70,31 +72,27 @@ async def __aenter__(self) -> "AzureDevOpsRateLimiter":
...
-            if retry_wait > 0:
-                logger.debug(
-                    f"Rate limit: waiting {retry_wait:.2f}s due to previous rate limit"
+            wait_time = self.seconds_until_reset
+            if wait_time > 0 and self._should_sleep():
+                logger.info(
+                    f"Rate limit: waiting {wait_time:.2f}s "
+                    f"(event_type={event.event_type}, "
+                    f"remaining={self._remaining}, limit={self._limit}, "
+                    f"utilization={self.utilization:.2%})"
```
mk-armah, line 81: make this human readable, share a how a full message will look like

integrations/azure-devops/azure_devops/client/azure_devops_client.py
```diff
@@ -66,10 +66,10 @@
 MAX_WORK_ITEMS_RESULTS_PER_PROJECT = 19999
 MAX_ALLOWED_FILE_SIZE_IN_BYTES = 1 * 1024 * 1024
 MAX_CONCURRENT_FILE_DOWNLOADS = 50
-MAX_CONCURRENT_REPOS_FOR_FILE_PROCESSING = 25
-MAX_CONCURRENT_REPOS_FOR_PULL_REQUESTS = 25
+MAX_CONCURRENT_REPOS_FOR_FILE_PROCESSING = 20
+MAX_CONCURRENT_REPOS_FOR_PULL_REQUESTS = 20
 MAX_SUBJECTS_PER_LOOKUP = 500
-MAX_CONCURRENT_PROJECTS = 5
+MAX_CONCURRENT_PROJECTS = 3
```
mk-armah: lets revert this

mk-armah (changes requested):

integrations/azure-devops/azure_devops/client/base_client.py
```diff
@@ -23,7 +22,6 @@ def __init__(self, personal_access_token: str) -> None:
         self._client = OceanAsyncClient(
             retry_config=RetryConfig(
                 retry_after_headers=[
-                    LIMIT_RESET_HEADER,
                     LIMIT_RETRY_AFTER_HEADER,
                 ],
```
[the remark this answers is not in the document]
dennis-quabynah-bilson: The change was to remove the explicit declaration since it is already in use by RetryConfig

integrations/azure-devops/azure_devops/client/azure_devops_client.py
dennis-quabynah-bilson: Changes reverted after an extended discussion with the team

PR #3233, opened 2026-05-18
[Integration][aws-v3] Fixed index-shift dropping enrichment onto wrong resources in concurrent actions

oiadebayo:
# Description

What -
Fix a silent index-shift bug in `aws-v3` where enrichment data (tags, encryption, public access, ownership controls, location, queue attributes, lifecycle/repository policies, etc.) could be attached to the wrong AWS resource whenever a sibling resource in the same batch returned a recoverable error.

The bug lived in every action that used `asyncio.gather(..., return_exceptions=True)` over a batch of inputs and then `continue`-d on a recoverable error without appending a placeholder. The shortened list was then merged positionally in `ResourceInspector.inspect` against sibling actions' full-length lists, shifting every entry after the failure by one slot.

Affected services:
- S3 (`GetPublicAccessBlockAction`, `GetBucketOwnershipControlsAction`, `GetBucketEncryptionAction`, `GetBucketLocationAction`, `GetBucketTaggingAction`)
- SQS (`GetQueueAttributesAction`, `ListQueueTagsAction`)
- RDS (`ListTagsForResourceAction`)
- Lambda (`ListTagsAction`)
- Organizations (`ListTagsForResourceAction`)
- Shared helper `execute_concurrent_aws_operations`, which propagates the fix to ECR (`GetRepositoryPolicyAction`, `GetLifecyclePolicyAction`, `ListTagsForResourceAction`), ECS (`DescribeTaskDefinitionsAction`), and EC2 (`DescribeVolumeAttributeAction`)

Why -
This corrupts catalog data without any error surfaced to the user.

How -
- Every recoverable-skip site now does `results.append({})` before `continue`, preserving one entry per input item in input order. `ResourceInspector` already guards merge with `if item:` so the placeholders are no-ops at merge time, they just preserve positional alignment.
- `execute_concurrent_aws_operations` docstring updated to document the new contract ("Returns one entry per input item; an empty dict is returned for items that hit a recoverable error").
- For RDS / Lambda / SQS the "Successfully fetched X for N resources" log now tracks a separate `success_count` so the placeholder entries don't inflate the reported count.
- Added per-action index-alignment tests (3 inputs, middle one fails recoverably) for every affected action, including helper-based callers in ECR / ECS / EC2.
- Added a new `tests/core/modeling/test_resource_inspector.py` with two scenarios that lock in the contract end-to-end: one proves placeholders correlate sibling actions correctly, the other documents the buggy shape (skipped index without placeholder) and asserts the resulting misalignment so any future reintroduction of this bug class trips a test instead of silently shipping.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [x] Support Multi account

## Screenshots

N/A

## API Documentation

N/A - no new AWS API surface added; existing recoverable-error semantics preserved.


mk-armah (approved):
LGTM

integrations/aws-v3/aws/core/exporters/aws_lambda/function/actions.py
```diff
@@ -25,21 +25,24 @@ async def _execute(self, functions: list[dict[str, Any]]) -> list[dict[str, Any]
...
                     continue
                 else:
                     logger.error(
                         f"Error fetching tags for Lambda function '{function_name}': {tag_result}"
                     )
                     raise tag_result
             results.extend(cast(list[dict[str, Any]], tag_result))
-        logger.info(f"Successfully fetched tags for {len(results)} Lambda functions")
+            success_count += 1
+        logger.info(f"Successfully fetched tags for {success_count} Lambda functions")
```
VenfiOranai, line 45: In case of `success_count` being different from `len(results)`, do we perhaps want to log a warning? Relevant to all other instances as well if yes

integrations/aws-v3/aws/core/exporters/organizations/account/actions.py
```diff
@@ -33,12 +33,14 @@ async def _execute(self, accounts: List[Dict[str, Any]]) -> List[Dict[str, Any]]
...
                     continue
                 elif is_resource_not_found_exception(result):
                     logger.warning(
                         f"Failed to list parents for account {accounts[idx]['Id']}: {result}"
                     )
                     results.append({"Parents": []})
+                    continue
                 else:
                     raise result
             results.append(cast(Dict[str, List[Dict[str, Any]]], result))
```
shalev007, line 46: can result herer be the Exception object?

shalev007 (approved):
LGTM, but is currently a bandage solution to the issue.
TBH we need to re-think how we add extra info as relying on order is a bad practice.
@oiadebayo @mk-armah

PR #3252, opened 2026-05-24
[Integration][AWS-V3] Fixed issues with state field on Organizations accounts

oiadebayo:
# Description

What - Added State field to AccountProperties and relaxed Pydantic validation from extra="forbid" to extra="ignore" in the aws-v3 Organizations account model.

Why - AWS Organizations API introduced State on accounts (Sept 2025) and is deprecating Status by Sept 9, 2026. The model only declared Status, so any response containing State caused a Pydantic validation error and blocked resync entirely.

How - Declared both Status and State as Optional on AccountProperties. Changed extra="forbid" to extra="ignore" to guard against future field additions. Updated tests to cover State-only, both-fields, and unknown-field scenarios.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


mk-armah (approved):
LGTM

shalev007 (approved):
LGTM

integrations/aws-v3/aws/core/exporters/organizations/account/models.py
```diff
@@ -11,12 +11,13 @@ class AccountProperties(BaseModel):
...
     Tags: List[Dict[str, Any]] = Field(default_factory=list)
-    Status: str = Field(default_factory=str)
+    Status: Optional[str] = None
+    State: Optional[str] = None
     JoinedTimestamp: Optional[datetime] = None
     JoinedMethod: Optional[str] = None
 
     class Config:
-        extra = "forbid"
+        extra = "ignore"
```
matan84: Please change to "allow"

matan84 (approved):

integrations/aws-v3/aws/core/exporters/organizations/account/models.py
mk-armah: @matan84 

I understand the benefit of allow since it could help us automatically capture new properties as they are introduced.
However, from the initial development phase, I noticed it also introduces a few concerns around consistency and model integrity:
- We currently enforce CamelCase across all fields in the model. Not all AWS actions/services return fields in CamelCase, and once we allow unrestricted propagation, we lose control over maintaining a consistent naming convention across the model.
- Metadata fields from individual actions would also flow into the model. Since multiple actions can expose the same top-level field names, some metadata fields may override each other.
- This becomes misleading because the metadata represents individual actions, not the aggregated model itself. Since models are composed from multiple action datasets, exposing action-level metadata at the model level can create ambiguity and incorrect assumptions about what the metadata actually describes.

I think we should be careful about preserving a controlled normalization layer between action payloads and the final model representation. Should we decide to allow new properties flow in seamlessly, which would be a good approach, a better way around this will involve more than setting extra=allow, we'd have to implement mechanisms to keep a consistent / predictable schema

PR #3253, opened 2026-05-24
[Integration][AWS-v3] Added support for GovCloud and China ARN partitions in Organizations strategy

oiadebayo:
# Description

What - Fixed ARN validation and role ARN construction in the Organizations strategy to accept all AWS partitions (`aws`, `aws-us-gov`, `aws-cn`).

Why - Self-hosted deployments in AWS GovCloud use `arn:aws-us-gov:iam::` ARNs, which were rejected by a hardcoded `startswith("arn:aws:iam::")` check, causing the integration to fail entirely during the Organizations health check.

How - Replaced the single-prefix check with a tuple of valid IAM ARN prefixes (`VALID_IAM_ARN_PREFIXES`). Also fixed `_build_role_arn` to use `details['partition']` from the parsed ARN instead of the hardcoded `aws` literal, so constructed member account role ARNs are correct in GovCloud and China regions.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


mk-armah (changes requested):
boto3 has an ArnParser class, do you think we can use that ?

oiadebayo:
> boto3 has an ArnParser class, do you think we can use that ?

The `ArnParser` class seems to be checking if the arn is the correct format only. It check if it follows `arn:partition:service:region:account:resource` but the partion itself is not validated. I parsed with it and validated after

integrations/aws-v3/aws/auth/strategies/organizations_strategy.py
```diff
@@ -54,19 +57,24 @@ def _get_organization_account_role_details(self) -> dict[str, str]:
...
+            arn_data = ArnParser().parse_arn(arn=organization_role_arn)
+        except InvalidArnException:
             raise AWSSessionError("account_role_arn must be a valid ARN")
 
-        arn_data = ArnParser().parse_arn(arn=organization_role_arn)
+        if (
+            arn_data["service"] != "iam"
+            or arn_data["partition"] not in VALID_AWS_PARTITIONS
+        ):
+            raise AWSSessionError("account_role_arn must be a valid ARN")
```
mk-armah, line 69: is this still needed ?

mk-armah (changes requested):
is VALID_AWS_PARTITIONS still needed

integrations/aws-v3/aws/auth/strategies/organizations_strategy.py
oiadebayo: Yes, the parser does not validate that the arn is an account_role_arn and  unless you are implying that the service validation only is enough, I think it is needed

mk-armah (approved):
LGTM

shalev007 (approved):
LGTM

matan84 (approved):

PR #3257, opened 2026-05-24
[Core] Add Independent Async Iterator Streaming Utility

emekanwaoma:
# Description

What - Added a shared Ocean utility for streaming independent async iterators concurrently while deferring iterator failures until surviving iterators finish.

Why - Independent sync streams should continue yielding retrievable data when one stream fails, while still surfacing errors at the end so incomplete syncs can avoid unsafe reconciliation deletes.

How - Added `stream_independent_async_iterators` with deferred error aggregation, cancellation-aware producer cleanup, bounded cleanup waiting, and focused tests for success, failure aggregation, backpressure, and cancellation paths.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### DIFF
--- a/port_ocean/utils/async_iterators.py
+++ b/port_ocean/utils/async_iterators.py
@@ -132,6 +132,7 @@ async def _consume_iterator(
     index: int,
     iterator: typing.AsyncIterable[T],
     queue: asyncio.Queue[_IteratorItem[T] | _IteratorFinished],
+    is_closing: asyncio.Event,
     context: str,
 ) -> None:
     error: Exception | None = None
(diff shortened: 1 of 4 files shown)


port_ocean/utils/async_iterators.py
```diff
@@ -107,3 +125,108 @@ async def main():
     async with semaphore:
         async for result in function():
             yield result
+
+
+async def _consume_iterator(
```
omby8888: the name is bit confusing as this function not consuming the queue, it produces to the queue.
maybe naming such as `_handle_iterator` or something will serve better here, wdyt?

omby8888 (approved):
LGTM

PR #3269, opened 2026-05-25
[Integration][Gitlab-v2] Implement api query params filtering for pipelines and jobs

melodyogonna:
# Description

What -

Why - Ingesting subset of pipeline and jobs in very large environments is a very slow affair, it requires fetching all pipelines and jobs, then using JQ to filter these down to the subset you want.

How - Introduce pipeline API query params. It may be that JQ may still be necessary in some cases, but then we'll be get subset of a subset rather than working on an entire dataset.

## Type of change

- [ x ] Non-breaking change (fix of existing functionality that will not change current behavior)

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/gitlab-v2/gitlab/clients/gitlab_client.py
```diff
@@ -390,12 +392,11 @@ async def get_projects_resource(
...
 
     async def _get_pipeline_jobs(
-        self, project_id: int | str
+        self, project_id: int | str, pipeline_params: Optional[dict[str, Any]] = None
     ) -> AsyncIterator[list[dict[str, Any]]]:
         # First get pipelines
         async for pipeline_batch in self.rest.get_paginated_project_resource(
-            str(project_id),
-            "pipelines",
+            str(project_id), "pipelines", params=pipeline_params
```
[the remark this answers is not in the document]
mk-armah: valid

mk-armah (changes requested):
Lets make sure we aren't mutating the api query params

mk-armah (approved):
LGTM

integrations/gitlab-v2/main.py
```diff
@@ -209,8 +209,13 @@ async def on_resync_pipelines(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
 
+        params = (
+            selector.api_query_params.generate_query_params()
+            if selector.api_query_params
+            else None
+        )
         async for pipelines_batch in client.get_projects_resource(
-            projects_batch, "pipelines"
+            projects_batch, "pipelines", params=params
         ):
```
raskinmaya: The new flow passes the same params dict through multiple async calls. Pagination logic usually mutates params (`per_page` insertion) in the REST layer. That means caller-owned params can be unexpectedly altered and reused with side effects

integrations/gitlab-v2/integration.py
```diff
@@ -40,6 +50,81 @@ class SearchQuery(BaseModel):
...
+        | None
+    ) = Field(
+        default=None,
+        title="Status",
+        description="Return only pipelines currently in the given execution status.",
+    )
+    source: str | None = Field(
+        default=None,
+        title="Source",
+        description="Return only pipelines triggered by the given source (e.g. push, schedule, web, merge_request_event, api).",
```
raskinmaya: `source` is a free str, and there is no cross-field validation. This pushes bad input directly to GitLab and relies on API errors instead of failing fast in config validation. Can we make the validation stricter?

integrations/gitlab-v2/main.py
```diff
@@ -209,8 +209,13 @@ async def on_resync_pipelines(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
             for project in projects_batch
         }
 
+        params = (
+            selector.api_query_params.generate_query_params()
+            if selector.api_query_params
+            else None
+        )
```
raskinmaya: Query params are recomputed inside the project-batch loop even though they are selector-level constants for the entire resync.

This refers [here](https://github.com/port-labs/ocean/pull/3269/changes#diff-0863b6f5aa347d648ed8ec779bf541736547be7c6d485d3d7a60bb9e520bfd1dR238) as well

integrations/gitlab-v2/tests/integration/test_integration.py
raskinmaya, line 1: No cover for failure modes (invalid date range, invalid source) or end-to-end selector plumbing at handler level

integrations/gitlab-v2/integration.py
```diff
@@ -40,6 +50,81 @@ class SearchQuery(BaseModel):
     )
 
 
+class PipelineQueryParams(QueryParamMixin):
```
raskinmaya: Why is [this](https://github.com/port-labs/ocean/pull/3269/changes#diff-5428dc0c8a5bea86af937a9bd6175c82ef351b1f0b0f8c11064e78fc695bb584R32) needed?
We could do:

```suggestion

class PipelineQueryParams(BaseModel):
    def generate_query_params(self) -> dict[str, Any]:
        return self.dict(exclude_none=True, exclude_unset=True)
```
& remove [this](https://github.com/port-labs/ocean/pull/3269/changes#diff-5428dc0c8a5bea86af937a9bd6175c82ef351b1f0b0f8c11064e78fc695bb584R32) class

integrations/gitlab-v2/main.py
melodyogonna: We traced and verified that this is not a problem. The pagination logic derives a new dictionary from the query params, and the "per_page" insertion has a guard on it.

integrations/gitlab-v2/integration.py
melodyogonna: Good point, updated.

integrations/gitlab-v2/tests/integration/test_integration.py
melodyogonna: These would fail selector validation?

integrations/gitlab-v2/main.py
melodyogonna: I see there is a valid concern, updated.

integrations/gitlab-v2/tests/integration/test_integration.py
raskinmaya: Partially. Invalid `source` and bad timestamp formats fail selector validation now, so tests for those are optional. Two things validation still doesn't cover: `updated_after > updated_before` (would reach GitLab), and handler plumbing (params actually forwarded to the client). Not blocking tho

raskinmaya (approved):

PR #3270, opened 2026-05-26
[Integration][Aikido] Align `scopeToTeam` selector with Port docs

dennis-quabynah-bilson:
# Description

Extended the scopeToTeam selector description to align with Port documentation.

## Type of change

- [x] Documentation (added/updated documentation)

## API Documentation

https://docs.port.io/build-your-software-catalog/sync-data-to-catalog/code-quality-security/aikido/?method=hosted&step=choose-method


integrations/aikido/integration.py
```diff
@@ -70,7 +70,7 @@ class IssueGroupSelector(Selector):
         default=False,
         alias="scopeToTeam",
         title="Scope To Team",
-        description="Whether to fetch issue groups scoped per team, enriched with team context",
+        description="Whether to fetch issue groups scoped per active team. When true, each issue group is enriched with <code>__team_id</code> and <code>__team_name</code> fields, which can be used to populate the <code>aikidoTeam</code> relation on the <code>aikidoIssueGroup</code> blueprint.",
```
mk-armah, line 73: I don't agree that we talk about blueprints here, descriptions here should be independent of that context, blueprint identifiers aren't static customers are free to name them however, so this pointer can be misleading ....

mk-armah (changes requested):

integrations/aikido/integration.py
dennis-quabynah-bilson: Agreed. I'll close this off. The docs should suffice

mk-armah (approved):
LGTM

PR #3286, opened 2026-05-28
[Core] Fix 401 from port-api cause resync to fail

Hadar-Vaknin:
# Description

Problem:
During long resyncs, Ocean keeps using a cached Port access token (last_token_object) for API calls such as POST /v1/entities/search.

Port may return 401 with "token is expired" while Ocean still treats the token as valid locally (last_token_object.expired is false), because local expiry is based on expiresIn and fetch time, which can differ from the JWT’s actual lifetime on Port’s side.

When that happened, retries did not recover:

Token refresh often did not run, because refresh was tied to the local expired flag.
Even when refresh ran, the new token was written to response.headers, but retries resend the original request, so the stale Authorization header was reused.
The resync then failed with a 401 on endpoints like entities/search.
## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


port_ocean/clients/port/retry_transport.py
```diff
@@ -11,41 +10,53 @@
...
             and self.port_client.auth.last_token_object is not None
-            and self.port_client.auth.last_token_object.expired
         )
 
+    async def _refresh_request_authorization(
+        self, request: httpx.Request
+    ) -> httpx.Request:
+        authorization = await self.port_client.auth.refresh_token()
+        headers = request.headers.copy()
+        headers["Authorization"] = authorization
```
omby8888:
```suggestion
        headers = request.headers.copy()
        headers["Authorization"] = await self.port_client.auth.refresh_token()
```

omby8888 (approved):

PR #3293, opened 2026-05-29
[Integration][Mend] Add new Mend.io SCA integration

hetsaliya-crestdata:
# Description

  What - Adds a new Mend.io integration that ingests Mend projects (`mend-project`) and SCA security findings
  (`sca-finding`) into Port.

  Why - Customers want their Mend SCA inventory and vulnerabilities mirrored in Port alongside other security-tool data.

  How -
  - Activation-key auth: base64 → JWT decode → `/api/v3.0/login` → `/api/v3.0/login/accessToken`, JWT held in-process and refreshed on TTL expiry or 401.
  - Cursor-based pagination against `/api/v3.0/orgs/{orgUuid}/projects/summaries` and `/api/v3.0/projects/{projectUuid}/dependencies/findings/security`.
  - Delta sync skips projects whose `lastScanTime` / `creationDate` did not change; `fullSyncIntervalMinutes` (default 720) periodically forces a full resync so finding-only changes (status updates, suppressions) are picked up.
  - Project list is cached for the duration of a single resync via `@cache_iterator_result()` so both kinds share one fetch.
  - `entityDeletionThreshold: 0` in `.port/resources/port-app-config.yml` prevents delta runs from deleting findings that weren't re-yielded.

  ## Type of change

  ### Core testing checklist

  - [ ] Docs PR link [here](#)

  ### Preflight checklist

  - [x] Handled rate limiting (via 401-retry on token expiry, ignored 502/503 transient errors)
  - [x] Handled pagination (cursor-based)
  - [ ] Support Multi account

  ## Screenshots

Port catalog view of mendProject and mendSCA blueprints populated from a test org.

mend-project-catalog
sca-finding-catalog

  ## API Documentation

  - Mend API: https://api-docs.mend.io/platform/3.0
  - Endpoints used:
    - `POST /api/v3.0/login`
    - `POST /api/v3.0/login/accessToken`
    - `POST /api/v3.0/orgs/{orgUuid}/projects/summaries`
    - `GET  /api/v3.0/projects/{projectUuid}/dependencies/findings/security`


integrations/mend/changelog/1.feature.md
omby8888, line 1: this shouldn't be pushed

integrations/mend/.port/spec.yaml
```diff
@@ -0,0 +1,20 @@
+title: Mend.io
+description: Mend SCA integration for Port Ocean
+icon: Mend
+features:
```
omby8888, line 5: We have currently a feature flag that you can set up in this file using `formMappingEnabled: true`, this will enable the mapping UX feature once this PR will be merged, if you need more details about it feel free to revert back :)

integrations/mend/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog - Ocean - mend
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- towncrier release notes start -->
```
omby8888, line 8: No changelog entry was created, pls add it manually or via towncrier

integrations/mend/tests/test_sample.py
omby8888, line 1: this is a default fallback to new integration, since u implemented tests this file can be removed

integrations/mend/Makefile
```diff
@@ -0,0 +1 @@
+/Users/het.saliya/Desktop/Github/ocean/port_ocean/cli/cookiecutter/../../../integrations/_infra/Makefile
```
omby8888:
```suggestion
include ../_infra/Makefile
```
did the cockiecutter did that? if so i'll open a bug for that

integrations/mend/integration.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class MendScaResourceConfig(ResourceConfig):
+    kind: Literal["sca-finding"]
+    selector: MendScaSelector
+
+
+class MendPortAppConfig(PortAppConfig):
+    resources: list[
+        MendProjectResourceConfig | MendScaResourceConfig | ResourceConfig
```
omby8888:
```suggestion
        MendProjectResourceConfig | MendScaResourceConfig
```
the default resource config can except and kind, while u want to limit acceptable kinds to `mend-project` and `sca-finding` only

integrations/mend/integration.py
omby8888, line 1: can u pls assign the title and description attributes to all of the fields here? in the CI we generate a JSON schema out of those models that grants us ability to display the user appropriate data in the UI

integrations/mend/mend/core/sync_state.py
```diff
@@ -0,0 +1,72 @@
+from datetime import datetime, timedelta
+from typing import Optional
+
+from loguru import logger
+
+# Process-local markers — survive across polling cycles within the same process.
+_in_process_last_sync: Optional[datetime] = None
```
omby8888, line 7: Ocean SaaS infra runs resyncs using `once` event listener (some customers also running with this method integrations on premise), meaning each resync would run on a different process. in that case we'll always run a full sync.
We don't have capability at the moment to save the state between processes, but since you allow to always running full sync i don't think that's a blocker for releasing this integration, wdyt?

integrations/mend/mend/clients/client.py
```diff
@@ -0,0 +1,133 @@
...
+        IgnoredError(
+            status=502,
+            message="Bad Gateway — transient upstream error",
+            type="BAD_GATEWAY",
+        ),
+        IgnoredError(
+            status=503,
+            message="Service Unavailable — transient server error",
+            type="SERVICE_UNAVAILABLE",
+        ),
```
omby8888: u might want to retry those instead of missing data by returning {}, in the current logic, the api might return 502/503 and you'll return empty projects set, which will cause the reconciliation to delete all existing project entities from Port, since your api didn't returned them in this resync

omby8888 (changes requested):

integrations/mend/changelog/1.feature.md
hetsaliya-crestdata: Removed — the file has been deleted.

integrations/mend/integration.py
hetsaliya-crestdata: Done — added title and description to all Field definitions: severity, kind (on both resource configs), selector, and resources.

integrations/mend/mend/core/sync_state.py
hetsaliya-crestdata: we only support the self-hosted Ocean plugin. In that setup, everything is managed through Docker, so I assume the value will persist across processes within that docker. Please let me know if I've misunderstood anything.

We also provide full synchronization support, so that approach will work as well. However, we introduced the concepts of full sync and partial sync to optimize REST API calls.

integrations/mend/mend/clients/client.py
hetsaliya-crestdata: Fixed — removed `502` and `503` from `_DEFAULT_IGNORED_ERRORS` and added exponential-backoff retry logic (up to 3 retries, 2^n  second delay) for transient errors. If all retries are exhausted the exception propagates, causing the sync to fail cleanly rather than silently returning an empty set.

Additionally, we've set `entityDeletionThreshold: 0` in `port-app-config.yml` as an extra safeguard — even if an empty response were returned, reconciliation would not delete existing entities.

integrations/mend/CHANGELOG.md
hetsaliya-crestdata: Manually added

integrations/mend/mend/core/sync_state.py
omby8888: even if self hosted, our helm chart has capability to run ocean as a cronjob, which will create new pod on each resync and that data will be lost.
also if u want the integration to only be ran as self hosted, u should add to the spec.yaml the following attribue:
```
saas:
  enabled: false
```

integrations/mend/mend/clients/client.py
omby8888: i'm not sure that what u want, it depends on how volatile those resources are in your system, because if they tend to become removed/irrelevant often, then you'd want to reflect that in Port, which will be denied by using the `entityDeletionThreshold: 0`

integrations/mend/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+
+## 0.1.0 (2026-06-03)
```
omby8888:
```suggestion
## 0.1.0-beta (2026-06-03)
```

integrations/mend/mend/clients/client.py
```diff
@@ -0,0 +1,136 @@
...
+            except httpx.HTTPStatusError as e:
+                status = e.response.status_code
+                if status == 401 and not retried_auth:
+                    logger.warning(
+                        f"401 at {url} — invalidating token and retrying once"
+                    )
+                    await self.authenticator.invalidate_token()
+                    retried_auth = True
+                    continue
+                if status in _TRANSIENT_STATUS_CODES and retry_count < _MAX_RETRIES:
```
omby8888: those transient statuses should be retried already by the http_async_client retry transport so no need for the handling in here imo.
regarding the 401 status handling, i recommend taking a look on how it was implemented in `integrations/github/github/clients/auth/retry_transport.py`, this inherits the `RetryTransport` and overrides the `before_retry_async` method, in which u can fetch a new token in and plan it in the returned httpx request that will be fired in the retry

integrations/mend/mend/core/sync_state.py
hetsaliya-crestdata: Okay got it. let it be because the things would also work as we are also providing the full sync as well.

integrations/mend/mend/clients/client.py
hetsaliya-crestdata: Yes, It would work for us as the resources are not removed.

omby8888 (approved):
Nice! Please make sure to validate the changelog date

hetsaliya-crestdata:
@omby8888  Thanks for providing the Approval. Can you please help on How can I trigger the required workflow  to merge the PR ?

omby8888:
i'll trigger those for u, seem like the `pyproject.toml` in this PR points to non updated version of `port_ocean`, which causes `validate-files` workflow to fail.

omby8888:
build is failing now because i think u bumped it manually rather than using poetry, pls run poetry lock + make install to validate the poetry.lock file

PR #3316, opened 2026-06-01
[Integration][Github] Add date and status filter selector for workflow-run kind

jasonnwakaeze-debug:
# Description

What - Add two optional selectors to the workflow-run kind: statuses, since and sinceDate.

Why -  The GitHub API returns up to 100 workflow runs per workflow by default (capped by max_runs). For orgs with active workflows, this means every resync fetches the same 100 most-recent runs regardless of relevance

How -

## Type of change

### Core testing checklist

- [X] Tested with at least 2 integrations from scratch
- [X] Tested with Kafka and Polling event listeners

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots
Screenshot 2026-06-08 at 17 13 48
Screenshot 2026-06-08 at 17 13 58

Shows the workflow run being called twice per workflow for statuses in_progress and completed

## API Documentation


integrations/github/integration.py
```diff
@@ -648,12 +648,51 @@ class GithubWorkflowConfig(ResourceConfig):
...
+            "requested",
+            "waiting",
+            "pending",
+        ]
+    ] = Field(
+        title="Status",
+        default=None,
+        description="Filter workflow runs by status or conclusion. When unset, all runs are returned.",
+    )
+    lookback_days: Optional[int] = Field(
```
melodyogonna: This same field is called "since" on the pull request selector, let's be consistent.

integrations/github/github/core/exporters/workflow_runs_exporter.py
```diff
@@ -42,7 +42,13 @@ async def get_paginated_resources[
         url = f"{self.client.base_url}/repos/{organization}/{repo_name}/actions/workflows/{options['workflow_id']}/runs"
         fetched_batch = 0
 
-        async for workflows in self.client.send_paginated_request(url):
+        params: dict[str, Any] = {}
+        if status := options.get("status"):
+            params["status"] = status
+        if created := options.get("created"):
+            params["created"] = created
```
melodyogonna: Let's have utility function generate these params.

melodyogonna (changes requested):
Some comments

integrations/github/github/core/exporters/workflow_runs_exporter.py
jasonnwakaeze-debug: do you envision more params being added, PR exporter also uses this format, breaking it out adds a layer of indirection for what amounts to two if-checks.

melodyogonna: Yeah, why not? I can see that workflow runs has a lot more filter query params than pull requests.

integrations/github/integration.py
```diff
@@ -648,12 +648,50 @@ class GithubWorkflowConfig(ResourceConfig):
...
+        ge=1,
+        description="Only fetch workflow runs created within the last N days.",
+    )
+
+    @property
+    def created_after(self) -> Optional[str]:
+        if self.since is None:
+            return None
+        cutoff = datetime.now(timezone.utc) - timedelta(days=self.since)
+        return f">={cutoff.strftime('%Y-%m-%dT%H:%M:%SZ')}"
```
omby8888: i think it would be better to make it since specific date rather than relative one

jasonnwakaeze-debug: but won't this lead to stale syncs @omby8888 a customer who sets 2024-01-01 is ingesting  an ever-growing backlog six months later.

melodyogonna (approved):
lgtm

integrations/github/integration.py
jasonnwakaeze-debug: We also already use this same pattern on the pull request selector (since: int) so it's consistent.

omby8888: ok yeah agree

omby8888 (approved):

integrations/github/integration.py
```diff
@@ -648,12 +648,50 @@ class GithubWorkflowConfig(ResourceConfig):
...
+            "neutral",
+            "skipped",
+            "stale",
+            "success",
+            "timed_out",
+            "in_progress",
+            "queued",
+            "requested",
+            "waiting",
+            "pending",
```
matan84: This should be a list and not a single value

integrations/github/integration.py
```diff
@@ -648,12 +648,50 @@ class GithubWorkflowConfig(ResourceConfig):
...
+        ]
+    ] = Field(
+        title="Status",
+        default=None,
```
matan84, line 673: Please set the current implementation's defaults as the defaults

integrations/github/integration.py
```diff
@@ -648,12 +648,50 @@ class GithubWorkflowConfig(ResourceConfig):
...
+        description="Filter workflow runs by status or conclusion. When unset, all runs are returned.",
+    )
+    since: Optional[int] = Field(
+        title="Lookback Days",
+        default=None,
```
matan84, line 678: Please put here 30 / what the default behavior in github is

matan84 (changes requested):

integrations/github/integration.py
jasonnwakaeze-debug: Actually, None is the current default behavior. no filter means all runs are returned, which preserves what existing customers already get.

integrations/github/integration.py
jasonnwakaeze-debug: The GitHub workflow runs API only accepts a single `status` query param supporting a list would require one API call per status value per workflow, which multiplies requests significantly at scale.

integrations/github/integration.py
jasonnwakaeze-debug: Setting a default here would break for existing customers, anyone who upgrades would silently lose runs older than 30 days on their next resync. There's no default set on the workflow runs api either

integrations/github/integration.py
matan84: The only other option we have is to have separate mapping blocks for the same data no? I think that customers would want more than 1 status for a single blueprint and this would restrict them to have X amount of the same mapping blocks for X statuses they want in Port. There should be a warning in the tooltip, and the recommendation should be to have as least as possible, but it should still be a list imo, lmk what you think

integrations/github/github/core/options.py
```diff
@@ -113,6 +113,8 @@ class ListWorkflowRunOptions(RepositoryIdentifier):
 
     workflow_id: Required[int]
     max_runs: Required[int]
+    status: NotRequired[Optional[str]]
```
matan84, line 116: shouldn't this be a list?

jasonnwakaeze-debug: Well, not on this side, ListWorkflowRunOptions maps directly to a single GitHub API request each instance is one HTTP call. status: Optional[str] reflects exactly what gets sent in that request.

integrations/github/github/webhook/webhook_processors/workflow_run/workflow_run_webhook_processor.py
```diff
@@ -32,7 +34,30 @@ async def handle_event(
...
             )
 
-        if action in WORKFLOW_DELETE_EVENTS:
+        selector_status = cast(
+            GithubWorkflowRunConfig, resource_config
+        ).selector.statuses
+        if selector_status:
+            if (
+                workflow_run["status"] not in selector_status
+                and workflow_run.get("conclusion") not in selector_status
```
mk-armah, line 42: what is the difference ?, is it possible for them to mismatch ? for instance workflow_run[status] != workflow_run.get(conclusion) ? which can cause True and False condition ?

mk-armah (dismissed):

integrations/github/github/webhook/webhook_processors/workflow_run/workflow_run_webhook_processor.py
jasonnwakaeze-debug: status tracks the run phase (queued/in_progress/completed) while conclusion tracks the outcome (success/failure/cancelled/etc.) — conclusion is always null until the run finishes. The selector accepts values from both fields, so we check both to support filtering by either phase or outcome. The and ensures we only delete if neither field matches. The condition predates the PR

matan84 (approved):

PR #3332, opened 2026-06-03
[integration][Github] Bug Github workflow runs api 500 errors cause sync failures

melodyogonna:
# Description

What - reducing page size on make paginated requests when 500 internal server errors are thrown

Why - To mitigate cases where the error seems to be caused by large page sizes.

How - cut the page size in half as long as it is still more than 25, adjust current_page to account for the change.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/github/github/clients/http/rest_client.py
```diff
@@ -44,27 +46,46 @@ async def send_paginated_request(
...
+
+                if current_page_size <= 25:
+                    raise
+
+                logger.warning(
+                    f"Got 500 for {method} {resource} at page={current_page} "
+                    f"per_page={current_page_size}; halving page size and retrying"
+                )
+                current_page = 2 * current_page - 1
+                current_page_size = current_page_size // 2
```
omby8888: since we not using any sort logic here, guessing that Github will miss some repos between those pages once this is shifting no? FYI @emekanwaoma

integrations/github/github/clients/http/rest_client.py
```diff
@@ -44,27 +46,46 @@ async def send_paginated_request(
...
+
+                yield items
+
+                if not (
+                    link_header := response.headers.get("Link")
+                ) or not self._get_next_link(link_header):
+                    break
+
+                current_page += 1
+            except httpx.HTTPStatusError as e:
```
omby8888: we'll only reach here after the 10 retries of the retry transport, to save some rate limits as we know Github will consecutively fail those request with fixed page size, how about we move that logic to the transport to save the extra 9 request each time that happens?

integrations/github/github/clients/http/rest_client.py
emekanwaoma: Actually I think the offset math is okay. When the per_page is halved, the page is also set to = 2×page−1, which keeps the same starting offset (e.g. failing on page 2 @ 100 → retry page 3 @ 50, both start at item 100). the gap between old/new pages are not skipped.

melodyogonna: Nah, the calculation accounts for what page we should be to avoid data loss.

integrations/github/github/clients/http/rest_client.py
melodyogonna: What if 500 internal server error is removed from retryable status codes instead? If I remember correctly it was added to counter this exact problem but it does seem to work for it.

integrations/github/github/clients/http/rest_client.py
omby8888: yeah the math checks out, I'm just wondering if this will act as expected due to the fact we do not use any sorting, just thinking with u if we worried that this fact will mess up the pages or should behave as expected

integrations/github/github/clients/http/rest_client.py
omby8888: i'm not sure that this is the only case where Github api returns 500, it could happen in non paginated request as well imo.
i still think moving this logic to the transport would be ideal

integrations/github/github/clients/http/rest_client.py
melodyogonna: Oh yeah, we did check that. Github list apis all have default sorting that is `createdAt desc`. The search api is different though, it is sorted by relevance by default.

integrations/github/github/clients/http/rest_client.py
emekanwaoma: I agree with Omri as well. I only have an issue with use case here. This self heal strategy should only be designed for paginated requests and for page_size pagination strategy. If we can have a way of making sure this the transport layer abstracts this for only that use case, then its okay. Otherwise, scoping it to Github rest-page-size pagination might be a good way to go. WDYT?

melodyogonna: Updated.

integrations/github/github/clients/auth/retry_transport.py
```diff
@@ -52,17 +64,58 @@ def __init__(
...
+        try:
+            per_page = int(url_params["per_page"])
+            # The first page request omits `page` (GitHub defaults to 1); later
+            # requests follow the Link header, which includes it explicitly.
+            page = int(url_params.get("page", 1))
+        except (KeyError, ValueError):
+            return request.url
+
+        if per_page <= MIN_PAGE_SIZE:
+            return request.url
```
omby8888, line 85: this features same exact logic as in `_page_reduction_exhausted`

melodyogonna: Does it? They're not really the same thing. `_page_reduction_exhausted` checks whether we've broken the page_size down to the smallest size we can (25) so returns true or false. The function you highlighted does not return true or false. I have extracted the one place where the logic is truly shared.

omby8888 (approved):
LGTM

PR #3350, opened 2026-06-05
[Integration][GitLab V2] Add support for deployment kind

eriport:
# Description

What -  Add deployment to the GitLab integration.

Why - Customers using GitLab CI/CD have no way to track deployments in Port. This PR unlocks DORA Deployment Frequency calculations and environment-level delivery visibility for GitLab-based workflows.

How -  Fan out per project using `GET /projects/:id/deployments`, enriching each deployment with the full `__project` object so all project fields are accessible in mappings. A pydantic validator on GitLabDeploymentSelector enforces the GitLab API constraint that `status=success` is required when a finished-at time window is set. `DeploymentWebhookProcessor` handles Deployment Hook events and updates both kinds from a single re-fetch. `deployment_events=True` added to GroupEvents so the integration subscribes on group webhook creation.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

Screenshot 2026-06-17 at 15 32 38
Screenshot 2026-06-17 at 15 32 46
Screenshot 2026-06-15 at 07 26 22
Screenshot 2026-06-15 at 07 26 50

## API Documentation

https://docs.gitlab.com/api/deployments/#list-all-project-deployments
https://docs.gitlab.com/api/deployments/#retrieve-a-deployment


integrations/gitlab-v2/main.py
```diff
@@ -79,6 +81,9 @@
 from gitlab.webhook.webhook_processors.branch_webhook_processor import (
     BranchWebhookProcessor,
 )
+from gitlab.webhook.webhook_processors.deployment_webhook_processor import (
+    DeploymentWebhookProcessor,
+)
```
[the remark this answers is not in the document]
eriport: ignore this, precommit didn't include this files at commit, hence first commit flag this for inclusion

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
...
+
+        if self.query_params:
+            params.update(self.query_params.build_query_params())
+        if self.finished_after or self.finished_before:
+            params["order_by"] = (
+                "id"  # default on GitLab API, tracking this explicitly here for clarity
+            )
+            params["sort"] = (
+                "asc"  # default on GitLab API, tracking this for same reason
+            )
```
raskinmaya: GitLab’s docs state that when using `finished_after`/`finished_before`, you must set `order_by=finished_at` and `status=success` ([ref](https://docs.gitlab.com/api/deployments/#list-project-deployments)). This branch sets `order_by=id` and even tests that as correct.

That can return wrong or incomplete results for the exact use case this feature targets (incremental “what finished recently” syncs). The tests also encode incorrect behavior and will block a fix

integrations/gitlab-v2/main.py
```diff
@@ -573,6 +578,54 @@ async def on_resync_folders(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+    ):
+        logger.info(
+            f"Processing batch of {len(projects_batch)} projects for deployment statuses"
+        )
+        async for deployments_batch in client.get_deployments(
+            projects_batch,
+            max_concurrent=DEFAULT_MAX_CONCURRENT,
+            params=selector.build_query_params() or None,
+        ):
+            yield deployments_batch
```
raskinmaya: Both handlers call `client.get_deployments()` with the same underlying GitLab list endpoint. GitLab has no separate deployment-status list API (unlike GitHub’s `/deployments/{id}/statuses`).

If a customer configures both kinds (likely, given GitHub parity), every resync doubles GitLab API calls across all projects. On large instances this is a real rate-limit and runtime problem.

Cleaner options would probably be:
- One resync path shared by both kinds, or
- Document clearly that deployment-status is a view/mapping of the same object, not a separate sync, or
- Drop deployment-status until GitLab exposes status history (if product allows)

integrations/gitlab-v2/gitlab/webhook/webhook_processors/deployment_webhook_processor.py
```diff
@@ -0,0 +1,56 @@
...
+            return WebhookEventRawResults(
+                updated_raw_results=[], deleted_raw_results=[]
+            )
+
+        deployment["__project"] = project
+
+        return WebhookEventRawResults(
+            updated_raw_results=[deployment],
+            deleted_raw_results=[],
+        )
```
raskinmaya, line 131: `handle_event` accepts `resource_config` but never uses it. Compare with `pipeline_webhook_processor.py` (lines 33-47), which honors `includeOnlyActiveProjects`.

This means that the impact for GitLab customers:
- Webhooks from archived projects still sync when `includeOnlyActiveProjects: true`
- Webhooks ignore `apiQueryParams` (`environment`, `status`, `updatedAfter`)
- Every status transition triggers a full refetch even when the deployment would be filtered out on resync

Either apply selector filtering in the processor (like `pipeline`) or document that webhooks are unfiltered - silently diverging from resync is worse

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
...
+        description="Selector for the GitLab deployment resource.",
+    )
+
+
+class GitLabDeploymentStatusResourceConfig(ResourceConfig):
+    kind: Literal[ObjectKind.DEPLOYMENT_STATUS] = Field(
+        title="GitLab Deployment Status",
+        description=(
+            "GitLab deployment status resource kind, representing the current status of a CI/CD deployment."
+        ),
```
raskinmaya: GitHub `deployment-status` is a history of status transitions per deployment. GitLab exposes a single deployment record with a status field - no status timeline API.

Users porting GitHub blueprints may expect multiple status entities per deployment. They’ll get one entity that overwrites on each transition. Worth a clear description in the resource config/docs, or naming that doesn’t imply GitHub semantics

integrations/gitlab-v2/gitlab/clients/gitlab_client.py
```diff
@@ -955,3 +955,53 @@ async def _resolve_file_references(
                 data[index] = await self._resolve_file_references(item, project_id, ref)
 
         return data
+
+    async def _enrich_deployment_with_full_project_context(
```
raskinmaya: New code duplicates the fan-out/semaphore/stream pattern but attaches the full project instead of path-only. That’s ~45 lines where a parameter on `_enrich_project_resources`/`get_projects_resource_with_enrichment` (e.g. `project_enrichment: "path" | "full"`) would suffice and stay DRY

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
...
+
+            if self.finished_after:
+                params["finished_after"] = self.finished_after.isoformat()
+            if self.finished_before:
+                params["finished_before"] = self.finished_before.isoformat()
+
+        return params
+
+
+class GitLabDeploymentStatusSelector(Selector):
```
raskinmaya: `GitLabDeploymentStatusSelector` is essentially `GitLabDeploymentSelector` minus `finished_after`/`finished_before`. Inheritance or a shared base (like `PipelineSelector(ProjectSelector)`) would cut ~40 lines and keep behavior in sync.

Same for query param building: `GitLabDeploymentQueryParams.build_query_params()` hand-rolls what `PipelineQueryParams.generate_query_params()` does via `dict(exclude_none=True, exclude_unset=True)` (lines 149-150)

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
...
+        return params
+
+
+class GitLabDeploymentSelector(Selector):
+    include_only_active_projects: bool = Field(
+        default=True,
+        alias="includeOnlyActiveProjects",
+        title="Include Only Active Projects",
+        description="If true, only include deployments from active projects (non-archived). If false, include deployments from all projects regardless of their active status.",
+    )
```
raskinmaya: Deployment selectors default to True. `ProjectSelector` defaults to `None` (no filter, lines 169-174). `release`/`tag`/`pipeline` inherit that.

Customers get different default behavior for deployments vs releases without opting in. Align with `ProjectSelector` unless there’s a strong product reason

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
...
+
+    def build_query_params(self) -> dict[str, Any]:
+        params: dict[str, Any] = {}
+        if self.environment:
+            params["environment"] = self.environment
+        if self.status:
+            params["status"] = self.status.value
+        if self.updated_after:
+            params["updated_after"] = self.updated_after
+        return params
```
raskinmaya: GitLab list deployments supports `updated_before`. `PipelineQueryParams` includes it (lines 130-135). Omitting it limits incremental sync windows and is inconsistent within the same file

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
...
+    )
+    updated_after: str | None = Field(
+        default=None,
+        alias="updatedAfter",
+        title="Updated After",
+        description=(
+            "Return deployments updated after this datetime (ISO 8601). "
+            "Recommended for incremental resyncs to avoid fetching the full history every cycle."
+        ),
+    )
```
raskinmaya, line 613: `updated_after` is a free-form string. `finished_after`/`finished_before` use `datetime` with `.isoformat()` and no timezone guarantee. Pipeline fields use `ISO_8601_DATETIME_REGEX` (lines 27-28, 124-146). Naive datetimes → API errors or wrong filters on self-hosted GitLab with stricter parsing

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
...
+    updated_after: str | None = Field(
+        default=None,
+        alias="updatedAfter",
+        title="Updated After",
+        description=(
+            "Return deployments updated after this datetime (ISO 8601). "
+            "Recommended for incremental resyncs to avoid fetching the full history every cycle."
+        ),
```
raskinmaya, line 600: Without `updatedAfter`, resync walks all deployments for all projects. GitLab deployment lists can be large and are ordered by default id ascending, which can expensive on first sync and on every periodic resync

integrations/gitlab-v2/gitlab/clients/gitlab_client.py
```diff
@@ -955,3 +955,53 @@ async def _resolve_file_references(
...
+        self,
+        project: dict[str, Any],
+        deployments_iterator: AsyncIterator[list[dict[str, Any]]],
+    ) -> AsyncIterator[list[dict[str, Any]]]:
+        """Enrich deployment data with full project data, including languages and search query results."""
+        async for deployments_batch in deployments_iterator:
+            yield [
+                {**deployment, "__project": project} for deployment in deployments_batch
```
raskinmaya: Resync attaches the full project from `get_projects`. Webhook attaches the sparse project object from the hook payload (missing archived, languages, etc.). JQ mappings can differ between resync and webhook for the same entity which can be surprising for users. Pipeline refretches the full pipeline object; consider `get_project(project_id)` for webhook enrichment if full context matters

integrations/gitlab-v2/gitlab/clients/gitlab_client.py
```diff
@@ -955,3 +955,53 @@ async def _resolve_file_references(
                 data[index] = await self._resolve_file_references(item, project_id, ref)
 
         return data
+
+    async def _enrich_deployment_with_full_project_context(
+        self,
+        project: dict[str, Any],
+        deployments_iterator: AsyncIterator[list[dict[str, Any]]],
+    ) -> AsyncIterator[list[dict[str, Any]]]:
+        """Enrich deployment data with full project data, including languages and search query results."""
```
raskinmaya: This mentions “languages and search query results” but the method only sets `__project`

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
     )
 
 
+class GitLabDeploymentQueryParams(BaseModel):
```
raskinmaya: I don't want to be nitpicking, but this PR creates naming inconsistency. `GitLabDeployment*` vs `GitlabMergeRequest*`, `GitlabIssue*`, `GitlabPortAppConfig`, etc. Let's have it aligned and rename the different classes to `GitlabDeployment*`

integrations/gitlab-v2/integration.py
```diff
@@ -569,6 +569,162 @@ class BranchResourceConfig(ResourceConfig):
...
+        default=None,
+        alias="updatedAfter",
+        title="Updated After",
+        description=(
+            "Return deployments updated after this datetime (ISO 8601). "
+            "Recommended for incremental resyncs to avoid fetching the full history every cycle."
+        ),
+    )
+
+    def build_query_params(self) -> dict[str, Any]:
```
raskinmaya: Again, `build_queryparams()` vs existing `generate_query_params()`. Same responsibility, different name. Ideally I would want them merged somehow

integrations/gitlab-v2/gitlab/webhook/webhook_processors/deployment_webhook_processor.py
```diff
@@ -0,0 +1,56 @@
...
+    events = ["deployment"]
+    hooks = ["Deployment Hook"]
+
+    async def get_matching_kinds(self, event: WebhookEvent) -> list[str]:
+        return [ObjectKind.DEPLOYMENT, ObjectKind.DEPLOYMENT_STATUS]
+
+    async def handle_event(
+        self,
+        payload: EventPayload,
+        resource_config: ResourceConfig,
```
raskinmaya, line 32: Unused parameter

integrations/gitlab-v2/tests/clients/test_gitlab_client.py
```diff
@@ -1943,3 +1943,164 @@ async def test_get_groups_no_min_access_level_passes_empty_params(
...
+            return_value=async_mock_generator([mock_deployments]),
+        ):
+            results = []
+            async for batch in client.get_deployments([project]):
+                results.extend(batch)
+
+        injected_project = results[0]["__project"]
+        assert injected_project["name"] == "Test Project"
+        assert injected_project["web_url"] == "https://gitlab.example.com/group/project"
+        assert injected_project["visibility"] == "private"
```
raskinmaya: Two tests assert full-project injection so one of them is redundant IMO

integrations/gitlab-v2/tests/webhook/webhook_processors/test_deployment_webhook_processor.py
```diff
@@ -0,0 +1,124 @@
...
+            project_id=project_id,
+            deployment_id=deployment_id,
+        )
+        assert len(result.updated_raw_results) == 1
+        enriched = result.updated_raw_results[0]
+        assert enriched["id"] == deployment_id
+        assert enriched["__project"] == project
+        assert enriched["__project"]["path_with_namespace"] == "root/test-project"
+        assert not result.deleted_raw_results
+
```
raskinmaya: Looking at the code [here](https://github.com/port-labs/ocean/pull/3350/changes#diff-f62ae6fbef62f9e0a49f17b0e76946fc46b56a79e23453ce322474e2b8415625R105) (lines 105-124), there's an overlap on `__project` enrichment

integrations/gitlab-v2/gitlab/webhook/webhook_processors/deployment_webhook_processor.py
raskinmaya: `updatedAfter`/`updatedBefore` are still not honored on webhooks. They only affect resync via `generate_query_params()`. Is that intentional? Even if it is, it’s still worth calling out explicitly so customers don’t assume live events respect the same incremental window as resync.

You can either:
Add post-fetch filtering on `deployment["updated_at"]` when those params are set, or if this is by design, document in the selector/docs/both that `updatedAfter`/`updatedBefore` are resync-only filters

integrations/gitlab-v2/integration.py
raskinmaya: Still valid as an operational concern IMO. `updatedAfter` is optional and the field description recommends it for incremental syncs, but there’s no default - first/periodic resync without it will walk the full deployment list per project.

I don’t think we need a schema default (that could surprise existing configs), but worth making this more visible for operators - e.g. a note in the integration docs/blueprint example that production GitLab setups should set `apiQueryParams.updatedAfter` for deployment resyncs

raskinmaya (approved):

PR #3351, opened 2026-06-05
[Integration][GitHub] Enhance closed pull requests with sinceDate selector with absolute date

dennis-quabynah-bilson:
# Description

- Added a `sinceDate` field to the `GithubPullRequestSelector` to allow fetching closed pull requests since a specified date.
- Updated related logic in the `RestPullRequestExporter` and `GraphQLPullRequestExporter` to utilize the new date filter.
- Adjusted the `max_results` handling to accommodate the new selector.
- Updated changelog to reflect the new feature in version 5.6.0.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/github/github/core/options.py
```diff
@@ -50,8 +50,9 @@ class ListPullRequestOptions(RepositoryIdentifier):
     """Options for listing pull requests."""
 
     states: Required[list[str]]
-    max_results: Required[int]
+    max_results: Required[Optional[int]]
     updated_after: Required[datetime]
+    use_close_date: NotRequired[bool]
```
omby8888: The context here is confusing as we use the `updated_after` date while filtering by the `closedAt` date. i think these dates should be 2 independent options and then we won't need the `use_close_date` which is bit ambiguous

integrations/github/integration.py
```diff
@@ -326,19 +326,25 @@ class GithubPullRequestSelector(RepoSearchSelector):
...
-        description="Max number of merged pull requests. Note: large numbers may cause rate limits. Merged PRs are only retrieved when 'closed' is selected in the state selector.",
+        description="Max number of merged pull requests. Defaults to 100 for the days lookback; with sinceDate set, leave empty to fetch all closed PRs back to the cutoff. Large numbers may cause rate limits. Merged PRs are only retrieved when 'closed' is selected in the state selector.",
     )
     since: int = Field(
         title="Closed PRs Lookback Days",
         default=60,
         ge=1,
         description="Numbers of days back for closed pull requests.",
     )
+    since_date: Optional[datetime] = Field(
```
omby8888: is this selector referring to the date that a PR was closed, or enable fetching all PRs that opened since this date? 
from the implementation seem like the first one but in any case let's emphasis that in the description

integrations/github/github/core/exporters/pull_request_exporter/core.py
```diff
@@ -133,31 +139,61 @@ async def _fetch_closed_pull_requests(
...
-
-            logger.info(
-                f"[Rest] Fetched closed pull requests batch of {batch_count} from {repo_name} from {organization} "
-                f"(total so far: {total_count + batch_count}/{max_results})"
-            )
-
-            enriched_batch = [
-                enrich_with_organization(
-                    enrich_with_repository(pr, repo_name), organization
+            if use_close_date:
```
omby8888: too many nesting here it's kinda hard to track the logic here, also imo this duplicates bellow

omby8888 (changes requested):

integrations/github/integration.py
omby8888: let the type be string as we get that through api so this can't be a datetime object, also wdyt on `closedSinceDate` as the selector name instead? ideally i would want that naming in all selectors related to closed PRs, or even object selector that unifies all closed PRs settings but we don't want to break api

integrations/github/github/core/exporters/pull_request_exporter/utils.py
```diff
@@ -1,17 +1,108 @@
...
+            total_count += len(in_window)
+
+            # closed_at <= updated_at and results are sorted by updated_at desc, so once
+            # updated_at falls before the cutoff no later page can hold an in-window PR.
+            if len(filter_prs_by_date(page, updated_at_field, cutoff)) < len(page):
+                logger.info(
+                    f"{log_prefix} Reached cutoff for closed pull requests of "
+                    f"{repo_name} from {organization}; stopping."
+                )
+                break
```
omby8888: instead of reiterating the whole page, we can just check if the last item's updated_at is before the cutoff, no?

integrations/github/github/core/exporters/pull_request_exporter/utils.py
```diff
@@ -1,17 +1,108 @@
...
+            )
+            yield [
+                enrich(pr) for pr in filter_prs_by_date(raw, updated_at_field, cutoff)
+            ]
+            total_count += raw_count
+
+    logger.info(
+        f"{log_prefix} Fetched total of {total_count} closed pull requests "
+        f"from {organization}/{repo_name}"
+    )
```
omby8888: i think this case can be treated exactly like the first case (if closed_mode), can we generalize this logic a bit? i think the function should just get cutoff and a date field to examine and it should be same logic from there

integrations/github/github/core/exporters/pull_request_exporter/core.py
```diff
@@ -281,12 +272,13 @@ async def _fetch_closed_pull_requests(
...
-        max_results: int,
-        updated_after: datetime,
+        max_results: Optional[int],
+        updated_after: Optional[datetime],
+        closed_after: Optional[datetime],
         pr_gql_options: PullRequestGraphQLOptions,
     ) -> ASYNC_GENERATOR_RESYNC_TYPE:
-
         repo_name = repo["name"]
+        order_by_field = "UPDATED_AT" if closed_after is not None else "CREATED_AT"
```
omby8888: why in the rest exporter we not setting that as well?

dennis-quabynah-bilson: Great point. I have added it

integrations/github/github/core/exporters/pull_request_exporter/core.py
```diff
@@ -296,48 +292,34 @@ async def _fetch_closed_pull_requests(
...
-
-            total_count += batch_count
-
-        logger.info(
-            f"[GraphQL] Fetched total of {total_count} closed pull requests from {organization}/{repo_name}"
-        )
+        async for batch in paginate_closed_pull_requests(
+            self.client.send_paginated_request(
+                generate_list_pull_requests_gql(
+                    pr_gql_options, order_by_field="UPDATED_AT"
```
omby8888: use defined const instead of magic string, but now it always be the sorting, no?

integrations/github/github/core/exporters/pull_request_exporter/core.py
```diff
@@ -296,48 +292,34 @@ async def _fetch_closed_pull_requests(
...
+                ),
+                variables,
+            ),
+            enrich=enrich,
+            max_results=max_results,
+            cutoff=cutoff,
+            include_field=include_field,
+            stop_field="updatedAt",
```
omby8888: use defined const instead of magic string

integrations/github/github/core/exporters/pull_request_exporter/core.py
```diff
@@ -296,48 +292,34 @@ async def _fetch_closed_pull_requests(
...
-                break
-
-            remaining = max_results - total_count
-            if remaining <= 0:
-                break
-
-            limited_batch = pr_nodes[:remaining]
-            batch_count = len(limited_batch)
+        cutoff, include_field = (
+            (closed_after, "closedAt")
```
omby8888: use defined const instead of magic string

integrations/github/github/core/exporters/pull_request_exporter/core.py
```diff
@@ -120,48 +128,34 @@ async def _fetch_closed_pull_requests(
...
-        logger.info(
-            f"[Rest] Fetched total of {total_count} closed pull requests from {organization}/{repo_name}"
-        )
+        async for batch in paginate_closed_pull_requests(
+            self.client.send_paginated_request(endpoint, params),
+            enrich=enrich,
+            max_results=max_results,
+            cutoff=cutoff,
+            include_field=include_field,
+            stop_field="updated_at",
```
omby8888: isn't it updatedAt? and use defined const instead of magic string

dennis-quabynah-bilson: For REST it is `updated_at`

integrations/github/github/core/exporters/pull_request_exporter/core.py
omby8888: i dont see it

dennis-quabynah-bilson: It's set on the REST side via the list-pulls query params  `sort=updated&direction=desc` ([per the docs](https://docs.github.com/en/rest/pulls/pulls)), which is the REST equivalent of the GraphQL UPDATED_AT desc ordering. I've now pulled those into named constants (REST_SORT_BY_UPDATED / REST_SORT_DIRECTION_DESC) next to the params dict in _fetch_closed_pull_requests, so it's explicit and parallel to PR_ORDER_BY_UPDATED_AT on the GraphQL side.

integrations/github/github/helpers/gql_queries.py
```diff
@@ -1,5 +1,9 @@
 from github.core.options import PullRequestGraphQLOptions
 
+# GraphQL pullRequests orderBy field values.
+PR_ORDER_BY_CREATED_AT = "CREATED_AT"
+PR_ORDER_BY_UPDATED_AT = "UPDATED_AT"
```
omby8888: pls move those to where the others defined in the core exporter

omby8888 (approved):

PR #3357, opened 2026-06-08
[Integration][Cursor] Cursor Ocean Integration

PeyGis:
# Description

What - Added a new Port-hosted Cursor ocean integration

Why - Improve installation experience and move away from the custom ocean installation option for Cursor

How - Pulled data from Cursor Admin and AI Code Tracking API

## Type of change

### Core testing checklist

- [ ] Docs PR link [[here](https://github.com/port-labs/port-docs/pull/4169)](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## API Documentation

Admin API = https://cursor.com/docs/account/teams/admin-api
Analytics API = https://cursor.com/docs/account/teams/analytics-api#model-usage


integrations/cursor/.port/resources/port-app-config.yml
```diff
@@ -0,0 +1,108 @@
...
+      endDate: 0d
+    port:
+      entity:
+        mappings:
+          identifier: ((.timestamp + "-" + .userEmail + "-" + .model + "-" + .kind) | gsub("[^A-Za-z0-9@_.+:\\\\/='-]"; "_"))
+          title: (.userEmail + " " + .model + " " + .kind)
+          blueprint: '"cursorUsageEvent"'
+          properties:
+            timestamp: (.timestamp | tonumber / 1000 | todate)
+            model: .model
```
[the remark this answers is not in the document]
PeyGis: timestamp is string so this comment does not apply.

integrations/cursor/clients/cursor_client.py
mk-armah, line 1: all resource specific functions are to be implemented on the exporters themselves, client only has http request functions , majorly send_api_request, and send_paginated_request

integrations/cursor/integration.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+    selector: CursorRelativeDateSelector
+
+
+class CursorUserModelUsageResourceConfig(ResourceConfig):
+    kind: Literal["cursor-user-model-usage"] = Field(
+        description="Cursor per-user model usage resource kind",
+        title="Cursor User Model Usage",
+    )
+    selector: CursorRelativeDateSelector
```
mk-armah: add description to the selector for selector propation enhacement

integrations/cursor/integration.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class CursorDailyUsageResourceConfig(ResourceConfig):
+    kind: Literal["cursor-daily-usage"] = Field(
+        description="Cursor daily usage data resource kind",
+        title="Cursor Daily Usage",
+    )
+    selector: CursorRelativeDateSelector
```
mk-armah: add description to the selector for selector propation enhacement

integrations/cursor/integration.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class CursorUsageEventResourceConfig(ResourceConfig):
+    kind: Literal["cursor-usage-event"] = Field(
+        description="Cursor usage events data resource kind",
+        title="Cursor Usage Events",
+    )
+    selector: CursorRelativeDateSelector
```
mk-armah: add description to the selector for selector propation enhacement

integrations/cursor/core/exporters/usage_events_exporter.py
```diff
@@ -0,0 +1,12 @@
...
+from core.exporters.abstract_exporter import AbstractCursorExporter
+from core.options import ListCursorAdminOptions
+
+
+class CursorUsageEventsExporter(AbstractCursorExporter):
+    async def get_paginated_resources(
+        self, options: ListCursorAdminOptions
+    ) -> ASYNC_GENERATOR_RESYNC_TYPE:
+        async for batch in self.client.get_usage_events_data(options):
+            yield batch
```
mk-armah: implement get_usage_events_data here directly, instead of the on the client

mk-armah (changes requested):

integrations/cursor/main.py
```diff
@@ -0,0 +1,112 @@
...
+        app_config = (
+            await ocean.integration.port_app_config_handler.get_port_app_config()
+        )
+        configured_kinds = {resource.kind for resource in app_config.resources}
+    except Exception as exc:
+        logger.warning(
+            f"Could not load Port app config for startup validation, "
+            f"falling back to validating the Analytics API only: {exc}"
+        )
+        configured_kinds = set()
```
shalev007: nit: I don't think this code is reachable since it'll probably fail in the `validate_integration_runtime` before the integration `on_start` is called

integrations/cursor/clients/cursor_client.py
```diff
@@ -0,0 +1,210 @@
...
+                response.raise_for_status()
+
+            wait_seconds = self._backoff_seconds * (2**attempt)
+            logger.warning(
+                f"Cursor API rate limit hit on {method} {path}; retrying in {wait_seconds:.1f}s "
+                f"(attempt {attempt + 1}/{self._max_retries})"
+            )
+            await asyncio.sleep(wait_seconds)
+
+        raise RuntimeError("Exceeded retry attempts while calling Cursor API")
```
shalev007: nit: This entire function is mostly a duplicate of the `RetryTransport` that it should already include in the `_client` object

integrations/cursor/main.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+_ANALYTICS_KINDS = {
+    ObjectKind.CURSOR_TEAM_MODEL_USAGE.value,
+    ObjectKind.CURSOR_USER_MODEL_USAGE.value,
+}
+_ADMIN_KINDS = {
+    ObjectKind.CURSOR_DAILY_USAGE.value,
+    ObjectKind.CURSOR_USAGE_EVENT.value,
+}
```
shalev007, line 87: nit: I don't think StrEnums in python 3.12+ need the `.value` they can just be `ObjectKind.CURSOR_DAILY_USAGE` for example

integrations/cursor/core/options_builder.py
```diff
@@ -0,0 +1,43 @@
+from datetime import datetime, timedelta, timezone
+
+from core.options import ListCursorAdminOptions, ListCursorAnalyticsOptions
+
+DEFAULT_PAGE = 1
+# Cursor's paginated endpoints accept large page sizes
+DEFAULT_PAGE_SIZE = 500
```
shalev007: we should make it configurable! as this can affect sync duration and error scoping can you add it to the `spec.yaml`

integrations/cursor/clients/cursor_client.py
```diff
@@ -0,0 +1,210 @@
...
+            pagination = payload.get("pagination", {}) or {}
+            total_pages = pagination.get("totalPages") or pagination.get("numPages")
+            progress = f"{page}/{total_pages}" if total_pages else f"{page}"
+            logger.info(f"Cursor pagination {path}: fetched page {progress}")
+
+            yield payload
+
+            if not bool(pagination.get("hasNextPage", False)):
+                break
+            page += 1
```
shalev007: What will happen if there's an error during the listing of pages and the retries are exhausted?
Currently it'll break the listing and the integration will never reach `total_pages` this is not resilient enough, the integration should try to avoid data loss whenever it can

integrations/cursor/clients/cursor_client.py
```diff
@@ -0,0 +1,210 @@
...
+        for attempt in range(self._max_retries + 1):
+            async with limiter:
+                response = await self._client.request(
+                    method=method,
+                    url=url,
+                    headers=self._headers,
+                    params=params,
+                    json=json_body,
+                    timeout=self._request_timeout_seconds,
+                )
```
shalev007: currently the client is already running a retry internally so the retries are actually `attempt * base client retries` this can exhaust the rate limit, if you want to use your own retry mechanism, I suggest you limit the client's one

shalev007 (changes requested):
Overall looks really good, simple understandable code good separation of scoped modules.
I marked the nitpicks with prefix of nit but the other are important questions need to be answered before approving :)

mk-armah (changes requested):
Please model/organize the tests directory as the implementation's

integrations/cursor/clients/cursor_client.py
```diff
@@ -0,0 +1,210 @@
...
+        api_host: str,
+        api_key: str,
+        request_timeout_seconds: int = 30,
+        max_retries: int = 5,
+        backoff_seconds: float = 1.0,
+    ) -> None:
+        if not api_host or not str(api_host).startswith(("http://", "https://")):
+            raise ValueError("cursor_api_host must be a valid http(s) URL")
+        if not api_key:
+            raise ValueError("cursor_api_key must be provided")
```
mk-armah: I think we can avoid these runtime checks due to the following reasons

1.   api_host is `type: url` and is required in the spec.yaml
2. api_key is required

Hence this will pre-validate

mk-armah (changes requested):

integrations/cursor/clients/cursor_client.py
PeyGis: I've moved the retry logic to use ocean's own transport layer so this should be handles by the framework's default mechanism.

shalev007: yeah but what if the the retry is exhausted currently its still breaking the pagination, 1 error = no more pagination

integrations/cursor/clients/cursor_client.py
```diff
@@ -0,0 +1,215 @@
...
+                continue
+            logger.info(f"Cursor pagination {path}: fetched page {page}/{total_pages}")
+            yield payload
+
+        if failed_pages:
+            raise RuntimeError(
+                f"Cursor pagination {path}: {len(failed_pages)}/{total_pages} page(s) "
+                f"failed after retries ({failed_pages}). Raising so Ocean skips its "
+                "delete phase and preserves the affected entities until the next resync."
+            )
```
shalev007: nit: I'd suggest use a custom error to show whats going on

integrations/cursor/clients/cursor_client.py
```diff
@@ -0,0 +1,215 @@
...
+            except httpx.HTTPError as exc:
+                failed_pages.append(page)
+                logger.error(
+                    f"Cursor pagination {path}: page {page}/{total_pages} failed "
+                    f"after retries; skipping it and continuing with the remaining "
+                    f"pages: {exc}"
+                )
+                continue
+            logger.info(f"Cursor pagination {path}: fetched page {page}/{total_pages}")
+            yield payload
```
shalev007: TBH this is a bit much and not what I was hoping for, can we simplify this?
1. I'm not sure why we need the `total_pages` if we can simply walk through `hasNextPage` also if theres an error we probably wont get both
2. in the second option (`hasNextPage` as fallback) if theres an error we still block pagination if the request fails anyway
3. we can simply use the `hasNextPage` with `failed_pages` collecting should be simpler as well

shalev007 (approved):
Approve to unblock but please fix the pagination request

integrations/cursor/clients/cursor_client.py
shalev007: Oh I just understood if the request fails there's no nextPage, so there's no pagination, OK my bad

integrations/cursor/core/exporters/user_model_usage_exporter.py
```diff
@@ -0,0 +1,32 @@
...
+        async for payload in self.client.send_paginated_request(
+            "GET",
+            USER_MODEL_USAGE_PATH,
+            params={
+                "startDate": options["startDate"],
+                "endDate": options["endDate"],
+            },
+        ):
+            # `data` maps each user email to a list of their daily breakdowns;
+            # flatten it into per-user-day records carrying the owning email.
```
mk-armah:
```suggestion
```

mk-armah (approved):
LGTM

PR #3364, opened 2026-06-09
[Integration][Datadog] Add support for exporting from multiple orgs

melodyogonna:
# Description

What - Adds support to enable exporting from multiple orgs.

Why - For customers with a lot of orgs but don't want to install the exporter N times where N = number of orgs.

How - Add a `datadogCredentialMap` which is a json data with credentials from all orgs we want to ingest.

## Type of change

- 

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/datadog/.port/spec.yaml
```diff
@@ -36,6 +36,11 @@ configurations:
     required: true
     type: string
     sensitive: true
+  - name: datadogCredentialMap
```
omby8888, line 39: we don't want to show this in the UI atm (because we don't have support for nested secret input), so pls open a PR to the monorepo and exclude this field in `packages/common-consts/src/consts.ts:580`

integrations/datadog/.port/spec.yaml
```diff
@@ -55,6 +60,12 @@ configurations:
     type: string
     required: false
     sensitive: true
+  - name: isMultiOrg
```
omby8888: does this really nessacery?

integrations/datadog/.port/spec.yaml
omby8888, line 39: each org can be on a different domain, so we should support different datadog api hosts

integrations/datadog/client_manager.py
```diff
@@ -1,11 +1,112 @@
...
+        try:
+            api_key = credentials["datadogApiKey"]
+            app_key = credentials["datadogApplicationKey"]
+        except (KeyError, TypeError) as e:
+            raise IntegrationMissingConfigError(
+                f"datadogCredentialMap entry for org '{org_id}' must include "
+                "'datadogApiKey' and 'datadogApplicationKey'"
+            ) from e
+        yield DatadogClient(
+            config["datadog_base_url"],
```
omby8888: we should allow overriding this per org (for FedRamp orgs), this would be the default in case no override provided

integrations/datadog/initialize_client.py
```diff
@@ -1,11 +1,112 @@
...
 from port_ocean.context.ocean import ocean
 
+from datadog.exceptions import IntegrationMissingConfigError
+
+
+def get_credential_map(config: dict[str, Any]) -> dict[str, Any]:
+    credential_map_string = config.get("datadog_credential_map")
+    if credential_map_string:
+        try:
+            parsed = json.loads(credential_map_string)
```
omby8888: having pydantic model for this config would help parsing be much cleaner and will save the afterwards checks

integrations/datadog/client_manager.py
```diff
@@ -1,11 +1,112 @@
...
+
+    def __init__(self, config: dict[str, Any]) -> None:
+        self.is_multi_org: bool = config["is_multi_org"]
+        if self.is_multi_org:
+            self._clients = list(init_client_for_multi_org(config))
+        else:
+            self._clients = [init_client_single_org(config)]
+        self._clients_by_org: dict[str | None, DatadogClient] = {
+            client.org_id: client for client in self._clients
+        }
```
omby8888: this can be inferred solely from the existence/absense of `datadog_credential_map`

integrations/datadog/initialize_client.py
```diff
@@ -1,11 +1,112 @@
...
+    events rather than rebuilt on every call.
+    """
+
+    def __init__(self, config: dict[str, Any]) -> None:
+        self.is_multi_org: bool = config["is_multi_org"]
+        if self.is_multi_org:
+            self._clients = list(init_client_for_multi_org(config))
+        else:
+            self._clients = [init_client_single_org(config)]
+        self._clients_by_org: dict[str | None, DatadogClient] = {
```
omby8888: this is duplicated data (the `_clients` list is redundant), holding only the clients map can also become a list by inspecting it's values

integrations/datadog/main.py
```diff
@@ -43,131 +45,153 @@
...
+        TeamExporter(client).get_paginated_resources(
+            ListTeamOptions.from_resource_config(
+                cast(TeamResourceConfig, event.resource_config)
+            )
         )
-    ):
+        for client in init_client()
+    )
+
+    async for teams in stream_async_iterators_tasks(*tasks):
```
omby8888:
```suggestion
    async for teams in stream_independent_async_iterators(*tasks):
```
there's this new utility @emekanwaoma wrote which i think would be useful here, it will perform best effort to fetch all data and will fail only at the end, this will prevent situation of single toxic org affecting all the rest. this is relevant for all other kinds as well

integrations/datadog/initialize_client.py
```diff
@@ -1,11 +1,112 @@
...
+
+def get_client_manager() -> DatadogClientManager:
+    """Return the process-wide client manager, building it on first use."""
+    global _client_manager
+    if _client_manager is None:
+        _client_manager = DatadogClientManager(ocean.integration_config)
+    return _client_manager
+
+
+def init_client() -> Iterator[DatadogClient]:
```
omby8888: no need for this anymore, replace all calls to `get_client_manager().clients` instead

integrations/datadog/datadog/webhook/webhook_processors/audit_trails/base_processor.py
```diff
@@ -85,9 +89,15 @@ async def handle_event(
                     deleted_raw_results=[deleted] if deleted else [],
                 )
 
+        client = self._get_client_for_payload(payload)
+        if client is None:
+            return WebhookEventRawResults(
```
omby8888, line 96: lets add warn log here for visibility

integrations/datadog/datadog/webhook/webhook_processors/base_webhook_processor.py
```diff
@@ -16,7 +16,21 @@ class BaseWebhookProcessor(AbstractWebhookProcessor):
...
+        return str(org_id) if org_id is not None else None
+
+    def _get_client_for_payload(self, payload: EventPayload) -> DatadogClient | None:
+        """Select the Datadog client responsible for the org that emitted the event.
+
+        Delegates to the client manager: single-org installs always use their sole
+        client; multi-org installs match the payload's org id against the configured
+        credential map, returning None (so callers skip the event) when nothing matches.
+        """
+        return self._client_manager.get_client_for_org(self._extract_org_id(payload))
```
omby8888: this would work different beween audit trails and webhook events, in the audit trails, an event will look like this:
```
{
    "date": "2026-06-08T07:51:48.423Z",
    "attributes": {
        "evt": {...},
        "auth_method": "SESSION",
        "org": {
            "name": "DPN | Port",
            "uuid": "dbb758e9-396e-11ef-800c-bea43249b5f6"
        },
        "usr": {...},
        "http": {...},
        "action": "deleted",
        "asset": {...},
        "uuid": "cEtjd09tR0ZLOA==",
        "network": {...},
        "timestamp": "2026-06-08T07:51:48.423Z",
        "status": "info"
    },
    "_id": "AZ6mN2PHAABwS2N3T21HRks4",
    "source": "audit",
    "message": "omrib@getport.io deleted team try",
    "status": "info"
}
```
so in that case we would need the `event.org.uuid` (i checked and it's equivalent to the `public_id` we get from the list orgs api).

Regarding webhook events, after digging around i saw that the org_id returned when using the `$ORG_ID` is an internal id that we can get from the api thus we cannot use it, so for webhook events i would wire that app using the org name (this means we should save that)

integrations/datadog/datadog/overrides.py
```diff
@@ -241,6 +241,13 @@ class RoleResourceConfig(ResourceConfig):
     )
 
 
+class OrgResourceConfig(ResourceConfig):
+    kind: Literal["org"] = Field(
```
omby8888:
```suggestion
    kind: Literal["organization"] = Field(
```

integrations/datadog/client_manager.py
omby8888, line 1: let's rename that file into `client_manager.py`

integrations/datadog/initialize_client.py
```diff
@@ -1,11 +1,112 @@
...
+        config["datadog_access_token"],
     )
+
+
+def init_client_for_multi_org(config: dict[str, Any]) -> Iterator[DatadogClient]:
+    credential_map = get_credential_map(config)
+    for org_id, credentials in credential_map.items():
+        try:
+            api_key = credentials["datadogApiKey"]
+            app_key = credentials["datadogApplicationKey"]
```
omby8888: ideally we should validate here if the credentials fits to the provided org_id, or there could be some mixups we want to avoid

omby8888 (changes requested):

integrations/datadog/.port/spec.yaml
melodyogonna: @omby8888 this is actually a string. I learned from an issue we had in Kafka where setting the map as an object/dict prevented it from being used in Github actions, we had to introduce another mapping config to compensate.

integrations/datadog/.port/spec.yaml
melodyogonna: Maybe not, I'll probably remove it before merging.

integrations/datadog/main.py
melodyogonna: Oh interesting, I'll take a look at that.

integrations/datadog/.port/spec.yaml
melodyogonna: I looked into this, I thought datadog uses just one base api, interesting.

integrations/datadog/initialize_client.py
melodyogonna: How do you think we should do that, make api calls for each org?

integrations/datadog/datadog/webhook/webhook_processors/base_webhook_processor.py
melodyogonna: I'm trying to see if using a public org id as the key would work, since it is already being returned by audit trail events, we just need a way to make it also return on monitor webhooks. I'm experimenting with adding org uuid in the webhook header set up.

melodyogonna: That didn't work out very well, I couldn't find the public id in the UI, which means using that would require a lot of complicated setup. I switched the credentialMap to be a json array

integrations/datadog/.port/spec.yaml
omby8888: i know it's string, but that would mean that api keys will be exposed as plain text in the UI and on api requests, which is not what we want

melodyogonna: Even though it is marked as sensitive? Wow, that's problematic.

integrations/datadog/datadog/webhook/webhook_processors/audit_trails/base_processor.py
```diff
@@ -85,12 +88,11 @@ async def handle_event(
                     deleted_raw_results=[deleted] if deleted else [],
                 )
 
-        resource = None
-        try:
-            resource = await self._fetch_resource(event, resource_config)
-        except HTTPStatusError as e:
-            if e.response.status_code == 404:
-                logger.warning("Resource returned 404, skipping event", error=str(e))
+        org_name = event.attributes.org.name if event.attributes.org else None
```
omby8888: in audit trails we should use the uuid rather the name, it's available in attributes.org.uuid

integrations/datadog/datadog/types.py
```diff
@@ -1,6 +1,29 @@
...
+
+
+class DatadogCredentialMap(BaseModel):
+    """The datadogCredentialMap JSON: a list of per-org credentials.
+
+    A list (not a dict keyed by org) because org names aren't unique — several
+    entries may share a name, and routing tries each candidate in turn.
+    """
+
+    __root__: List[OrgCredentials]
```
omby8888, line 30: I do think it should be saved as a dict by the org id, the customer should provide the public org ids and we'll valudate those against the provided credentials.
in case of the monitor webhooks where we dont get the id in the payload - we should iterate that dict's value and find the matching orgs by name

integrations/datadog/datadog/webhook/webhook_processors/base_webhook_processor.py
```diff
@@ -47,9 +87,12 @@ async def authenticate(
             logger.warning("Webhook authentication failed: invalid secret")
         return is_valid
 
+    def _get_auth_header_value(self, headers: dict[str, Any]) -> str | None:
+        return self._get_header_value(headers, PORT_AUTH_HEADER_NAME)
+
     @staticmethod
-    def _get_auth_header_value(headers: dict[str, Any]) -> str | None:
-        expected_header_name = PORT_AUTH_HEADER_NAME.lower()
+    def _get_header_value(headers: dict[str, Any], name: str) -> str | None:
```
omby8888: where does this being used?

integrations/datadog/datadog/webhook/webhook_processors/base_webhook_processor.py
```diff
@@ -11,12 +12,51 @@
...
+
 
 class BaseWebhookProcessor(AbstractWebhookProcessor):
 
     def __init__(self, event: WebhookEvent) -> None:
         super().__init__(event)
-        self.client: DatadogClient = init_client()
+        self._client_manager = get_client_manager()
+
+    async def _fetch_from_matching_client(
```
omby8888: why the abstraction here is required? imo the logic should be simply pull the correct client/s from the client manager in the base audit trail and base/each monitor events

integrations/datadog/datadog/webhook/types.py
```diff
@@ -30,10 +30,16 @@ class UrlDetails(BaseModel):
     url_details: UrlDetails
 
 
+class AuditTrailOrg(BaseModel):
+    name: str
+    uuid: Optional[str] = None
```
omby8888: this is not optional, it always exists in audit trails

integrations/datadog/datadog/webhook/types.py
```diff
@@ -30,10 +30,16 @@ class UrlDetails(BaseModel):
...
+
+
 class AuditTrailAttributes(BaseModel):
     evt: AuditTrailEvt
     action: str
     asset: AuditTrailAsset
+    org: Optional[AuditTrailOrg] = None
```
omby8888: not optional

integrations/datadog/.port/spec.yaml
omby8888: ohhh i missed that attribute, your'e right it will be hidden, in any case this is poor UX because we can't support such structure in th UI atm so best to hide that for ocean Saas

integrations/datadog/datadog/types.py
melodyogonna: Yeah I thought so as well, then I went looking for the public id and couldn't find it anywhere in the UI: [See comment](https://github.com/port-labs/ocean/pull/3364#discussion_r3431077150). Either the customer has to make API calls to get it or we have to.

integrations/datadog/datadog/webhook/webhook_processors/base_webhook_processor.py
melodyogonna: This is an artifact of when I tried using org id as the credential key, I was attempting to stamp the org_id in the webhook header during registration, and then retrieve it when webhooks are triggered so both monitor and audit trail events wouldn't have to rely on org name.

omby8888: ok but since it's not being used pls revert this change

melodyogonna: Yeah.

integrations/datadog/datadog/types.py
omby8888: i think it should be shown in the UI only for admins, according to Datadog at least it should, org i even prefer the customer would have provided that.
the name is not very helpful to us in that context

integrations/datadog/initialize_client.py
omby8888: yeah in the on_start maybe, if we do that we can get for each org it's name as well which will be useful in monitor webhook events

integrations/datadog/client_manager.py
```diff
@@ -0,0 +1,168 @@
...
+
+        Clients whose credentials are rejected (or whose org can't be resolved) are
+        dropped, so resyncs and live events only run against valid keys. No-op for
+        single-org installs, which keep their sole client untouched.
+        """
+        if not self.is_multi_org:
+            return
+
+        valid: list[DatadogClient] = []
+        for i in range(51):
```
omby8888: why 51?

integrations/datadog/client_manager.py
```diff
@@ -0,0 +1,168 @@
...
+            logger.warning(
+                f"Datadog rejected credentials for '{client.api_url}' "
+                f"({e.response.status_code})"
+            )
+            return None
+
+        orgs = result.get("orgs") or []
+        if not orgs:
+            return None
+        org = orgs[0]
```
omby8888: this logic should sit in the org exporter

integrations/datadog/client_manager.py
```diff
@@ -0,0 +1,168 @@
...
+            by_id[client.org_id].append(client)
+            by_name[self._normalize_org_name(client.org_name)].append(client)
+        self._clients_by_org_id = by_id
+        self._clients_by_org_name = by_name
+
+    @staticmethod
+    def _normalize_org_name(org_name: str | None) -> str | None:
+        return org_name.lower() if org_name is not None else None
+
+    def get_clients_by_org_id(self, org_id: str | None) -> list[DatadogClient]:
```
omby8888: this should result only single client not several as public id is unique in datadog

integrations/datadog/client_manager.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+
+    def __init__(self, config: dict[str, Any]) -> None:
+        self.is_multi_org: bool = bool(config.get("datadog_credential_map"))
+        self._clients: list[DatadogClient] = (
+            list(init_client_for_multi_org(config))
+            if self.is_multi_org
+            else [init_client_single_org(config)]
+        )
+        self._clients_by_org_id: dict[str | None, list[DatadogClient]] = {}
```
omby8888:
```suggestion
        self._client_by_org_id: dict[str, DatadogClient] = {}
```

integrations/datadog/client_manager.py
```diff
@@ -0,0 +1,168 @@
...
+        self._clients_by_org_name: dict[str | None, list[DatadogClient]] = {}
```
omby8888:
```suggestion
        self._clients_by_org_name: dict[str, list[DatadogClient]] = {}
```

integrations/datadog/client_manager.py
```diff
@@ -0,0 +1,168 @@
...
+        if not public_id or not name:
+            return None
+        return public_id, name
+
+    def _build_indexes(self) -> None:
+        by_id: dict[str | None, list[DatadogClient]] = defaultdict(list)
+        by_name: dict[str | None, list[DatadogClient]] = defaultdict(list)
+        for client in self._clients:
+            by_id[client.org_id].append(client)
+            by_name[self._normalize_org_name(client.org_name)].append(client)
```
omby8888: by id should use `=` instead of append, and if this is a new key won't append throw exception?

integrations/datadog/main.py
```diff
@@ -41,159 +49,173 @@
...
+MAX_CONCURRENT_CLIENTS = 100
+
+
+async def _resync_across_orgs(
+    build_iterator: Callable[[DatadogClient], ASYNC_GENERATOR_RESYNC_TYPE],
+    context: str,
+) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    semaphore = asyncio.Semaphore(MAX_CONCURRENT_CLIENTS)
+    tasks = (
+        semaphore_async_iterator(semaphore, functools.partial(build_iterator, client))
```
omby8888: very nice!!

integrations/datadog/client_manager.py
melodyogonna: Trying to prevent cyclic dependency

integrations/datadog/client_manager.py
melodyogonna: Lol I mistakenly pushed this. Nvm, deleted, it was a load testing mechanism to test semaphore.

integrations/datadog/datadog/core/exporters/org_exporter.py
```diff
@@ -0,0 +1,50 @@
...
+        Docs: https://docs.datadoghq.com/api/latest/organizations/#list-your-managed-organizations
+        """
+        url = f"{self.client.api_url}/api/v1/org"
+        result = await self.client.send_api_request(url)
+        orgs = result.get("orgs") or []
+        if orgs:
+            yield orgs
+
+    @cache_coroutine_result()
+    async def get_resource(  # type: ignore[override]
```
omby8888: why is this annotation needed? pls revert that if possible

omby8888 (approved):

PR #3387, opened 2026-06-16
[Integration][Azure-DevOps] Add bounded concurrency to prevent 429s during webhook setup and pipeline-run resync

jasonnwakaeze-debug:
# Description

What -

- Added bounded concurrency to `defaultTeam` project enrichment and `includeMembers` team enrichment to prevent unbounded concurrent API calls that exhausted the ADO TSTU budget.
- Raised `max_backoff_wait` from 60s to 300s so `Retry-After` and `X-RateLimit-Reset` headers are honored up to ADO's 5-minute rolling window instead of being clipped prematurely.
- Added a shared throttle circuit-breaker to `AzureDevOpsRateLimiter`: a `ReadTimeout` now pauses all subsequent ADO requests for 300s, preventing continued retries against an already-throttled.

Why -  On each resync, the unbounded gather fired 50 concurrent get_single_project calls per page, immediately exhausting the shared TSTU budget.   
The previous max_backoff_wait default (60s) also clipped waits shorter than ADO's 5-minute window, causing the integration to re-enter the same rate limit window and thrash repeatedly.

How -

## Type of change

### Core testing checklist

- [X] Tested with at least 2 integrations from scratch

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots
Screenshot 2026-06-16 at 15 53 22

## API Documentation


matan84 (approved):

integrations/azure-devops/azure_devops/client/base_client.py
```diff
@@ -69,6 +74,8 @@ async def send_request(
                 )
                 raise e
         except httpx.HTTPError as e:
+            if isinstance(e, ReadTimeout):
```
emekanwaoma: ocean retry transport may still retry connect timeouts with normal backoff.
do we need to add the shared 5‑min hold for connect failures?

emekanwaoma (changes requested):
tiny comment

integrations/azure-devops/azure_devops/client/base_client.py
jasonnwakaeze-debug: I thought about that, but ConnectTimeout typically indicates transient network failure, which recovers in seconds. The retry transport's backoff handles that better than a flat 5-min hold, which would massively overshoot for a network blip. 

The 5-min signal is calibrated specifically to ADO's TSTU reset window. ADO's documented throttling behavior is to hold HTTP requests server-side (ReadTimeout), not sure it's evidence enough to refuse TCP connections

emekanwaoma (approved):
lgtm

PR #3388, opened 2026-06-16
[Core] Refactor PollingEventListener to streamline resync logic

yaelibarg:
# Description

The polling & kafka event listeners should only look into the resync requests (for apply mapping purposes)


port_ocean/core/event_listener/polling.py
```diff
@@ -141,6 +130,7 @@ async def resync() -> None:
                     )
                     if should_resync:
                         resync_reason = "Detected integration resync request"
+                        resync_request_updated_at = resync_request.get("updatedAt", "")
                 except Exception as error:
```
Hadar-Vaknin: Talked about it f2f but noting it here anyway, u are still depending on integration updates in:
- should_resync_from_resync_request
- the polling loop still querying the integration for it's updated at

Hadar-Vaknin (commented):
I see that the tests on polling are failing, u might need to make adaption there

Hadar-Vaknin (approved):

PR #3393, opened 2026-06-17
[Core] Implement redis consumer for the Live Events new infra

Hadar-Vaknin:
# Description

Adds Redis stream consumption for live events in Ocean core, as part of the new live-events infrastructure where incoming webhooks are received by an ingestion service and queued to an integration-specific Redis stream.

When the LIVE_EVENTS_REDIS_STREAM_ENABLED organization feature flag is on, LiveEventsProcessorManager consumes events from Redis instead of the in-process HTTP → LocalQueue path. Existing webhook processors are unchanged — events are deserialized, routed by webhookPath, and passed through the same processing pipeline.

Desgin doc: https://www.notion.so/Live-Events-Infra-Improvement-Design-Document-37c9be8ff7e38183bb92f18d005eaf9b?source=copy_link

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,228 @@
...
+
+        while self._running:
+            try:
+                response = await self._redis.xreadgroup(
+                    groupname=self._consumer_group,
+                    consumername=self._consumer_name,
+                    streams={self._stream_key: ">"},
+                    count=10,
+                    block=self._settings.block_ms,
+                )
```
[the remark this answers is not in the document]
Hadar-Vaknin: This will be implemented in another PR: a worker that periodically consume the group PEL.

port_ocean/core/handlers/webhook/processor_manager.py
```diff
@@ -67,85 +73,110 @@ async def start_processing_event_messages(self) -> None:
...
+            stream_key = resolve_live_events_stream_key_from_base_url(base_url)
+            self._redis_consumer = RedisStreamConsumer(
+                redis_settings=config.live_events.redis,
+                stream_key=stream_key,
+                on_message=self._on_redis_stream_message,
+                consumer_group=config.live_events.redis.consumer_group,
+                registered_paths=set(self._processors_classes.keys()),
+            )
+            await self._redis_consumer.start()
+            logger.info(
```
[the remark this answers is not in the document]
Hadar-Vaknin: Expected behavior.

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,228 @@
...
+
+    async def _ensure_consumer_group(self) -> None:
+        assert self._redis is not None
+        try:
+            await self._redis.xgroup_create(
+                self._stream_key,
+                self._consumer_group,
+                id="0",
+                mkstream=True,
+            )
```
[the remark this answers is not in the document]
Hadar-Vaknin: Fixed

port_ocean/core/handlers/webhook/processor_manager.py
```diff
@@ -67,85 +73,110 @@ async def start_processing_event_messages(self) -> None:
...
+                "Live events Redis stream consumer enabled",
+                stream_key=stream_key,
+            )
+            return
```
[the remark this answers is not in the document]
Hadar-Vaknin: When the flag is on, we are consuming from the Redis stream, not the in-memory queue.

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+
+    def _resolve_consumer_group(self) -> str:
+        integration = ocean.config.integration
+        return f"{integration.type}.{integration.identifier}"
+
+    def _decode_base64_pem(self, value: str) -> str:
+        return base64.b64decode(value).decode()
+
+    def _materialize_client_tls_files(self) -> tuple[str | None, str | None]:
```
raskinmaya, line 67: If `_redis_client_kwargs()` is called multiple times (e.g. reconnection), this function creates new temp files without cleaning up the old ones, causing a resource leak. Let's call `_cleanup_tls_files()` at the start of this function and I think we should be good

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+                    for message_id, fields in messages:
+                        await self._handle_message(message_id, fields)
+            except asyncio.CancelledError:
+                break
+            except Exception as error:
+                logger.exception(
+                    "Unexpected error in Redis stream read loop",
+                    error=str(error),
+                )
+                await asyncio.sleep(1)
```
raskinmaya: This hardcoded `sleep` on error is too simplistic IMO. Repeated failures will hammer Redis with reonnection attemps. Better to use an exponential backoff with jitter

port_ocean/core/handlers/webhook/processor_manager.py
```diff
@@ -67,85 +73,110 @@ async def start_processing_event_messages(self) -> None:
...
-                Tuple[ResourceConfig | None, AbstractWebhookProcessor, int | None]
-            ] = []
-            try:
-                event = await queue.get()
-                with logger.contextualize(
-                    worker=worker_id,
-                    webhook_path=path,
-                    trace_id=event.trace_id,
+    async def _on_redis_stream_message(self, path: str, event: WebhookEvent) -> None:
+        await self._process_webhook_event(path, 0, event)
```
raskinmaya: We always pass `worker_id=0`, losing the parallelism that queue-based processing has with `config.event_workers_count`. This could bottleneck high-volume integrations. If the intent is single-threaded processing, document why. If parallelism is desired, spawn multiple coroutines from the read loop or use a worker pool

port_ocean/core/handlers/webhook/webhook_event.py
```diff
@@ -1,16 +1,26 @@
...
+EventHeaders = dict[str, str]
+
+
+class WebhookOriginalRequest(Protocol):
+    """Minimal request interface for webhook signature verification."""
+
+    @property
+    def headers(self) -> Mapping[str, Any]: ...
+
+    async def body(self) -> bytes: ...
```
raskinmaya, line 23: I'm sorry for the megila but 2 points here:

- `WebhookOriginalRequest` is defined here as the minimal interface for signature verification, but `_WebhookRequestAdapter` lives in `redis_stream_consumer`. That splits the contract from its only non-request implementation and makes the redis path harder to discover.
I'd consider moving `_WebhookRequestAdapter` into `webhook_event` and importing it from the consumer. That keeps webhook request semantics in one place and makes the dependency direction clearer: consumer -> webhook model, not the other way around

- Another thing is that types are slightly inconsistent:
`Protocol`: `headers` -> `Mapping[str, Any]`
`Adapter`: `headers` -> `dict[str, str]`
Runtime-wise this is fine, but it’s sloppy at the type level. Since `EventHeaders` is already `dict[str, str]` in this module, I'd align the `Protocol` to that (or reuse `EventHeaders`) so the adapter structurally satisfies the interface without `Any`

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+    @staticmethod
+    def _normalize_webhook_path(webhook_path: str) -> str:
+        """Map ingestion HTTP paths to integration processor paths.
+
+        Integrations register processors relative to the integration router
+        (e.g. ``/webhook``), while ingestion forwards the public URL suffix
+        (e.g. ``integration/webhook``).
+        """
+        path = f"/{webhook_path.strip('/')}"
+        integration_marker = "/integration/"
```
raskinmaya: Can you make this a module-level constant? Something like:


```
_INTEGRATION_PATH_PREFIX = "/integration/"
```

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+    async def _read_loop(self) -> None:
+        assert self._redis is not None
+
+        while self._is_running:
+            try:
+                response = await self._redis.xreadgroup(
+                    groupname=self._consumer_group,
+                    consumername=self._consumer_name,
+                    streams={self._stream_key: ">"},
+                    count=10,
```
raskinmaya: Let's make this configurable via `LiveEventsRedisSettings` for tuning

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+    @classmethod
+    def _parse_raw_json_to_dict(
+        cls, raw_value: str | None, field_name: str
+    ) -> dict[str, Any]:
+        if raw_value is None:
+            return {}
+
+        parsed: Any = json.loads(raw_value)
+        while isinstance(parsed, str):
+            parsed = json.loads(parsed)
```
raskinmaya: This `while` loop handles deep JSON encoding - why would the upstream ever double-encode?
This should be limited explicitly to 1-2 levels of unwrapping with a clear warning if more is detected

port_ocean/config/settings.py
```diff
@@ -112,6 +112,34 @@ class ActionsProcessorSettings(BaseOceanModel, extra=Extra.allow):
     workers_count: int = Field(default=1)
 
 
+class LiveEventsRedisSettings(BaseOceanModel, extra=Extra.allow):
+    url: str = Field(default="redis://localhost:6379")
+    username: str | None = None
+    password: str | None = Field(default=None, sensitive=True)
+    enable_tls: bool = False
```
raskinmaya, line 121: There is no validation that `enable_tls=True` aligns with `redis://` URL or vice versa. A user could set `enable_tls=False` with a `rediss://` URL, leading to confusing behavior. You can add a Pydantic validator, something like:

```
@validator("enable_tls", always=True)
def validate_tls_consistency(cls, v, values):
    url = values.get("url", "")
    if url.startswith("rediss://") and not v:
        raise ValueError("enable_tls should be True when using rediss:// URL")
```

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+
+        if not isinstance(parsed, dict):
+            raise InvalidLiveEventsRedisStreamFieldError(field_name)
+        return parsed
+
+    @classmethod
+    def _parse_json_object_field(
```
raskinmaya: This is just composing `_get_field` with `_parse_raw_json_to_dict`. It's only called once (for `headers`). I think you should either inline ir or justify it's existence. For `payload`, you're calling `_get_field` separately then `_parse_raw_json_to_dict`

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+        if integration_marker in path:
+            path = f"/{path.split(integration_marker, 1)[1]}"
+        return path
+
+    @staticmethod
+    def _get_field(fields: dict[str, str], field_name: str) -> str | None:
+        for key, value in fields.items():
+            if key.lower() == field_name.lower():
+                return value
+        return None
```
raskinmaya: You iterate over all remaining keys after finding a match because there's no early return (there is, but Python doesn't optimize the iterator). A little more Pythonic:

```suggestion
        return next(
            (v for k, v in fields.items() if k.lower() == field_name.lower()),
            None
        )
```

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+
+        cert_path, key_path = self._materialize_client_tls_files()
+        if cert_path and key_path:
+            kwargs["ssl_certfile"] = cert_path
+            kwargs["ssl_keyfile"] = key_path
+
+        return kwargs
+
+    async def start(self) -> None:
+        self._redis = Redis.from_url(self._settings.url, **self._redis_client_kwargs())
```
raskinmaya, line 116: This uses default connection pooling. Clients might want to configure `max_connection`

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -0,0 +1,291 @@
...
+                self._consumer_group,
+                id="$",
+                mkstream=True,
+            )
+        except ResponseError as error:
+            if "BUSYGROUP" not in str(error):
+                raise
+
+    async def _read_loop(self) -> None:
+        assert self._redis is not None
```
raskinmaya: Using `assert` for runtime checks in production code is an anti-pattern. Please use explicit checks. This is also relevant [here](https://github.com/port-labs/ocean/pull/3393/changes#diff-faac178ebcde4a5a31430fa331a8499d116430a0c09771ae025f44f6f836c2daR157)

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: Deleted it, xreadgroup already blocks for block_ms (default 1000ms), so we never hit line 195.

port_ocean/core/handlers/webhook/processor_manager.py
Hadar-Vaknin: Parallelism is acquired with multiple LE pods with the new infrastructure that was intended to replace the whole existing logic (in-memory queue + workers) as written in the doc.
I'll add a comment that explains why it's always 0.

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/config/settings.py
Hadar-Vaknin: I actually did it on purpose to make the enable_tls flag the only one that is responsible for the tls auth, no matter what url the user set.

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: Removed it, the gateway doesn't do multiple encodings anyway.

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: The consumer runs a single _read_loop doing sequential xreadgroup calls, so in practice we are using one connection at a time.
The pool mainly helps if multiple coroutines share the same client and issue concurrent Redis commands

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/core/handlers/webhook/webhook_event.py
Hadar-Vaknin: 👍

port_ocean/config/settings.py
raskinmaya: This is a real mismatch, not just cosmetic: with `enable_tls=True` the consumer force-sets `SSLConnection` even on a `redis://` URL, and a `rediss://` URL with `enable_tls=False` skips all SSL kwargs and relies on `from_url` inferring TLS from the scheme. A `@validator` (the v1 one works here) would prevent the footgun

raskinmaya (approved):

PR #3401, opened 2026-06-19
[Integration][Azure DevOps] Add first commit enrichment to Azure DevOps builds for DORA metrics

dennis-quabynah-bilson:
# Description

**What** — Adds an opt-in `enrichWithFirstCommit` selector flag to the Azure DevOps `build` kind. When enabled, each build is enriched under `__firstCommit` with the earliest commit it shipped (`__sha`, `__timestamp` in UTC, `__commitCount`).

**Why** — Powers the DORA **Lead Time for Changes** metric, computed as `finishTime - __firstCommit.__timestamp`. Without the first-commit timestamp, lead time can't be derived from build data alone.

**How** —
- New `AzureDevopsBuildSelector` exposing `enrichWithFirstCommit` (default `false`), wired into `AzureDevopsBuildConfig` and read in the `build` resync handler.
- `generate_builds()` / `iter_builds()` accept an `enrich_with_first_commit` flag. When set, `_enrich_builds_with_first_commit` fans out over each batch via `process_in_queue` (capped at `MAX_CONCURRENT_BUILDS_FOR_FIRST_COMMIT = 25`).
- `_attach_first_commit` pages the build `/changes` endpoint and picks the chronologically earliest change using `_parse_change_timestamp` (ISO 8601 parse, not string compare — handles fractional seconds). Falls back to the build's `sourceVersion` via a new `get_commit()` call when no changes are returned. Enrichment failures are caught and logged without failing the build.
- Opt-in by default; adds **one extra API call per build** only when enabled.
- Example mapping updated with a `leadTimeHours` jq expression; version bumped `0.9.20 → 0.9.21` with changelog entry.
- Added 4 unit tests: enrichment, time-vs-string ordering, `sourceVersion` fallback, and the no-data skip path.

## Type of change

## Integration testing checklist

- [ ] Resync able to create entities (builds with `__firstCommit`)
- [x] Updated example raw data / mapping in the `examples` folder (`example-mappings.yaml`)
- [ ] Ran the integration with example data and verified `leadTimeHours` is computed
- [ ] Docs PR link [here](#)

## Preflight checklist

- [x] Handled rate limiting (concurrency capped at 25 via `process_in_queue`; enrichment is opt-in)
- [x] Handled pagination (`/changes` paged via continuation token)
- [x] Support Multi account (enrichment runs per organization in `iterate_per_organization`)

## API Documentation

- [Builds - Get Build Changes](https://learn.microsoft.com/en-us/rest/api/azure/devops/build/builds/get-build-changes)
- [Commits - Get Commit](https://learn.microsoft.com/en-us/rest/api/azure/devops/git/commits/get)


integrations/azure-devops/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "azure-devops"
-version = "0.9.20"
+version = "0.9.21"
```
matan84:
```suggestion
version = "0.10.0"
```

matan84 (approved):

PR #3404, opened 2026-06-21
[Core] Incremental Sync Infrastructure

danamsalem-ux:
# Description

### What
Ocean core support for incremental sync: INCREMENTAL_RESYNC event, @ocean.on_incremental_resync, sync_incremental(), cursor GET/SAVE via Port API, active_incremental_cursor() for handlers, and T1/T2 strategy helpers. ONCE listener runs incremental when incrementalSyncEnabled is set. Includes tests + fake-integration POC.

### Why
Avoid full resync cost on every tick — fetch only changes since the last successful run, with per-kind cursors persisted in integ-service.

### How
For each registered kind: load cursor (seed now − interval if missing) → set active_incremental_cursor → run existing ETL via process_resource → save cursor on success, retry on failure. Handlers are separate from full resync but reuse the same pipeline; skips reconciliation and full-resync sync-metrics lifecycle.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Touches core sync/ETL paths and Port entity load semantics for incremental events (no reconciliation, direct upserts); cursor correctness on partial failure is intentional but operational misconfiguration could miss or duplicate data until a full resync.
> 
> **Overview**
> Adds **incremental sync** to Ocean core so integrations can sync only changes since the last successful run, with per-kind cursors stored in integ-service.
> 
> **New surface area:** `EventType.INCREMENTAL_RESYNC`, `@ocean.on_incremental_resync`, `ocean.on_incremental_resync`, and `sync_incremental()` orchestrate registered kinds through the existing ETL (`process_resource`) while binding the cursor via `active_incremental_cursor()` / `with_active_incremental_cursor`. Port client gains GET/PUT/POST/DELETE on `/integration/{id}/cursor`; `CursorStore` wraps that per `(kind, index)`.
> 
> **Behavior:** Cursors seed from `now − interval` on first run; `next_cursor` is fixed at run start; failed kinds do not advance the cursor and stop later kinds. Incremental runs skip full-resync reconciliation, Port diff-before-upsert, and per-kind sync metrics/DSP kind lifecycle (load path upserts directly for non-`RESYNC` events). T1/T2 helpers (`ServerSideTimestampStrategy`, `ClientSideCutoffStrategy`, `paginate_with_strategy`) support handler implementations.
> 
> **Config & scheduling:** `IntegrationSettings` adds `incremental_sync_enabled` and `incremental_sync_interval` (parsed via new `parse_interval_to_minutes` and SaaS `incrementalSyncInterval` / shared `read_app_spec_interval` helpers). The **ONCE** listener runs `sync_incremental` instead of full resync when incremental sync is enabled (e.g. Helm CronJob).
> 
> **CI:** `build-rc-integrations` now depends on `publish-core-rc` so integration RC images build after core is published or skipped.
> 
> Tests cover settings parsing, cursor context/store, strategies, and `sync_incremental` semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e4b584113719e9056ad636155850f1bcf862bde. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -1184,6 +1195,144 @@ async def _handle_resync_abortion(
...
+        ):
+            app_config = await self.port_app_config_handler.get_port_app_config(
+                use_cache=False
+            )
+
+            incremental_resources = [
+                (index, resource_cfg)
+                for index, resource_cfg in enumerate(app_config.resources)
+                if self.event_strategy["incremental"].get(resource_cfg.kind)
+            ]
```
raskinmaya, line 1271: This filter skips resources when the integration registered only a catch-all incremental handler (`kind=None`). That is inconsistent with `_collect_resync_functions`, which supports both kind-specific and catch-all listeners. I’d reuse the same “is this resource supported?” logic here, otherwise `on_incremental_resync(..., kind=None)` looks supported but won’t actually run

port_ocean/core/incremental/strategies.py
```diff
@@ -0,0 +1,205 @@
...
+        stop_field: str,
+        query_params: dict[str, Any] | None = None,
+    ) -> None:
+        self._stop_field = stop_field
+        self._query_params = query_params or {}
+
+    def build_params(self, cursor: datetime | None) -> dict[str, Any]:
+        if cursor is None:
+            return {}
+        return dict(self._query_params)
```
raskinmaya, line 152: I think this should always return `query_params`, not only when cursor exists. For T2 APIs we rely on descending time order to make pagination deterministic; on the first run, returning `{}` means we may fetch in arbitrary/default order, which is risky against third-party APIs and makes the strategy semantics inconsistent. First-run behavior is exactly where third-party API defaults hurt you. The cleaner solution is simpler too: always return `dict(self._query_params)`
WDYT?

port_ocean/core/incremental/strategies.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def should_stop(self, item: dict[str, Any], cursor: datetime | None) -> bool:
+        """Return ``True`` when *item* predates *cursor* and pagination should stop."""
+        if cursor is None:
+            return False
+        raw = item.get(self._stop_field)
+        if not raw:
+            return False
+        item_time = datetime.fromisoformat(str(raw))
+        return item_time < cursor
```
raskinmaya, line 176: This parse path is a bit fragile for third-party timestamps. A lot of APIs still return values like `2026-06-01T12:00:00Z`, and `fromisoformat` has historically been picky about that format. We already have `convert_str_to_utc_datetime` in the codebase; reusing it here would make this safer and avoid duplicating date-parsing behavior

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -1184,6 +1195,144 @@ async def _handle_resync_abortion(
...
+                        resync_id=event.id,
+                        integration_id=ocean.config.integration.identifier,
+                        integration_type=ocean.config.integration.type,
+                    )
+                    return
+
+            await ocean.app.lifecycle_client.notify_resync_finished(
+                resync_id=event.id,
+                integration_id=ocean.config.integration.identifier,
+                integration_type=ocean.config.integration.type,
```
raskinmaya: The gating here looks inconsistent: `notify_resync_started` is sent only when DSP is enabled, but `notify_resync_failed`/`notify_resync_finished` are unconditional. Was that intentional? It feels cleaner to keep all three under the same DSP condition, otherwise incremental runs may emit lifecycle events in modes where full resync does not

port_ocean/core/incremental/cursor_store.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+
+    async def delete_all(self) -> None:
+        """Remove all cursors for this integration.
+
+        Should be called after a successful full resync so the next incremental
+        run starts from a clean baseline.
+        """
+        await self._client.delete_integration_cursors()
+        logger.debug("All incremental cursors deleted")
```
raskinmaya: `delete_all` says it “should be called after a successful full resync,” but cursor invalidation for mapping changes (and delete) already lives in port-api/integ-service. Either drop/unused-wrap this ocean helper until something actually needs to call delete from ocean, or update the docstring to match the real ownership. Not blocking if intentional

port_ocean/core/incremental/strategies.py
danamsalem-ux: luv ya

port_ocean/core/integrations/mixins/sync_raw.py
danamsalem-ux: actually you added the dsp enabled check haha, ill add it

port_ocean/core/integrations/mixins/sync_raw.py
danamsalem-ux: I think its preffered we keep this logic here, since the logic in _collect_resync_functions has support for the old way to register resync hanlders: 
class MyIntegration(BaseIntegration):
    async def _on_resync(self, kind: str) -> list:

which is different than the current decorator way

port_ocean/core/integrations/mixins/sync_raw.py
raskinmaya: mi melech al halal

raskinmaya (approved):

PR #3406, opened 2026-06-21
[Integration][gitlab-v2] Add `trigger_pipeline` workflow action node

ozzs:
# Description

What - Add a `trigger_pipeline` action node to the `gitlab-v2` integration, enabling Port workflows to trigger GitLab CI/CD pipelines and track their completion.

Why - Port's workflow system supports `INTEGRATION_ACTION` nodes that delegate execution to Ocean integrations. GitLab was missing this capability, which GitHub already has via `dispatch_workflow`.

How - Follows the exact pattern established by the GitHub integration:
- **Executor** (`gitlab/actions/trigger_pipeline_executor.py`) extracts `project`, `ref`, and optional `pipelineVariables` from the run's `execution_properties`, calls `GitLabClient.trigger_pipeline()`, then immediately calls `update_run_started()` with the pipeline URL and a correlation ID (`gl_{project_id}_{pipeline_id}`). Variables are passed in GitLab's `[{"key": k, "value": v}]` format; non-string values are JSON-serialized. GitLab API errors surface the `.message` field from the response body.
- **Webhook processor** (`gitlab/webhook/webhook_processors/trigger_pipeline_webhook_processor.py`) listens for `Pipeline Hook` events, filters to terminal statuses only (`success`, `failed`, `canceled`, `skipped`), looks up the Port run by external ID, and calls `report_run_completed()`. Three early-exit guards: no matching run, `reportPipelineStatus=false`, or run already completed.
- **Registry** (`gitlab/actions/registry.py`) registers the executor with Ocean.
- **`integration.py`** — fixes a pre-existing bug where the custom `GitlabLiveEventsProcessorManager` wasn't synced to `execution_manager._webhook_manager` after replacement, causing action webhook processors to be registered into the old (unused) manager and never receive events.
- **`main.py`** — calls `register_actions_executors()` at module level; the executor's `WEBHOOK_PROCESSOR_CLASS` auto-registers the webhook processor on `/hook/{group_id}`.
- **`spec.yaml`** — declares `actionsProcessingEnabled: true` and the `trigger_pipeline` action with inputs: `project`, `ref`, `reportPipelineStatus` (default `true`), and `pipelineVariables`.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

N/A

## API Documentation

- [Create pipeline – GitLab REST API](https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline)
- [Pipeline webhooks – GitLab](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#pipeline-events)


integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
```diff
@@ -0,0 +1,60 @@
...
+            pipeline = await self.client.trigger_pipeline(project, ref, variables)
+        except httpx.HTTPStatusError as e:
+            try:
+                message = e.response.json().get("message", str(e))
+            except Exception:
+                message = str(e)
+            raise Exception(f"Failed to trigger pipeline: {message}")
+
+        external_id = build_external_id(pipeline["project_id"], pipeline["id"])
+        await ocean.port_client.update_run_started(
```
raskinmaya, line 60: The executor always calls `update_run_started`, putting the run in `InProgress`. When `reportPipelineStatus=false`, the webhook skips completion and nothing else completes the run.

If this flag is meant to be fire-and-forget, shouldn't the executor complete the run immediately after a successful trigger instead of waiting for a webhook?

integrations/gitlab-v2/gitlab/webhook/webhook_processors/trigger_pipeline_webhook_processor.py
```diff
@@ -0,0 +1,81 @@
...
+
+    async def handle_event(
+        self, payload: EventPayload, resource_config: ResourceConfig
+    ) -> WebhookEventRawResults:
+        project_id = payload["project"]["id"]
+        pipeline_id = payload["object_attributes"]["id"]
+        status = payload["object_attributes"]["status"]
+
+        external_id = build_external_id(project_id, pipeline_id)
+        run = await ocean.port_client.find_run_by_external_id(external_id)
```
raskinmaya, line 46: I think there might be a race on fast-failing pipelines;
The flow today:
1. Trigger pipeline
2. Patch Port with `externalRunId`
3. Wait for terminal pipeline webhook

If gitlab emits the terminal webhook before step 2 finishes, `find_run_by_external_id` returns `None`, the processor logs at debug and returns, and the run never completes

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
```diff
@@ -0,0 +1,60 @@
...
+)
+
+
+class TriggerPipelineExecutor(AbstractGitlabExecutor):
+    ACTION_NAME = "trigger_pipeline"
+    WEBHOOK_PROCESSOR_CLASS = TriggerPipelineWebhookProcessor
+    WEBHOOK_PATH = "/hook/{group_id}"
+
+    async def _get_partition_key(self, run: ActionRun | WorkflowNodeRun) -> str | None:
+        return run.execution_properties.get("project")
```
raskinmaya: This forces sequential execution per project. Gitlab's `POST /projects/:id/pipeline` synchronously returns `id`, `project_id` and `web_url` for the created pipeline (see [here](https://docs.gitlab.com/api/pipelines/#create-a-new-pipeline)); tracking uses `GET /projects/:id/pipelines/:pipeline_id` and webhooks carry `project.id` + `object_attributes.id`. Unlike github workflow dispatch, where polling + workflow-level partition key is needed, there is no create-time ambiguity here - concurrent triggers on the same project get distinct pipeline IDs, so IMO, this can be removed

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
```diff
@@ -0,0 +1,60 @@
+import json
```
raskinmaya, line 1: Can we add tests? I'm missing some cases:
1. Happy path trigger + `update_run_started`
2. `reportPipelineStatus=false` completes without webhook
3. terminal webhook completes run
4. duplicate terminal webhook ignored
5. `manual`/already terminal pipeline from API

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
```diff
@@ -0,0 +1,60 @@
...
+    async def execute(self, run: ActionRun | WorkflowNodeRun) -> None:
+        project = run.execution_properties.get("project")
+        ref = run.execution_properties.get("ref")
+
+        if not (project and ref):
+            raise Exception("project and ref are required")
+
+        raw_variables: dict[str, Any] = run.execution_properties.get(
+            "pipelineVariables", {}
+        )
```
raskinmaya: This does not handle:
1. `pipelineVariables: null` -> `.items()` on `None` -> crash
2. non-dict jq output (list/string) -> crash

Github's `_parse_inputs` at least assumes a dict after extraction

integrations/gitlab-v2/.port/spec.yaml
raskinmaya, line 1: What's the reason for the changes in this file?

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
```diff
@@ -0,0 +1,60 @@
...
+        ]
+
+        try:
+            pipeline = await self.client.trigger_pipeline(project, ref, variables)
+        except httpx.HTTPStatusError as e:
+            try:
+                message = e.response.json().get("message", str(e))
+            except Exception:
+                message = str(e)
+            raise Exception(f"Failed to trigger pipeline: {message}")
```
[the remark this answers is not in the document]
raskinmaya: @ozzs

integrations/gitlab-v2/.port/spec.yaml
```diff
@@ -29,12 +29,37 @@ saas:
...
+        required: true
+      - name: reportPipelineStatus
+        title: Report pipeline status
+        type: boolean
+        description: Whether to report pipeline completion status back to Port
+        default: true
+      - name: pipelineVariables
+        title: Pipeline variables
+        type: jqObject
+        description: Key-value pairs passed as CI/CD variables to the pipeline
```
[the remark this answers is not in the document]
raskinmaya: @ozzs

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
```diff
@@ -0,0 +1,60 @@
...
+
+        external_id = build_external_id(pipeline["project_id"], pipeline["id"])
+        await ocean.port_client.update_run_started(
+            run,
+            pipeline["web_url"],
+            external_id,
+        )
```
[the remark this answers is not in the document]
raskinmaya: @ozzs

integrations/gitlab-v2/gitlab/webhook/webhook_processors/trigger_pipeline_webhook_processor.py
```diff
@@ -0,0 +1,81 @@
...
+            return False
+        status = event.payload.get("object_attributes", {}).get("status")
+        return status in TERMINAL_PIPELINE_STATUSES
+
+    async def handle_event(
+        self, payload: EventPayload, resource_config: ResourceConfig
+    ) -> WebhookEventRawResults:
+        project_id = payload["project"]["id"]
+        pipeline_id = payload["object_attributes"]["id"]
+        status = payload["object_attributes"]["status"]
```
omby8888: use `.get` instead of direct call to prevent possible `KeyError`

integrations/gitlab-v2/.port/spec.yaml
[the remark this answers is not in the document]
ozzs: Already handled.
`pyproject.toml` is bumped to `0.7.41` and `CHANGELOG.md` has a matching `## 0.7.41 (2026-06-21)` entry with the feature description. No changes needed.

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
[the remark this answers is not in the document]
ozzs: Fixed.
`base_client.py` now only swallows 403/404 for `GET`. non-GET methods re-raise `HTTPStatusError`, which the executor's existing catch block already handles. Added a defensive guard in the executor as a fallback for any incomplete response that slips through.

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
ozzs: I changed it so that when `reportPipelineStatus=false`, the executor now calls `report_run_completed(success=True)` immediately after `update_run_started()`, making it a true fire-and-forget. The webhook processor's guard is kept as-is since it's still the right thing to skip if somehow a webhook arrives for such a run.

integrations/gitlab-v2/gitlab/webhook/webhook_processors/trigger_pipeline_webhook_processor.py
ozzs: It's the same window that exists in the GitHub executor and is accepted as-is. Closing it would require retrying `find_run_by_external_id` with backoff in the webhook processor, which adds non-trivial complexity for an edge case (pipeline failing in <100ms)

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
ozzs: Removed.
The partition key was copied from GitHub where polling creates ambiguity across concurrent triggers. GitLab returns the pipeline ID synchronously so there's no tracking ambiguity - parallel triggers on the same project are safe.

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
ozzs: Added tests, covering all five cases plus variable serialization and non-success terminal statuses:
* `tests/actions/test_trigger_pipeline_executor.py`
* `tests/webhook/webhook_processors/test_trigger_pipeline_webhook_processor.py`

LMK what you think :)

integrations/gitlab-v2/gitlab/actions/trigger_pipeline_executor.py
ozzs: Fixed.
`or {}` guards against `null`, `isinstance(dict)` check rejects non-dict jq output with a clear error

integrations/gitlab-v2/.port/spec.yaml
ozzs: The formatter fixes it for me when I save the file 😕

`actionsProcessingEnabled: true` opts the integration into the actions framework, and the actions block declares the `trigger_pipeline` action with its inputs so Port's UI can render the configuration form.

integrations/gitlab-v2/gitlab/webhook/webhook_processors/trigger_pipeline_webhook_processor.py
raskinmaya: Okay, might be worth documenting though

raskinmaya (approved):

integrations/gitlab-v2/gitlab/webhook/webhook_processors/trigger_pipeline_webhook_processor.py
ozzs: Documented in the `handle_event` method where the `run is None` guard is

PR #3410, opened 2026-06-22
[Integration][GitHub] Add enrichWithFirstCommit flag for deployment lead time

dennis-quabynah-bilson:
# Description

**What** — Adds an opt-in `enrichWithFirstCommit` selector flag to the GitHub `deployment` kind. When enabled, each deployment is enriched under `__firstCommit` with the earliest commit shipped since the previous deployment to the same environment (`__sha`, `__timestamp` in UTC, `__commitCount`).

**Why** — Enables the DORA *Lead Time for Changes* metric, computed as `created_at - __firstCommit.__timestamp`.

**How** — A new selector field on `GithubDeploymentSelector` (`integration.py`) flows through `ListDeploymentsOptions` (`options.py`) and `resync_deployments` (`main.py`) into the deployment exporter (`deployment_exporter.py`), which compares against the previous deployment to the same environment. Defaults to `false`; when on, it adds one extra API call per deployment.

## Type of change

## Screenshots
image
image

- [x] Added example raw data / mapping / expected result and unit tests for the new enrichment
- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [x] Support Multi account


integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+                logger.info(
+                    f"Fetched batch of {len(deployments)} deployments from repository {repo_name} from {organization}"
+                )
+                yield [
+                    enrich_with_organization(
+                        enrich_with_repository(deployment, cast(str, repo_name)),
+                        organization,
+                    )
+                    for deployment in deployments
+                ]
```
omby8888: this can be moved to function as it's duplicated later

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+            return
+
+        buffered: list[dict[str, Any]] = []
```
omby8888: why do we save the deployments batch into an outer variable? i dont see where it's being used outside of this loop

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+        organization: str,
+        repo_name: str,
+        deployment: dict[str, Any],
+        predecessor_sha: Optional[str],
+        semaphore: asyncio.Semaphore,
+    ) -> dict[str, Any]:
+        async with semaphore:
+            return await self._attach_first_commit(
+                organization, repo_name, deployment, predecessor_sha
+            )
```
omby8888: this can be replaced entirely by `semaphore_async_iterator` from port_ocean.utils.async_iterators at the task level

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+                    timestamped,
+                    key=lambda commit: _parse_timestamp(
+                        commit["commit"]["committer"]["date"]
+                    ),
+                )
+                return {
+                    **earliest,
+                    "__sha": earliest.get("sha"),
+                    "__timestamp": earliest["commit"]["committer"]["date"],
+                    "__commitCount": comparison.get("total_commits") or len(commits),
```
omby8888: what does this means in this context of the first commit? think it should be in the outer deployment raw data directly no?

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+                for commit in commits
+                if ((commit.get("commit") or {}).get("committer") or {}).get("date")
             ]
-            yield batch
+            if timestamped:
+                earliest = min(
+                    timestamped,
+                    key=lambda commit: _parse_timestamp(
+                        commit["commit"]["committer"]["date"]
+                    ),
```
omby8888: this would be misleading in case the timestamp is not parsable to datetime as u return max datetime in that case, i don't think u can safely determine the first commit in that case so i would skip setting it for a deployment with that case

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+                self.client, organization, repo_name, predecessor_sha, deployment_sha
+            )
+            commits = comparison.get("commits") or []
+            timestamped = [
+                commit
+                for commit in commits
+                if ((commit.get("commit") or {}).get("committer") or {}).get("date")
             ]
-            yield batch
+            if timestamped:
```
omby8888:
```suggestion
            if len(timestamped):
```

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+        if predecessor_sha and predecessor_sha != deployment_sha:
+            comparison = await fetch_commit_diff(
+                self.client, organization, repo_name, predecessor_sha, deployment_sha
+            )
+            commits = comparison.get("commits") or []
+            timestamped = [
+                commit
+                for commit in commits
+                if ((commit.get("commit") or {}).get("committer") or {}).get("date")
             ]
```
omby8888:
```suggestion
            commits = [
                commit
                for commit in comparison.get("commits", [])
                if commit.get("commit", {}).get("committer", {}).get("date")
            ]
```

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -1,16 +1,30 @@
...
+from collections import defaultdict
+from datetime import datetime, timezone
+from typing import cast, Any, Optional
 from github.clients.http.rest_client import GithubRestClient
 from github.core.exporters.abstract_exporter import AbstractGithubExporter
 from github.helpers.utils import (
     enrich_with_repository,
     parse_github_options,
     enrich_with_organization,
+    fetch_commit_diff,
```
omby8888, line 10: shouldn't it be imported from utils?

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+        if committer.get("date"):
+            return {
+                **commit,
+                "__sha": commit.get("sha") or deployment_sha,
+                "__timestamp": committer["date"],
+                "__commitCount": 1,
+            }
+        return None
+
+    async def _get_commit(
```
omby8888: this should be moved to utils as it could be relevant in other places

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -42,19 +56,158 @@ async def get_paginated_resources[
...
+
+        commit = await self._get_commit(organization, repo_name, deployment_sha)
+        committer = (commit.get("commit") or {}).get("committer") or {}
+        if committer.get("date"):
+            return {
+                **commit,
+                "__sha": commit.get("sha") or deployment_sha,
+                "__timestamp": committer["date"],
+                "__commitCount": 1,
+            }
```
omby8888: should be unified in one place rather than being duplicated

omby8888 (changes requested):
Nice work!

integrations/github/github/core/exporters/deployment_exporter.py
dennis-quabynah-bilson: I buffered the repo's deployments, group by environment, and sort by created_at so the pairing is correct regardless of page boundaries or response order. It only buffers when the opt-in flag is on; the default path still streams page-by-page.

integrations/github/github/core/exporters/deployment_exporter.py
```diff
@@ -31,30 +63,158 @@ async def get_resource[
...
                     organization,
+                    repo,
+                    deployment,
+                    predecessor_sha,
+                ),
+            )
+            for deployment, predecessor_sha in self._pair_predecessors(buffered)
+        ]
+        async for enriched in stream_async_iterators_tasks(*streams):
+            yield enriched
```
omby8888: shouldn't this whole block be indented under the self.client.send_paginated_request(endpoint, params) iteration? that's what i meant also with the buffered variable, we don't use it imo

dennis-quabynah-bilson: Oh I see what you meant

quabynah-bilson: The block can't move under the `send_paginated_request` loop without breaking correctness. We page at `PAGE_SIZE = 100`, and pairing is per-environment by `created_at` (`_pair_predecessors`), so a deployment's predecessor can sit on a *later* page — enriching per-batch would mis-pair every deployment at a page boundary (the predecessor resolves to `None` and we fall back to a single-commit count). Buffering the full set first is what makes the pairing page-order-independent; it's covered by `test_first_commit_pairs_predecessor_regardless_of_order`, and I added `test_first_commit_pairs_predecessor_across_pages` (predecessor delivered on a second page) to lock that in.

I did tighten the readability though: the loose `buffered` accumulator is now a single comprehension, and the enrichment moved into `_stream_first_commit_enrichment`, so there's no stray outer variable hanging around.

omby8888 (approved):
LGTM

PR #3414, opened 2026-06-22
[Integration][Gitlab-v2] Fix trigger_pipeline runs stuck in IN_PROGRESS

raskinmaya:
# Description
Fixes `trigger_pipeline` workflow runs staying stuck in `IN_PROGRESS` after the GitLab pipeline has already finished.

What changed
- Webhook race: retry lookup by `externalRunId` with exponential backoff (~7.5s) when the completion webhook arrives before Port saves the ID; completion always re-fetches the run by externalRunId before reporting status
- Duplicate webhook completion: treat 409 Conflict (run already completed) as a no-op so a late or retried webhook does not error after the run has already been reported
- `ACTION` webhook post-processing: skip entity sync for action webhook results without a mapped resource, fixing `Resource has not been set` after successful webhook completion
- Visibility: add Port run logs during trigger (Triggering…, Pipeline triggered: <url>) so progress shows beyond "Pending polling"
- Webhook-only completion: relies on GitLab pipeline webhooks for status updates (same model as GitHub `dispatch_workflow`); no background polling fallback, so the single action worker is not held in the event loop while waiting for pipeline completion

Happy path:
image
image

Failed run:
image
image

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/gitlab-v2/gitlab/actions/pipeline_completion.py
```diff
@@ -0,0 +1,70 @@
...
+    finishes writing. Default: 0.5s, 1s, 2s, 4s → ~7.5s total."""
+    delay = initial_delay
+    for attempt in range(retries):
+        run = await ocean.port_client.find_run_by_external_id(external_id)
+        if run is not None:
+            return run
+        if attempt < retries - 1:
+            await asyncio.sleep(delay)
+            delay *= 2
+    return None
```
omby8888: this should be in core

integrations/gitlab-v2/integration.py
```diff
@@ -755,7 +756,16 @@ class GitlabHandlerMixin(HandlerMixin):
...
+    async def sync_raw_results(
+        self, webhook_events_raw_result: list[WebhookEventRawResults]
+    ) -> None:
+        syncable_results = [
+            result
+            for result in webhook_events_raw_result
+            if result._resource is not None
+        ]
+        if syncable_results:
+            await super().sync_raw_results(syncable_results)
```
omby8888, line 768: what does this means?

raskinmaya: `TriggerPipelineWebhookProcessor` is registered as an `ACTION` webhook processor. Unlike catalog sync processors (which are tied to a `ResourceConfig` kind like `project` or `merge-request`), action processors don't map to any catalog resource - core passes `resource=None` when creating them.

After processing, the webhook manager batches all results and calls `sync_raw_results()`. The base implementation in `LiveEventsMixin` expects every result to have a resource set, because it transforms raw webhook data into catalog entities. Action processors like `trigger_pipeline` only complete a Port run and return empty raw results - they never touch the catalog.

Without this filter, passing action results into `super().sync_raw_results()` would fail when it tries to read `webhook_event_raw_result.resource` on a result where `_resource` is `None`.

The override filters the batch down to results that have a resource (`result._resource is not None`) before delegating to the base sync path. Action-only webhook results are effectively no-ops for catalog sync and are safely skipped

PR #3419, opened 2026-06-24
[Integration][GitHub] Add resync rate-limit quota reservation to preserve headroom for webhooks

eriport:
# Description

What - 

During heavy synchronization processes, the integration could exhaust the GitHub API rate limit entirely. Because there was no reserved quota for critical incoming webhooks, this exhaustion caused non-recoverable deadlocks.  This PR introduces a configurable utilization threshold to pause resync events early, saving the remaining rate limit budget for webhook processing.

Why -

When a large sync operation runs, it can aggressively consume 100% of the GitHub API rate limit. If a critical webhook arrives while the rate limit is completely exhausted, the integration cannot execute the necessary API calls to handle the webhook or recover, resulting in an unresolvable pod deadlock. By halting resyncs early (e.g., at 85-95% utilization), the application safeguards the remaining budget exclusively for webhook continuity.

How -

Implemented a rate limit utilization check and a configurable threshold value that controls when resync events pause execution.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

Screenshot 2026-06-24 at 11 26 20
Screenshot 2026-06-24 at 11 28 06
Screenshot 2026-06-24 at 14 26 08

## API Documentation


integrations/github/github/clients/http/base_client.py
```diff
@@ -199,3 +206,48 @@ def send_paginated_request(
...
+
+        try:
+            if event.event_type != EventType.RESYNC:
+                logger.debug(
+                    "Not a resync event, no need to check rate limit threshold."
+                )
+                return
+        except EventContextNotFoundError:
+            # we are not in an event context, we assume this is a standard action/webhook and should not be throttled
+            return
```
melodyogonna: This logic is wrong. You still need to handle rate-limits on other types of events, you just don't want to block them with quota reservation.

melodyogonna (changes requested):

integrations/github/github/clients/http/base_client.py
```diff
@@ -17,6 +21,8 @@
         AbstractGitHubAuthenticator,
     )
 
+_RATE_LIMIT_USAGE_THRESHOLD: float = 95.0
```
omby8888:
```suggestion
_DEFAULT_RATE_LIMIT_RESYNC_USAGE_THRESHOLD: float = 95.0
```

integrations/github/github/clients/http/base_client.py
```diff
@@ -199,3 +206,48 @@ def send_paginated_request(
...
+
+            if delay > 0:
+                logger.bind(
+                    api_type=self.rate_limiter.api_type,
+                    delay=delay,
+                    current_utilization=ratelimit_info.utilization_percentage,
+                    threshold=ratelimit_threshold,
+                ).warning(
+                    "Rate limit utilization exceeds threshold during resync. "
+                    "Pausing execution to save remaining budget for webhooks."
```
omby8888:
```suggestion
                    "Pausing execution to save remaining budget for webhooks/actions."
```

integrations/github/github/clients/http/base_client.py
```diff
@@ -105,6 +111,7 @@ async def make_request(
     ) -> Response:
         """Make a request to the GitHub API with GitHub rate limiting and error handling."""
 
+        await self._wait_if_resync_threshold_exceeded()
```
omby8888: seem weird that the base client is responsible to coordinate that, shouldn't the rate limiter handle it?

integrations/github/.port/spec.yaml
```diff
@@ -66,6 +66,11 @@ configurations:
     required: false
     default: false
     description: When enabled, the integration will not update existing webhook configurations on GitHub.
+  - name: ratelimitReservationThreshold
```
omby8888:
```suggestion
  - name: resyncRatelimitReservationThreshold
```

omby8888: this should be validated as resonable value somewhere, as there's no validation atm if value is <0 or >100

eriport: done, now it fallsback to default value upon invalid input rather than breaking the entire sync

integrations/github/github/clients/http/base_client.py
eriport: it's much better here, because, being placed at this entry point of the base request method ensures it automatically guards all outgoing calls.

melodyogonna (approved):
lgtm

integrations/github/github/clients/http/base_client.py
omby8888: but this will happen in any case, my point is that all of this handling should be encapsulated within the rate limiter, in the `_enforce_rate_limit` implementation, that way only one request will perform this logic while the others wait for the semaphore to be released

omby8888 (approved):
Nice work!

PR #3438, opened 2026-06-28
[Integration][Aws-v3] Prevent redundant per-resource work in ResourceBuilder

raskinmaya:
# Description
Jira [bug](https://getport.atlassian.net/browse/PORT-18102)

## Summary
The shared `ResourceBuilder` (used by every AWS kind, most notably the property-heavy `AWS::EC2::Instance`) does avoidable work on every resource. This removes the two redundant steps with byte-for-byte identical output.

## Changes
- No empty-default dump/merge: `with_properties` now just stashes the data instead of dumping the always-fresh empty `Properties` model and merging it back. The builder takes the model class and constructs once in build, so `Properties` is validated exactly once (no empty default instantiated and discarded).
- No JSON round-trip: build replaces `json.loads(model.json(...))` with a direct `.dict()` plus an in-place leaf converter (`_to_jsonable`) that only rewrites non-JSON-native leaves (`datetime`/`date` → ISO, others via `pydantic_encoder`).
- `resource_inspector`: hoists the model template / `model_cls` / `resource_type` out of the per-resource loop instead of calling `model_factory()` per resource.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


integrations/aws-v3/aws/core/modeling/resource_builder.py
```diff
@@ -1,78 +1,149 @@
...
+        if resource_type is None:
+            type_field = self._model_cls.__fields__.get("Type")
+            if type_field is None or type_field.default is None:
+                raise ResourceBuilderError(
+                    "No `Type` set for the resource model and the model defines no "
+                    "default, use `with_type` to set the resource type."
+                )
+            resource_type = type_field.default
+
+        # Supplying ``Properties`` directly skips the model's empty default
```
VenfiOranai: Kinda pointless comment, it feels pretty coupled to the fix rather than explaining something special about the code

integrations/aws-v3/aws/core/modeling/resource_builder.py
```diff
@@ -1,78 +1,149 @@
...
-        resource_json = self._model.json(exclude_unset=True, by_alias=True)
-        return json.loads(resource_json)
+        # Always supply ``Type`` so it is present in the output (``exclude_unset``
+        # would otherwise drop a value that came from the model default). Fall
+        # back to the model's own default when ``with_type`` was not called, and
+        # fail explicitly if the model defines no default either.
+        resource_type = self._type
+        if resource_type is None:
+            type_field = self._model_cls.__fields__.get("Type")
+            if type_field is None or type_field.default is None:
```
VenfiOranai: Not sure how necessary this condition/raise is

integrations/aws-v3/aws/core/modeling/resource_builder.py
```diff
@@ -1,78 +1,149 @@
...
-            raise ValueError(
-                "No data has been set for the resource model, use `with_data` to set data."
+        if self._properties is None:
+            raise ResourceBuilderError(
+                "No data has been set for the resource model, use `with_properties` to set data."
             )
 
-        resource_json = self._model.json(exclude_unset=True, by_alias=True)
-        return json.loads(resource_json)
+        # Always supply ``Type`` so it is present in the output (``exclude_unset``
```
VenfiOranai: I think it might be worth shortening the comment, it's a bit wordy. Just like "Explicitly set type due to exclude_unset=True call"

integrations/aws-v3/aws/core/modeling/resource_builder.py
```diff
@@ -1,78 +1,149 @@
...
         Set the type of the resource model.
         """
-        self._model.Type = type
+        self._type = type
         return self
 
     def build(self) -> Dict[str, Any]:
         """
-        Finalize and return the constructed resource model.
+        Construct the resource model in a single validation pass and return it
```
VenfiOranai: Redundant change

integrations/aws-v3/aws/core/modeling/resource_builder.py
```diff
@@ -1,78 +1,149 @@
...
 
         Returns:
             Self: The builder instance for method chaining.
         """
-
-        current_data = self._model.Properties.dict(exclude_unset=True)
-        current_data.update(data)
-        self._model.Properties = type(self._model.Properties)(**current_data)
-        self._props_set = True
+        self._properties = (
```
VenfiOranai: Why add the merging functionality? If it ever becomes relevant then I'd get it, but seeing as there's only one call to this (and it's unlikely there'll be more), I'm not sure it's worth

VenfiOranai: Same for extra context btw

integrations/aws-v3/aws/core/modeling/resource_builder.py
```diff
@@ -1,78 +1,149 @@
...
+
+
+class ResourceBuilder[ResourceModelT: ResourceModel[Any], TProperties: BaseModel]:
     """
     Builder class for constructing AWS resource models with strongly-typed properties.
 
-    Provides a fluent interface to set multiple fields on the `Properties`
-    attribute of a given resource model, which is expected to be a subclass of
-    `BaseResponseModel` parameterized with a Pydantic `BaseModel` for its properties.
+    Provides a fluent interface to collect the resource's `Type`, `Properties`
```
VenfiOranai, line 42: I think it might be worth removing the emphasis on the one run/empty model, that behavior is the irregular one, you're bringing it in line with standard Pydantic usage (and so the comment feels a bit bloated mentioning the obvious)

integrations/aws-v3/aws/core/modeling/resource_builder.py
```diff
@@ -1,78 +1,149 @@
...
+        Construct the resource model in a single validation pass and return it
+        as a JSON-native dict.
 
         Returns:
             Dict[str, Any]: The built resource model dictionary.
         """
-        if not self._props_set:
-            raise ValueError(
-                "No data has been set for the resource model, use `with_data` to set data."
+        if self._properties is None:
```
VenfiOranai: Feels a bit pointless considering the current usage, no?

VenfiOranai (changes requested):

VenfiOranai (approved):
Looks great

PR #3440, opened 2026-06-28
[Core] Feat: xautoclaim worker with Redis based leader election

Hadar-Vaknin:
# Description

Adds a PEL (Pending Entry List) requeue worker for Redis live-events consumers. When a webhook message is delivered to a consumer but never ACKed (e.g. pod crash mid-processing), it stays in the PEL. This worker periodically scans for stuck entries, re-enqueues them for retry, and discards messages that exceed a configurable retry limit.

Only one pod runs the scan at a time, coordinated via Redis-based leader election (SET NX PX + heartbeat). Every pod runs the election loop so failover is automatic when the leader dies.

Design doc: 

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation


port_ocean/consumers/pel_requeue/worker.py
```diff
@@ -0,0 +1,206 @@
...
+        self._is_running = False
+        self._lifecycle_task: asyncio.Task[None] | None = None
+        self._leader_election = RedisLeaderElection(
+            redis=redis,
+            leader_key=settings.leader_key,
+            pod_id=settings.pod_id,
+            name=_PEL_LEADER_ELECTION_NAME,
+            ttl_ms=settings.leader_ttl_ms,
+            heartbeat_seconds=settings.leader_heartbeat_seconds,
+        )
```
raskinmaya: Do we need leader election at all? `XAUTOCLAIM` atomicity + a shared consumer name already prevent duplicate claims.

All pods call `XAUTOCLAIM` with the same consumer name `"pel-requeue-worker"` and the same min-idle threshold. Redis serializes commands, and a successful claim resets the entry's idle time to ~0 - so the moment one pod claims an entry, any other pod's `XAUTOCLAIM` skips it (idle now below threshold). That means concurrent scanning across pods can't double-claim or double-requeue the same message; correctness is already guaranteed by Redis itself.

If that's right, the leader election (`leader_election` + its tests + the `_ensure_leadership`/`_run_leader_cycle`/`heartbeat` machinery) isn't buying us correctness - only a reduction in redundant scan load from N pods. A jittered "every pod scans periodically" loop would be far simpler and delete this entire subsystem (and with it the lock-safety bug from my other comment).

Can you justify the leader-election approach? If the goal is purely to cut Redis load, let's:
(a) say so explicitly in the docstring
(b) weigh that against the complexity - a small periodic scan from a handful of pods is cheap. If there's a correctness case I'm missing (e.g. a window where two pods can requeue the same entry), please spell it out, because I can't think of one

port_ocean/consumers/pel_requeue/leader_election.py
```diff
@@ -0,0 +1,102 @@
...
+        """Give up leadership and delete the lock key (best-effort)."""
+        self._is_leader = False
+        if self._heartbeat_task is not None:
+            self._heartbeat_task.cancel()
+            await asyncio.gather(self._heartbeat_task, return_exceptions=True)
+            self._heartbeat_task = None
+        try:
+            current = await self._redis.get(self._leader_key)
+            if current == self._pod_id:
+                await self._redis.delete(self._leader_key)
```
raskinmaya: Between the `GET` and the `DELETE`/`PEXPIRE`, our TTL can expire and another pod can win the election, so the key now holds its `pod_id`. We then delete (or `PEXPIRE`-extend) a key we no longer own - deleting the new leader's lock, or prolonging someone else's. Either way you briefly get two leaders both running `XAUTOCLAIM`/`XADD`, which is exactly the invariant this component exists to prevent. This is the classic unsafe distributed-lock bug

Please make each op a single atomic, owner-checked command via a Lua script. This will collapse `GET` + `DELETE`/`PEXPIRE` into one atomic step so we only act while we still own the key

This is relevant both here and [here](https://github.com/port-labs/ocean/pull/3440/changes#diff-5262e41d648c2987081cd12ca51cd571fd6fb802f8cd4a1aeb91ee708241618fR81) in lines 81-83

(if we drop leader election entirely per my other comment and lean on `XAUTOCLAIM` atomicity, this whole class of bug disappears)

port_ocean/consumers/pel_requeue/worker.py
```diff
@@ -0,0 +1,206 @@
...
+        await self._scan_and_requeue()
+        await asyncio.sleep(self._settings.scan_interval_seconds)
+
+        if not self._leader_election.is_leader:
+            logger.info(
+                "PEL requeue worker lost leadership, re-entering election",
+                stream_key=self._stream_key,
+            )
+
+    async def _scan_and_requeue(self) -> None:
```
raskinmaya, line 81: This function loops over potentially many pages with no `is_leader` check. Leadership is only re-checked after the full scan + sleep, [here](https://github.com/port-labs/ocean/pull/3440/changes#diff-69b6fa749814feffe463318d436c7cb9f872891be251276a2b0f8b2098142b5eR131). If the TTL lapses mid-scan and another pod wins, both pods run `XAUTOCLAIM`/`XADD` concurrently, defeating the single-leader guarantee. At minimum the page loop [here](https://github.com/port-labs/ocean/pull/3440/changes#diff-69b6fa749814feffe463318d436c7cb9f872891be251276a2b0f8b2098142b5eR142) should bail when `self._leader_election.is_leader` bacomes `false`

port_ocean/consumers/redis_stream_consumer.py
raskinmaya, line 260: We ACK in `finally` here, even when `_on_message` raises. That means a processing failure is acked and silently dropped - it never lingers in the PEL, so the requeue worker never sees it. The only thing that can leave an entry in the PEL is a hard pod death (like `SIGKILL` or OOM) between `XREADGROUP` and this `finally`.
That makes the worker crash-recovery only

port_ocean/consumers/pel_requeue/worker.py
```diff
@@ -0,0 +1,158 @@
...
+                stream_key=self._stream_key,
+            )
+
+    async def _handle_stuck_message(
+        self, message_id: str, fields: dict[str, str]
+    ) -> None:
+        requeue_count = int(fields.get("requeue_count", "0"))
+        max_requeue_count = self._settings.max_requeue_count
+
+        if requeue_count >= max_requeue_count:
```
raskinmaya, line 154: When `requeue_count >= max_requeue_count`, the message is `XACK`ed and gone, with only a `logger.warning`. No dead-letter stream, no metric/count. For a system whose entire purpose is "don't lose stuck events", quietly deleting them after 3 tries is the worst outcome. Preferably `XADD` to a `<stream>:dead-letter` so it's recoverable & consider adding a metric

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: That's the whole point of the pel worker. It's intended for re-enqueueing messages that were not acknowledged for the past X time, not failed messages.
 "PEL = stuck/crash recovery," not "PEL = failed-processing retry."

port_ocean/consumers/pel_requeue/settings.py
raskinmaya, line 1: Six fields are declared in `LiveEventsRedisSettings`, then mirrored one-for-one into a frozen `PELRequeueWorkerSettings` dataclass, plus a `from_live_events_redis_settings` converter that does nothing but `*1000` and `float()` casts. That's three places to touch for every knob. The worker is always constructed from `LiveEventsRedisSettings` anyway (see [here](https://github.com/port-labs/ocean/pull/3440/changes#diff-faac178ebcde4a5a31430fa331a8499d116430a0c09771ae025f44f6f836c2daR126)). Maybe pass `redis_settings` (+ `stream_key`/`consumer_group`) straight into the worker, or keep the dataclass but delete the per-field defaults?

port_ocean/consumers/pel_requeue/worker.py
Hadar-Vaknin: DLQ is not within the scope of the task, as written in the doc. It is planned for the next phases.

port_ocean/consumers/pel_requeue/settings.py
```diff
@@ -0,0 +1,38 @@
...
+    from port_ocean.config.settings import LiveEventsRedisSettings
+
+_PEL_CONSUMER_NAME = "pel-requeue-worker"
+
+
+@dataclass(frozen=True)
+class PELRequeueWorkerSettings:
+    stream_key: str
+    consumer_group: str
+    stuck_timeout_ms: int = 60_000
```
raskinmaya: Why is it 60 second here and 600 [here](https://github.com/port-labs/ocean/pull/3440/changes#diff-c4ab144128bdbbe50dd95b726ae9c2f4cbc7306025e07a9e1db3dff7c5d4b3bbR150)?

port_ocean/consumers/pel_requeue/worker.py
```diff
@@ -0,0 +1,158 @@
...
+                        "Failed to handle stuck PEL message, skipping",
+                        message_id=message_id,
+                        stream_key=self._stream_key,
+                        error=str(error),
+                    )
+
+            if next_cursor == "0-0" or not messages:
+                break
+
+            cursor = next_cursor
```
raskinmaya: `XAUTOCLAIM` can legitmately return a non-zero cursor with an empty message batch (e.g. the scanned id-range was all tombstoned/deleted entries). Breaking on `not message` aborts the pass early and leaves later PEL entries un-claimed until the next `scan_interval` (30s). Termination should be driven by the cursor alone (`next_cursor == "0-0"`), not by message emptiness

port_ocean/consumers/pel_requeue/settings.py
```diff
@@ -0,0 +1,38 @@
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from port_ocean.config.settings import LiveEventsRedisSettings
+
+_PEL_CONSUMER_NAME = "pel-requeue-worker"
```
raskinmaya: A leading underscore means module-private; importing it elsewhere isn't very good. Let's rename it to `PEL_CONSUMER_NAME` if it's needed elsewhere

port_ocean/consumers/pel_requeue/worker.py
```diff
@@ -0,0 +1,158 @@
...
+        self._is_running = False
+        self._lifecycle_task: asyncio.Task[None] | None = None
+
+    @property
+    def _stream_key(self) -> str:
+        return self._settings.stream_key
+
+    @property
+    def _consumer_group(self) -> str:
+        return self._settings.consumer_group
```
raskinmaya: These just proxy `self._settings.*`. Consider inlining them, they add lines and an extra thing to mock in tests

port_ocean/consumers/pel_requeue/settings.py
```diff
@@ -0,0 +1,38 @@
...
+        return cls(
+            stream_key=stream_key,
+            consumer_group=consumer_group,
+            stuck_timeout_ms=redis_settings.pel_stuck_timeout_seconds * 1000,
+            max_requeue_count=redis_settings.pel_max_requeue_count,
+            scan_interval_seconds=float(redis_settings.pel_scan_interval_seconds),
+            xautoclaim_count=redis_settings.pel_xautoclaim_count,
+            lifecycle_error_backoff_seconds=float(
+                redis_settings.pel_lifecycle_error_backoff_seconds
+            ),
```
raskinmaya: Both `scan_interval_seconds` and `lifecycle_error_backoff_seconds` are `float`s here but `int`s in the dataclass, forcing the cast. Let's align it

port_ocean/tests/consumers/pel_requeue/test_worker.py
```diff
@@ -0,0 +1,367 @@
+"""Tests for PELRequeueWorker."""
```
raskinmaya, line 1: Missing some tests:
1. Duplicate/slow-processing test
2. Non-zero-cursor-empty-batch test

port_ocean/consumers/pel_requeue/worker.py
```diff
@@ -0,0 +1,158 @@
...
+                cursor,
+                count=self._settings.xautoclaim_count,
+            )
+
+            if not result:
+                break
+
+            next_cursor: str = result[0]
+            messages: list[tuple[str, dict[str, str]]] = (
+                result[1] if len(result) > 1 else []
```
raskinmaya, line 101: That discards `result[2]` and relies on the per-message `try`/`except` to swallow tombstoned/`None`-field entries, which produces a `logger.exception` per dead entry. Can you handle the deleted list explicitly to avoid noisy error logs?

port_ocean/consumers/redis_stream_consumer.py
raskinmaya: ok fair

port_ocean/consumers/pel_requeue/worker.py
raskinmaya: Then let's at least add a metric for that (take a look at how the resources metrics are exposed, should be small)

port_ocean/consumers/pel_requeue/settings.py
Hadar-Vaknin: 👍

port_ocean/consumers/pel_requeue/settings.py
Hadar-Vaknin: 👍

port_ocean/consumers/pel_requeue/settings.py
Hadar-Vaknin: 👍

port_ocean/consumers/pel_requeue/worker.py
Hadar-Vaknin: 👍

port_ocean/consumers/pel_requeue/settings.py
Hadar-Vaknin: 👍

port_ocean/consumers/pel_requeue/worker.py
Hadar-Vaknin: 👍

port_ocean/tests/consumers/pel_requeue/test_worker.py
Hadar-Vaknin: 👍

port_ocean/consumers/pel_requeue/worker.py
Hadar-Vaknin: we already have a metric for failed events. Take a look at Connect SLO dashboard

port_ocean/consumers/pel_requeue/worker.py
Hadar-Vaknin: Handled explicitly now. XAUTOCLAIM's third response element (deleted_ids, Redis 7.0+) is logged at info level - no XACK needed since Redis already cleaned the PEL. 
For Redis 6.x tombstoned entries that still appear as (id, None) in the messages list, we XACK and skip them without hitting the exception path. The per-message try/except remains only for real processing failures.

Hadar-Vaknin: 👍

port_ocean/consumers/pel_requeue/worker.py
raskinmaya: someone cooked here

raskinmaya (approved):

PR #3442, opened 2026-06-28
[Integration][AzureDevops] Fix Azure DevOps Rate-Limit Retry Coordination

emekanwaoma:
# Description

What - Added Azure DevOps-specific retry handling for rate-limit responses and transport timeouts.

Why - Azure DevOps can return `429` during high-volume resyncs. Previously, the generic retry transport retried the failed request, but did not notify the shared Azure DevOps rate limiter, allowing other requests to continue hitting ADO and potentially exhaust retries.

How - Added an ADO retry transport that coordinates `429` and timeout retries through the shared rate limiter, waits 360 seconds before retrying, refreshes auth headers before retry, logs retry attempts, and caps retries at 10 attempts.

## Type of change

### Core testing checklist

- [] If new resource kind is added or updated, validate that live-events for that resource are working as expected
- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting

## Screenshots

## API Documentation

### DIFF
--- a/integrations/azure-devops/azure_devops/client/client_manager.py
+++ b/integrations/azure-devops/azure_devops/client/client_manager.py
@@ -1,3 +1,5 @@
+import os
+from contextlib import suppress
 from typing import Optional
 
 from loguru import logger
@@ -5,13 +7,29 @@ from port_ocean.context.event import event
 from port_ocean.context.ocean import ocean
 
 from azure_devops.client.auth import ACCOUNT_MODE_MULTIPLE, build_auth_provider
-from azure_devops.client.azure_devops_client import AzureDevopsClient
+from azure_devops.client.azure_devops_client import (
+    AZURE_DEVOPS_CLIENT_CACHE_KEY,
+    AzureDevopsClient,
+)
+from azure_devops.client.rate_limiter import AzureDevOpsRateLimiterRegistry
 from azure_devops.helpers.validate_config import (
     parse_organization_urls,
     validate_azure_devops_config,
 )
 
 CLIENT_MANAGER_CACHE_KEY = "azure_devops_client_manager"
+
+
+def _reset_clients_after_fork() -> None:
+    """Clear cached clients after fork so children recreate httpx and limiter state."""
+    with suppress(Exception):
+        event.attributes.pop(CLIENT_MANAGER_CACHE_KEY, None)
+        event.attributes.pop(AZURE_DEVOPS_CLIENT_CACHE_KEY, None)
+    AzureDevOpsRateLimiterRegistry.reset()
+
+
+if hasattr(os, "register_at_fork"):
+    os.register_at_fork(after_in_child=_reset_clients_after_fork)
 
 
 class AzureDevopsClientManager:
--- a/integrations/azure-devops/azure_devops/client/client_manager.py
+++ b/integrations/azure-devops/azure_devops/client/client_manager.py
@@ -1,5 +1,3 @@
-import os
-from contextlib import suppress
 from typing import Optional
 
 from loguru import logger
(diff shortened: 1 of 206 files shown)


integrations/azure-devops/azure_devops/client/client_manager.py
```diff
@@ -14,6 +20,18 @@
 CLIENT_MANAGER_CACHE_KEY = "azure_devops_client_manager"
 
 
+def _reset_clients_after_fork() -> None:
+    """Clear cached clients after fork so children recreate httpx and limiter state."""
```
danamsalem-ux: why do we need to do this?

integrations/azure-devops/azure_devops/client/rate_limiter.py
```diff
@@ -65,22 +93,31 @@ def should_wait_for_retry_after(self) -> float:
...
 
-        Used when throttling manifests as a timeout (no response headers available).
+        Used when throttling manifests as a 429 or timeout. The next request to
+        enter the limiter holds the shared lock while sleeping, so later requests
+        queue instead of continuing to hit Azure DevOps during the cooldown.
         """
         async with self._lock:
-            throttle_end_time = time.time() + seconds
-            if self._throttle_until is None or throttle_end_time > self._throttle_until:
+            now = time.time()
```
danamsalem-ux, line 77: why is this also relevant?

integrations/azure-devops/azure_devops/client/rate_limiter.py
```diff
@@ -121,7 +158,7 @@ async def __aenter__(self) -> "AzureDevOpsRateLimiter":
             ):
                 reset_wait = self.seconds_until_reset
                 if reset_wait > 0:
-                    logger.debug(
+                    logger.warning(
```
danamsalem-ux, line 134: is this specifically for debugging now? or you want to make it permanent?

integrations/azure-devops/azure_devops/client/rate_limiter.py
```diff
@@ -9,6 +10,33 @@
 LIMIT_REMAINING_HEADER = "x-ratelimit-remaining"
 LIMIT_RESET_HEADER = "x-ratelimit-reset"
 LIMIT_RETRY_AFTER_HEADER = "retry-after"
+ADO_RATE_LIMIT_WINDOW_SECONDS = 360
```
danamsalem-ux, line 12: i see you are using 3 different rate limit windows, as they are between 6-12 minutes. Why did you choose this ones exactly? because they seem a bit too long

integrations/azure-devops/azure_devops/client/rate_limiter.py
danamsalem-ux: i'll clarify - i compare this to a working solution we have in github retry transport for a similar use case. In that scenario we update the rate limit fields before and after each call to avoid more calls to be made to the api and reduce the rate limit times. Here i see you have added the signal throttling as well, and im trying to understand why is it needed? and why not just use the basic retry headers update

integrations/azure-devops/azure_devops/client/client_manager.py
emekanwaoma: since we introduced shared cached clients/rate limiters, this prevents child processes from reusing parent-process client and lock state. It’s mostly a safety fix for multi_process. We've solved tickets as such in the past.

integrations/azure-devops/azure_devops/client/rate_limiter.py
emekanwaoma: it is a little unique here because there are cases where:
- ADO returns 429 without useful reset headers.
- ADO causes ReadTimeout or ConnectTimeout, where there is no response and therefore no headers.
- The retry transport sees the 429 before the normal send_request() path finishes.

So yes, the basic header updates are still used, but signal_throttle() handles the particular ADO gap where throttling happens without enough headers to inform the rest of the workers.

integrations/azure-devops/azure_devops/client/rate_limiter.py
emekanwaoma: I plan on making it permanent for better visibility. I left some others as debug

integrations/azure-devops/azure_devops/client/rate_limiter.py
emekanwaoma: I chose them to give more time in case there are other integrations running and hitting ADOs rate limit with same token.

integrations/azure-devops/azure_devops/client/client_manager.py
danamsalem-ux: You refer the rate limit registry you have added in this pr? i just dont see how it solves the current issue. The clientManager was added for multi accounts clients that will need to manage multiple ado orgs in one integration with the same token, the fetching of the data is made org by org sequentially.
It just seems an overkill for the problem that we just need to update the rate limit headers here. Correct me if im wrong here

integrations/azure-devops/azure_devops/client/rate_limiter.py
danamsalem-ux: I think its a bit much to wait 12 minutes for rate limits, especially when the api returns different values. I do think we should add a jitter instead

danamsalem-ux (approved):

PR #3449, opened 2026-06-30
[Integration][GitHub] Add update_repo_external_properties action to sync Port blueprint properties to GitHub external properties

omby8888:
# Description

What - Added a new `update_repo_external_properties` integration action that writes Port blueprint property values to GitHub repository custom (external) properties.

Why - Customers want bidirectional sync between Port and GitHub: Port ingests GitHub external properties during resync, and automations should be able to push Port entity changes back to GitHub so repository metadata stays consistent across both systems.

How -
  - `UpdateRepoExternalPropertiesExecutor` — PATCHes GitHub's org-level external properties API with mapped property values for a target repository
  - `spec.yaml` — registers the action with `org`, `repo`, and `externalPropertiesMapping` (jqObject) inputs
  - `registry.py` — registers the new action executor
  - Unit tests for the executor and helper logic

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [x] Support Multi account

## Screenshots

N/A — action-only change, no new resync kinds.

## API Documentation

- [Set values for a custom property for repositories in an organization](https://docs.github.com/en/rest/orgs/custom-properties?apiVersion=2022-11-28#set-values-for-a-custom-property-for-repositories-in-an-organization)


integrations/github/github/actions/update_repo_external_properties_executor.py
```diff
@@ -0,0 +1,103 @@
...
+from pydantic import BaseModel
+
+from github.actions.abstract_github_executor import AbstractGithubExecutor
+from port_ocean.context.ocean import ocean
+from port_ocean.core.models import ActionRun, WorkflowNodeRun
+
+
+class ExternalProperty(BaseModel):
+    property_name: str
+    value: str | None
```
raskinmaya: `value` is typed and coerced as `str | None`, but GH accepts `null | string | array` depending on schema for custom properties. Coercing everything with `str(value)` will break real schemas

integrations/github/github/actions/update_repo_external_properties_executor.py
```diff
@@ -0,0 +1,103 @@
...
+        org = run.execution_properties.get("org")
+        repo = run.execution_properties.get("repo")
+        return f"{org}/{repo}"
+
+    async def execute(self, run: ActionRun | WorkflowNodeRun) -> None:
+        org = run.execution_properties.get("org")
+        repo = run.execution_properties.get("repo")
+        external_properties_mapping = run.execution_properties.get(
+            "externalPropertiesMapping"
+        )
```
raskinmaya, line 49: `org` and `repo` are required in spec but never validated here. Missing values produce partition key `"None/None"` and URL `/orgs/None/...` (look [here](https://github.com/port-labs/ocean/pull/3449/changes#diff-dc3b31f8ba1bbdf8352d0917d5b56f003a7a035e9da8778de1ba8f4ca617644dR1-R103)); `dispatch_workflow_executor` raises `InvalidActionParametersException` for missing inputs - please could you match that pattern? Also validate `externalPropertiesMapping` is a `dict`; a malformed `jqObject` will crash on `.items()` [here](https://github.com/port-labs/ocean/pull/3449/changes#diff-dc3b31f8ba1bbdf8352d0917d5b56f003a7a035e9da8778de1ba8f4ca617644dR1-R103)

integrations/github/github/actions/update_repo_external_properties_executor.py
```diff
@@ -0,0 +1,103 @@
...
+                property_name=name,
+                value=None if value is None or value == "" else str(value),
+            )
+            for name, value in external_properties_mapping.items()
+        ]
+
+
+class PatchExternalPropertiesBody(BaseModel):
+    repository_names: list[str]
+    properties: list[ExternalProperty]
```
raskinmaya: IMO, for a single-repo action, the org batch endpoint is the wrong abstraction: extra `repository_names` wrapper, org-level permissions, and asymmetry with ingest. A simpler approach would be:
```
PATCH /repos/{org}/{repo}/properties/values
{"properties": [{"property_name": "...", "value": ...}]}
```

That lets you drop the `PatchExternalPropertiesBody` entirely

integrations/github/github/actions/update_repo_external_properties_executor.py
```diff
@@ -0,0 +1,103 @@
...
+import httpx
+from loguru import logger
+from pydantic import BaseModel
+
+from github.actions.abstract_github_executor import AbstractGithubExecutor
+from port_ocean.context.ocean import ocean
+from port_ocean.core.models import ActionRun, WorkflowNodeRun
+
+
+class ExternalProperty(BaseModel):
```
raskinmaya: Two Pydantic models + `from_dict` for a one-shot API call feels heavy. The exporter already does the inverse transform inline (array <-> dict). Unless these models are reused or schema-validated, an inline list comprehension would be enough. WDYT?

I also think that `@staticmethod` on the model is an odd home for serialization logic

integrations/github/github/actions/update_repo_external_properties_executor.py
```diff
@@ -0,0 +1,103 @@
...
+            logger.info("Processing external property update")
+
+            if not external_properties_mapping:
+                logger.info(
+                    "No properties changes detected — nothing to push to GitHub"
+                )
+                await ocean.port_client.report_run_completed(
+                    run, success=True, message="No changes to apply."
+                )
+                return
```
raskinmaya: `externalPropertiesMapping` is `required: true` in spec, but empty or missing mapping is treated as success ("No changes to apply"). Feels like a product mismatch - either make the input optional in spec, or fail fast on empty mapping so automations don't silently no-op

integrations/github/github/actions/update_repo_external_properties_executor.py
```diff
@@ -0,0 +1,103 @@
...
+        with logger.contextualize(org=org, repo=repo):
+            logger.info("Processing external property update")
+
+            if not external_properties_mapping:
+                logger.info(
+                    "No properties changes detected — nothing to push to GitHub"
+                )
+                await ocean.port_client.report_run_completed(
+                    run, success=True, message="No changes to apply."
+                )
```
raskinmaya: Won't the message be ignored for plain `ActionRun`s? It only patches status (take a look at `actions_and_workflow_runs`). Success/skip message won't show in the UI; I think you'll need to use something like the `post_run_log` I used when working on the gitlab `trigger_pipeline` feature

integrations/github/github/actions/update_repo_external_properties_executor.py
```diff
@@ -0,0 +1,103 @@
...
+                return
+
+            github_properties = ExternalProperty.from_dict(external_properties_mapping)
+
+            try:
+                await self.rest_client.make_request(
```
raskinmaya, line 67: Writing custom proeprty values requires fine-grained permissions (take a look [here](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2026-03-10#repository-permissions-for-custom-properties) and [here](https://docs.github.com/en/rest/authentication/permissions-required-for-fine-grained-personal-access-tokens?apiVersion=2026-03-10#organization-permissions-for-custom-properties)). Many existing GH app installation won't have these. Worth documenting and logging a clear error on 403

integrations/github/github/actions/update_repo_external_properties_executor.py
```diff
@@ -0,0 +1,103 @@
...
+            if not external_properties_mapping:
+                logger.info(
+                    "No properties changes detected — nothing to push to GitHub"
+                )
+                await ocean.port_client.report_run_completed(
+                    run, success=True, message="No changes to apply."
+                )
+                return
+
+            github_properties = ExternalProperty.from_dict(external_properties_mapping)
```
raskinmaya: GH only accepts values for properties already defined in the org schema. Arbitrary keys in the mapping will return 422. The spec example implies free-form keys so customers need to pre-provision properties in GH first

omby8888: not relevant to external props

integrations/github/github/actions/update_repo_external_properties_executor.py
omby8888: this endpoint accepts only string/null according to the docs they've sent me

integrations/github/github/actions/update_repo_external_properties_executor.py
omby8888: i will make it as a separate fix to core, that report_run_completed with also public the message if it's an action
that would be on a seperate PR though, i'm fine with leaving it like that atm

raskinmaya (approved):

PR #3452, opened 2026-06-30
[Core] Add Redis stream consumer processing latency observability

Hadar-Vaknin:
# Description

- Add observability to the Redis stream consumer: log time_until_consumed_ms (queue-to-consume) and time_until_acked_ms (queue-to-ack) from the message queuedAt field.

- Include stream_key on Redis stream consumer log lines for easier filtering in production.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

### DIFF
--- a/port_ocean/consumers/redis_stream_consumer.py
+++ b/port_ocean/consumers/redis_stream_consumer.py
@@ -199,16 +199,20 @@ class RedisStreamConsumer(AbstractLiveEventsConsumer):
             except Exception as error:
                 logger.exception(
                     "Unexpected error in Redis stream read loop",
+                    stream_key=self._stream_key,
                     error=str(error),
                 )
 
     async def _handle_message(self, message_id: str, fields: dict[str, str]) -> None:
         start_time = time.monotonic()
         webhook_path: str | None = None
-        time_until_consumed_ms = self._time_until_consumed_ms(fields.get("queuedAt"))
+        time_until_consumed_ms = self._time_until_consumed_ms(
+            fields.get("queuedAt"), stream_key=self._stream_key
+        )
         if time_until_consumed_ms is not None:
             logger.info(
                 "Redis stream message consumed",
+                stream_key=self._stream_key,
                 message_id=message_id,
                 time_until_consumed_ms=time_until_consumed_ms,
             )
@@ -217,6 +221,7 @@ class RedisStreamConsumer(AbstractLiveEventsConsumer):
             if not raw_webhook_path:
                 logger.warning(
                     "Redis stream message missing webhookPath, acknowledging",
+                    stream_key=self._stream_key,
                     message_id=message_id,
                 )
                 return
@@ -225,6 +230,7 @@ class RedisStreamConsumer(AbstractLiveEventsConsumer):
             if webhook_path not in self._registered_paths:
                 logger.warning(
                     "No processors registered for webhookPath, acknowledging",
+                    stream_key=self._stream_key,
                     webhook_path=webhook_path,
                     message_id=message_id,
                 )
@@ -254,14 +260,18 @@ class RedisStreamConsumer(AbstractLiveEventsConsumer):
         except Exception as error:
             logger.exception(
                 "Failed to handle Redis stream message",
+                stream_key=self._stream_key,
                 message_id=message_id,
                 error=str(error),
             )
         finally:
             elapsed_ms = round((time.monotonic() - start_time) * 1000, 2)
-            time_until_acked_ms = self._time_until_consumed_ms(fields.get("queuedAt"))
+            time_until_acked_ms = self._time_until_consumed_ms(
+                fields.get("queuedAt"), stream_key=self._stream_key
+            )
             logger.info(
                 "Redis stream message processed",
+                stream_key=self._stream_key,
                 message_id=message_id,
                 webhook_path=webhook_path,
                 elapsed_ms=elapsed_ms,
@@ -271,7 +281,10 @@ class RedisStreamConsumer(AbstractLiveEventsConsumer):
 
     @staticmethod
     def _time_until_consumed_ms(
-        queued_at: str | None, *, now: datetime | None = None
+        queued_at: str | None,
+        *,
+        now: datetime | None = None,
+        stream_key: str | None = None,
     ) -> float | None:
         if not queued_at:
             return None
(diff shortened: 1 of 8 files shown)


port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -246,19 +261,68 @@ async def _handle_message(self, message_id: str, fields: dict[str, str]) -> None
...
+    def _time_until_consumed_ms(
+        queued_at: str | None,
+        *,
+        now: datetime | None = None,
+        stream_key: str | None = None,
+    ) -> float | None:
+        queued_time = RedisStreamConsumer._parse_queued_at(
+            queued_at, stream_key=stream_key
+        )
+        return RedisStreamConsumer._time_since_queued_ms(queued_time, now=now)
```
raskinmaya: This isn't used - only in tests;
`handle_message` calls `_parse_queued_at` and `_time_since_queued_ms` directly. Drop the helper; test those instead

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -246,19 +261,68 @@ async def _handle_message(self, message_id: str, fields: dict[str, str]) -> None
...
+
```
raskinmaya, line 368: Three helpers for one subtraction. At ms precision, `int(queued_at) + (now_ns - queued_ns) / 1e6` is enough. Two helpers max (parse + delta) - mainly so invalid `queuedAt` only warns once, would be better

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -198,17 +199,30 @@ async def _read_loop(self) -> None:
...
+            fields.get("queuedAt"), stream_key=self._stream_key
+        )
+        time_until_consumed_ms = self._time_since_queued_ms(queued_time)
+        if time_until_consumed_ms is not None:
+            logger.info(
+                "Redis stream message consumed",
+                stream_key=self._stream_key,
+                message_id=message_id,
+                time_until_consumed_ms=time_until_consumed_ms,
+            )
```
raskinmaya: Two `logger.info` lines per message on a hot path (both this one and [here](https://github.com/port-labs/ocean/pull/3452/changes#diff-faac178ebcde4a5a31430fa331a8499d116430a0c09771ae025f44f6f836c2daL255)). WDYT about changing this one's level to debug?

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -246,19 +261,68 @@ async def _handle_message(self, message_id: str, fields: dict[str, str]) -> None
         except Exception as error:
             logger.exception(
                 "Failed to handle Redis stream message",
+                stream_key=self._stream_key,
                 message_id=message_id,
                 error=str(error),
             )
         finally:
             elapsed_ms = round((time.monotonic() - start_time) * 1000, 2)
+            time_until_acked_ms = self._time_since_queued_ms(queued_time)
```
raskinmaya, line 315: This is computed before `await self._ack()` - it's queue -> processed, not queue -> ack. Rename it to `time_until_processed_ms`, or measure after `xack`

port_ocean/consumers/redis_stream_consumer.py
```diff
@@ -246,19 +261,68 @@ async def _handle_message(self, message_id: str, fields: dict[str, str]) -> None
...
+    def _time_since_queued_ms(
+        queued_time: datetime | None,
+        *,
+        now: datetime | None = None,
+    ) -> float | None:
+        if queued_time is None:
+            return None
+
+        reference_time = now or datetime.now(timezone.utc)
+        return round((reference_time - queued_time).total_seconds() * 1000, 2)
```
raskinmaya: Gateway stamps `queuedAt`; consumer uses its own clock. Skew can yield negative deltas. Clamp at `0` or warn on `< 0` so bad samples don't hit dashboards

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: I'm guessing u talking about the log in the finally? 
I can't see what the here link is pointing at

Anyways, I need it for metrics and dashboards (one for event consumed and one from event processed), so it can't be debug.

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍 
measured after xack

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: 👍

port_ocean/consumers/redis_stream_consumer.py
raskinmaya: Why can't you calculate the `time_until_consumed_ms`, but just put it as part of the event processed log extra fields?

raskinmaya (approved):

port_ocean/consumers/redis_stream_consumer.py
Hadar-Vaknin: I think it will be better for observability and tracing the live events with 2 logs - one for consumed and one for finished processing.

raskinmaya: I get the tracing intent, but in this flow consume and process are synchronous in the same `_handle_message` call - there's no real window where a second log helps debugging that one field on the processed log wouldn't.

Also, live event drop rates are already very high. Your feature is meant to fix that, so we should expect more events actually getting processed, not fewer. Doubling log volume (2× info per message on a hot path) feels like the wrong tradeoff when `time_until_consumed_ms` is already computed at the top of the function - it's a one-line add to the processed log's extra fields

For dashboards: one log with `time_until_consumed_ms`, `elapsed_ms`, and `time_until_acked_ms` gives us queue-wait, processing time, and end-to-end without 2× ingest cost

PR #3453, opened 2026-06-30
[Core] Prevent action run starvation from high-volume actions

omby8888:
## Summary
- Track queued run counts per action type in the execution manager (increment on enqueue, decrement on dequeue)
- Add `max_runs_buffer_util_pct_per_action` config — a percentage of `runs_buffer_high_watermark` — to cap how much buffer any single action type may occupy
- Pass saturated action types as `excludeActionTypes` on claim-pending requests (action runs + workflow node runs)

Port task: https://app.getport.io/taskEntity?identifier=task_thg0xh

## Test plan
- [ ] Set `OCEAN__ACTIONS_PROCESSOR__MAX_RUNS_BUFFER_UTIL_PCT_PER_ACTION` and verify saturated action types are excluded from claim-pending
- [ ] Verify other action types continue to be claimed when one type fills its quota
- [ ] Confirm Port claim-pending APIs honor `excludeActionTypes` (requires backend support)


port_ocean/core/models.py
```diff
@@ -196,11 +196,16 @@ class ActionRun(BaseModel):
     id: str
     status: RunStatus
     payload: IntegrationActionInvocationPayload
+    actionIdentifier: str
```
raskinmaya: Making this field required is contract breaking.
This field is parsed from two different Port endpoints:

- `claim_pending_action_runs` -> `ActionRun.parse_obj(...)`
- `get_run_by_external_id` -> `ActionRun.parse_obj(...)`, reached from `find_run_by_external_id`

Making it required means: if we don't pass `actionIdentifier` on these payloads, `parse_obj` raises `ValidationError`. For claim-pending that lands in the broad `except Exception` in `_poll_action_runs` -> an infinite failing re-poll loop, i.e. actions stop working entirely. For `get_run_by_external_id`, `find_run_by_external_id` swallows the exception and returns `None`, so the GitHub/GitLab webhook processors would silently stop resolving action runs by external id. Instead, you can make it `actionIdentifier: str = ""` and degrade gracefully

port_ocean/core/handlers/actions/execution_manager.py
```diff
@@ -262,6 +315,7 @@ async def _add_run_to_queue(
             if await queue.size() == 0:
                 await self._active_sources.put(queue_name)
             self._deduplication_set.add(run.id)
+            self._increment_action_identifier_queue_count(run.action_identifier)
```
raskinmaya: We increment `_action_identifier_queue_counts` for every run via `run.action_identifier`, and for `WorkflowNodeRun` that resolves to `nodeIdentifier`. `_get_saturated_action_identifiers` returns one flat list that is passed as `exclude_action_identifiers` only into `claim_pending_action_runs`.
Two problems:
- A saturated workflow node id is sent as exclude to the action claim-pending endpoint - wrong namespace. If a `nodeIdentifier` ever collides with an `actionIdentifier`, you'd wrongly exclude an action, but TBH I'm not too worried about this
- The poll alternates wf/action but `exclude` is never forwarded to `claim_pending_wf_node_runs`. So workflow nodes can never be throttled even though they consume the same shared buffer. I think either track the two namespaces separately, or explicitly scope this feature to action runs and don't count WF nodes

port_ocean/core/handlers/actions/execution_manager.py
```diff
@@ -351,15 +410,20 @@ async def _handle_partition_queue_once(self, partition_name: str) -> None:
                     run = await asyncio.wait_for(
                         queue.get(), timeout=QUEUE_GET_TIMEOUT_SECONDS
                     )
+                    got_run = True
                     if run.id in self._deduplication_set:
                         self._deduplication_set.remove(run.id)
+                        self._decrement_action_identifier_queue_count(
```
raskinmaya: Increment is unconditional on enqueue; decrement only happens if `run.id` in `self._deduplication_set` at dequeue. If those two sets ever desync (e.g. a run dequeued whose id was already removed), the count never decrements, the identifier stays `>= limit` forever, and that action is permanently excluded - the opposite of the goal. The count should track enqueue/dequeue directly and symmetrically

port_ocean/config/settings.py
```diff
@@ -108,10 +108,47 @@ class StreamingSettings(BaseOceanModel, extra=Extra.allow):
...
+            "becoming reclaimable (milliseconds)."
+        ),
+    )
+    poll_check_interval_seconds: int = Field(
+        default=10,
+        ge=1,
+        description="Seconds between claim-pending polling attempts.",
+    )
+    workers_count: int = Field(
+        default=3,
```
raskinmaya, line 135: Is this default change coordinated with platform?

port_ocean/clients/port/mixins/actions_and_workflow_runs.py
```diff
@@ -39,19 +39,25 @@ def _wf_node_completion_patch(
         return patch
 
     async def claim_pending_runs(
-        self, limit: int, visibility_timeout_ms: int
+        self,
+        limit: int,
+        visibility_timeout_ms: int,
+        exclude_action_identifiers: list[str] = [],
```
raskinmaya: I think a null value would be better here as a default if not passed

port_ocean/clients/port/mixins/actions.py
```diff
@@ -29,16 +29,22 @@ async def create_action(
         handle_port_status_code(response, should_log=should_log)
 
     async def claim_pending_action_runs(
-        self, limit: int, visibility_timeout_ms: int
+        self,
+        limit: int,
+        visibility_timeout_ms: int,
+        exclude_action_identifiers: list[str] = [],
     ) -> list[ActionRun]:
+        body: dict[str, object] = {
```
raskinmaya: The rest of this client use `dict[str, Any]`. Maybe we can align it for consistency?

port_ocean/core/handlers/actions/execution_manager.py
raskinmaya, line 1: Please update the docstrings in this file, there's some stale info here

port_ocean/core/models.py
```diff
@@ -196,11 +196,16 @@ class ActionRun(BaseModel):
...
 
     @property
     def action_type(self) -> str:
         return self.payload.integrationActionType
 
+    @property
+    def action_identifier(self) -> str:
+        return self.actionIdentifier
```
raskinmaya: This adds `action_identifier` that just returns `self.actionIdentifier`. It's justifiable for the `ActionRun | WorkflowNodeRun` duck-typed interface (that mirrors `action_type`), but consider a pydantic `Field(alias=...)` so there aren't two public names for one value

port_ocean/config/settings.py
omby8888: no, the current value utilzes super low amount of the resources of this pod, so we can safely increase it

port_ocean/core/models.py
```diff
@@ -205,43 +272,67 @@ def action_type(self) -> str:
...
-class ClaimedWorkflowNodeRun(WorkflowNodeRun):
-    config: dict[str, Any]
-    result: WorkflowNodeRunResult | None = None
+        return self.integration_config.integrationActionExecutionProperties
 
     @property
-    def action_type(self) -> str:
-        return self.config["integrationInvocationType"]
+    def buffer_utilization_key(self) -> str:
+        return f"{self.id}-{self.action_type}"
```
omby8888: This needs to be finalized by workflows team

raskinmaya: let's leave a TODO comment in the code to change that later

port_ocean/core/models.py
raskinmaya: I see the shape changed to required `action: Action`. This should be safe in this context, right?

port_ocean/core/models.py
raskinmaya: Still present (lines 263-265), is that intentional?

raskinmaya (approved):

port_ocean/clients/port/mixins/actions_and_workflow_runs.py
```diff
@@ -39,38 +34,52 @@ def _wf_node_completion_patch(
...
-        runs: list[ActionRun | WorkflowNodeRun]
-        if self._poll_wf_node:
-            runs = list(
-                await self.claim_pending_wf_node_runs(
-                    limit=limit, visibility_timeout_ms=visibility_timeout_ms
+        self,
+        limit: int,
+        visibility_timeout_ms: int,
+        exclude_action_identifiers: list[str] | None = None,
+        exclude_workflow_invocation_types: list[str] | None = None,
```
Amichai-Skaliter-Port: Why `exclude_workflow_invocation_types` and not `exclude_wf_nodes_uid` for example?

port_ocean/clients/port/mixins/actions_and_workflow_runs.py
```diff
@@ -39,38 +34,52 @@ def _wf_node_completion_patch(
...
+        runs: list[IntegrationRun] = []
+        for workflow in (self._claim_workflow_first, not self._claim_workflow_first):
+            if remaining <= 0:
+                break
+            if workflow:
+                claimed: list[IntegrationRun] = list(
+                    await self.claim_pending_wf_node_runs(
+                        limit=remaining,
+                        visibility_timeout_ms=visibility_timeout_ms,
+                        exclude_invocation_types=exclude_workflow_invocation_types,
```
Amichai-Skaliter-Port: Same as above - why use the term `invocation_Types`?

port_ocean/clients/port/mixins/workflow_nodes.py
```diff
@@ -19,22 +18,28 @@ def __init__(self, auth: PortAuthentication, client: httpx.AsyncClient):
         self.client = client
 
     async def claim_pending_wf_node_runs(
-        self, limit: int, visibility_timeout_ms: int
-    ) -> list[ClaimedWorkflowNodeRun]:
+        self,
+        limit: int,
+        visibility_timeout_ms: int,
+        exclude_invocation_types: list[str] | None = None,
```
Amichai-Skaliter-Port: why "invocation_types"?

PR #3457, opened 2026-07-01
[Integration][Github] Implement property reduction GitHub

melodyogonna:
# Description

What - Strip heavy fields from the Graphql paginated API if the requests continue to timeout.

Why - To prevent data loss we'd rather return fewer fields than skip a pull request. So in the case that we can't fetch all the fields in a pull request, let's take out potentially heavy relations and try again.

How - Implemented with the concept of fallback queries, which allows us to specify progressively lighter queries.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Changes resync behavior when queries time out—some PRs may sync with fewer GraphQL fields—and pagination retry logic is more complex, though scoped per page with tests.
> 
> **Overview**
> Adds a **last-resort recovery path** when GitHub GraphQL pull-request list pagination still fails after the retry transport shrinks page size to the floor. Gateway timeouts (**502/504/499**) are now distinguished from a generic **500**: only the gateway codes trigger switching to a lighter query; **500** still gets page-size backoff only.
> 
> `send_paginated_request` accepts optional **`fallback_queries`**. Each page is fetched via `_fetch_page`, which tries the primary query at normal page size, then—on an “over-budget” gateway error—retries that **same page** with each fallback at `MIN_GRAPHQL_PAGE_SIZE`. The next cursor page **resets** to the full query and page size so one heavy page does not degrade the whole sync.
> 
> PR list export wires in a single fallback that omits **`EXPENSIVE_PR_GRAPHQL_FIELDS`** (reviews, assignees, labels, etc.) via new `extra_excluded_fields` on query builders, skipped when the user already excludes those fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f88bfccc3afeeeb8c7019f37b720e135955680c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/github/github/helpers/gql_queries.py
```diff
@@ -184,7 +186,28 @@
...
-def generate_pr_fields(options: PullRequestGraphQLOptions) -> str:
+# PullRequest fields whose per-node resolution is expensive on large orgs
+# (team/membership expansion, actor visibility, check rollups). Stripped as a
+# last resort when a query keeps exceeding GitHub's GraphQL execution budget.
+EXPENSIVE_PR_GRAPHQL_FIELDS = [
+    "reviewRequests",
+    "reviews",
+    "assignees",
+    "statusCheckRollup",
+]
```
shalev007, line 199: Good approach but we're not sure which field is expensive or which field is expensive for which user are those all of the fields that are from different nodes?
if not maybe we should add all of those fields to this const and keep the pull-request graphql request bare naked

integrations/github/github/core/exporters/pull_request_exporter/core.py
```diff
@@ -242,6 +243,26 @@ async def get_paginated_resources[
...
+        """
+        full = generate_list_pull_requests_gql(
+            pr_gql_options, order_by_field=order_by_field
+        )
+        stripped = generate_list_pull_requests_gql(
+            pr_gql_options,
+            order_by_field=order_by_field,
+            extra_excluded_fields=EXPENSIVE_PR_GRAPHQL_FIELDS,
+        )
+        return [stripped] if stripped != full else []
```
shalev007: why is the `full` even necessary if we don't use it at all?
we either return `[stripped] or []`

integrations/github/github/helpers/gql_queries.py
melodyogonna: Added some more fields.

shalev007 (approved):
Approved!

PR #3463, opened 2026-07-01
[Integration] Add Claude Managed Agents integration

talsabagport:
# Description

What - Adds a new Ocean integration for Claude Managed Agents (Anthropic's beta agent platform): resync support for `agent`, `environment`, `session`, `vault`, `memory-store`, and `skill`, live-events (webhooks) to keep `session`/`vault` entities fresh and report `trigger_agent` run status back to Port, and two self-service actions (`create_agent`, `trigger_agent`) invokable from Port workflows.

Why - Lets Port users manage and invoke Claude Managed Agents directly from their software catalog and workflows, and keep agent/session state in sync without manually polling the Anthropic API.

How - Wraps the official Anthropic Python SDK (Managed Agents beta) in a thin async client (`AnthropicClient`), with per-resource `include_archived` selectors, a dedicated `OceanAsyncClient` for SaaS IP-blocking/SSL, and proactive throttling on the create-endpoints RPM limit via response headers. Webhook processors handle `session`/`vault` upserts and correlate `trigger_agent` webhook events back to the originating action run (using session events to detect idle/error state and extract the agent's response).

## Type of change

- [ ] If new resource kind is added or updated in the integration, add example raw data, mapping and expected result to the `examples` folder in the integration directory. (covered via the linked docs PR - `_cma_*` payload/mapping/blueprint files - no separate `examples` folder added)
- [ ] If resource kind is updated, run the integration with the example data and check if the expected result is achieved (N/A - only added kinds)
- [x] Docs PR link [here](https://github.com/port-labs/port-docs/pull/4287)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [ ] Support Multi account (N/A - single Anthropic API key/org per integration instance)

## Screenshots

image
image
image

## API Documentation

https://platform.claude.com/docs/en/api/overview
https://platform.claude.com/docs/en/managed-agents/overview


integrations/claude-managed-agents/main.py
```diff
@@ -0,0 +1,100 @@
...
+async def on_resync_skills(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+    exporter = create_skills_exporter()
+    async for page in exporter.get_paginated_resources():
+        if page:
+            yield page
+
+
+@ocean.on_start()
+async def on_start() -> None:
+    # Anthropic webhooks are configured in the Claude Console (there is no
```
omby8888, line 93: just make that clear in the Docs

integrations/claude-managed-agents/actions/abstract_executor.py
```diff
@@ -0,0 +1,60 @@
...
+# off until the create rate limit window resets. A percentage (rather than an
+# absolute count) keeps this correct regardless of the account's RPM tier.
+CREATE_RATE_LIMIT_HEADROOM_RATIO = 0.1
+
+
+class AbstractAnthropicExecutor(AbstractExecutor):
+    """Base executor for Claude Managed Agents actions."""
+
+    WEBHOOK_PROCESSOR_CLASS: Optional[Type[AbstractWebhookProcessor]] = None
+    WEBHOOK_PATH: str = ""
```
omby8888: remove those if not being used

integrations/claude-managed-agents/actions/create_agent_executor.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    ACTION_NAME = "create_agent"
+
+    async def execute(self, run: ActionRun | WorkflowNodeRun) -> None:
+        props = run.execution_properties
+        name = props.get("name")
+        model = props.get("model")
+        if not (name and model):
+            raise ValueError("name and model are required")
```
omby8888: use `InvalidActionParametersException` instead in all places

integrations/claude-managed-agents/actions/create_agent_executor.py
```diff
@@ -0,0 +1,54 @@
...
+        try:
+            agent = await self.client.create_agent(
+                name=name, model=model, system=system, extra=extra
+            )
+        except Exception as error:
+            logger.error(f"Failed to create Claude agent for run {run.id}: {error}")
+            await ocean.port_client.report_run_completed(
+                run, False, f"Failed to create agent: {error}"
+            )
+            return
```
omby8888: to be aligned with other actions, let's raise `ActionExecutionError` instead of report the run as completed.
this is the execution manager's responsibility to do that

integrations/claude-managed-agents/webhook_processors/abstract_webhook_processor.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @staticmethod
+    def get_event_type(payload: EventPayload) -> str:
+        data = payload.get("data") or {}
+        return data.get("type", "")
+
+    @staticmethod
+    def get_resource_id(payload: EventPayload) -> str:
+        data = payload.get("data") or {}
+        return str(data.get("id") or "")
```
omby8888, line 53: those fallbacks should be validated in validate_payload so u can safely remove those from here

integrations/claude-managed-agents/webhook_processors/trigger_agent_webhook_processor.py
```diff
@@ -0,0 +1,382 @@
...
+            return
+
+        await ocean.port_client.report_run_completed(run, success)
+
+    @classmethod
+    def get_processor_type(cls) -> WebhookProcessorType:
+        return WebhookProcessorType.ACTION
+
+    async def should_process_event(self, event: WebhookEvent) -> bool:
+        return self.get_event_type(event.payload) in {
```
omby8888: use `_INTERACTION_EVENT_TYPES`

integrations/claude-managed-agents/actions/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    """Build a best-effort console link to the session for run output.
+
+    Falls back to the console root if the workspace id wasn't captured (e.g.
+    the anthropic-workspace-id header was unexpectedly absent).
+    """
+    if not workspace_id:
+        return "https://platform.claude.com/"
+    return (
+        f"https://platform.claude.com/workspaces/{workspace_id}/sessions/{session_id}"
+    )
```
omby8888: are those url correlated to the `anthropicApiHost` param?

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,130 @@
...
+        props = run.execution_properties
+        agent_id = props.get("agentId")
+        environment_id = props.get("environmentId")
+        prompt = props.get("prompt")
+        session_id = props.get("sessionId")
+        config = props.get("config")
+        if config is None:
+            config = {}
+        elif not isinstance(config, dict):
+            raise ValueError("config must be an object")
```
omby8888: use `InvalidActionParametersException`

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if not (agent_id and prompt):
+            raise ValueError("agentId and prompt are required")
+
+        if session_id:
+            session = await self._ensure_session_continuable(session_id)
+            resolved_session_id = session_id
+        else:
+            if not environment_id:
+                raise ValueError(
```
omby8888: use `raise ActionExecutionError` instead

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,130 @@
...
+                )
+            api_config = await normalize_session_config(config)
+            session = await self.client.create_session(
+                agent_id,
+                environment_id,
+                extra=api_config or None,
+            )
+            resolved_session_id = session.get("id")
+            if not resolved_session_id or not isinstance(resolved_session_id, str):
+                raise ValueError("Session was created but no session id was returned")
```
omby8888: same

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,130 @@
...
+            build_session_link(self.client.get_workspace_id(), resolved_session_id),
+            external_id,
+            extra_output={
+                "sessionId": resolved_session_id,
+                "userMessageEventId": user_message.id,
+            },
+        )
+
+        if not props.get("reportSessionStatus", True):
+            await ocean.port_client.report_run_completed(
```
omby8888, line 140: Since live events requires manual setup for this to work, do we sure that we want this parameter rather than closing the run?

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,130 @@
...
+
+    New sessions for the same agent are executed sequentially via the agent id
+    partition key; continuations are serialized per session id.
+    """
+
+    ACTION_NAME = "trigger_agent"
+    WEBHOOK_PROCESSOR_CLASS = TriggerAgentWebhookProcessor
+    WEBHOOK_PATH = WEBHOOK_PATH
+
+    async def _get_partition_key(self, run: ActionRun | WorkflowNodeRun) -> str | None:
```
omby8888: u sure that it required to be sequential?

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,130 @@
...
+        session_id = run.execution_properties.get("sessionId")
+        if session_id:
+            return session_id
+        return run.execution_properties.get("agentId")
+
+    async def _ensure_session_continuable(self, session_id: str) -> dict[str, Any]:
+        session = await self.client.get_session(session_id)
+        status = session.get("status")
+        if status != "idle":
+            raise ValueError(
```
omby8888: use `raise ActionExecutionError`

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,130 @@
...
+            session_id,
+            types=["session.status_idle"],
+            order="desc",
+            limit=1,
+        ):
+            idle_event = batch[0]
+            if isinstance(idle_event, BetaManagedAgentsSessionStatusIdleEvent):
+                stop_reason = idle_event.stop_reason
+                if isinstance(stop_reason, BetaManagedAgentsSessionRequiresAction):
+                    raise ValueError(
```
omby8888: same

integrations/claude-managed-agents/actions/abstract_executor.py
```diff
@@ -0,0 +1,60 @@
...
+        loaded (the action execution manager does not provide one), so we open a
+        fresh context here. Catalog reflection must never fail the action run, so
+        failures are logged and swallowed.
+        """
+        try:
+            async with event_context(EventType.HTTP_REQUEST, trigger_type="machine"):
+                await ocean.integration.port_app_config_handler.get_port_app_config(
+                    use_cache=False
+                )
+                await ocean.register_raw(kind, [raw])
```
omby8888, line 97: not sure that this wrapper is needed

integrations/claude-managed-agents/actions/abstract_executor.py
```diff
@@ -0,0 +1,60 @@
...
+        except Exception as error:
+            logger.warning(
+                f"Failed to upsert {kind} into the catalog (continuing): {error}"
+            )
```
omby8888: maybe add a warn run log for that as well

integrations/claude-managed-agents/.port/spec.yaml
```diff
@@ -0,0 +1,105 @@
...
+  - name: webhookSigningSecret
+    type: string
+    required: false
+    sensitive: true
+    description: The signing secret from the Claude Console webhook configuration. Required only to enable live events (webhooks). Refer <a href="https://platform.claude.com/docs/en/api/typescript/beta/webhooks" target="_blank">Anthropic webhooks documentation</a>.
+  - name: githubAuthorizationToken
+    type: string
+    required: false
+    sensitive: true
+    description: Optional GitHub personal access token injected into github_repository session resources when authorization_token is omitted. Configure this when workflow forms collect repository URLs only.
```
omby8888, line 54: can u further alaborate what this is for? i'm missing some context regarding why this parameter needed

integrations/claude-managed-agents/actions/session_config.py
omby8888, line 1: this can sit in utils imo

integrations/claude-managed-agents/clients/anthropic_client.py
```diff
@@ -0,0 +1,302 @@
...
+        # A dedicated OceanAsyncClient gets us SaaS IP-blocking and consistent SSL
+        # verification. Ocean's own retry layer is disabled (max_attempts=0) since
+        # the Anthropic SDK already retries internally;
+        self._http_client = OceanAsyncClient(
+            timeout=ocean.config.client_timeout,
+            retry_config=RetryConfig(max_attempts=0),
+        )
+        default_headers = (
+            {"anthropic-version": anthropic_version} if anthropic_version else None
+        )
```
omby8888: could be defined directly in the _client as there is the only usage of those

integrations/claude-managed-agents/clients/anthropic_client.py
```diff
@@ -0,0 +1,302 @@
...
+_WORKSPACE_ID_HEADER = "anthropic-workspace-id"
+
+
+def _serialize(obj: Any) -> dict[str, Any]:
+    """Convert an SDK model instance into a plain JSON-serialisable dict."""
+    if hasattr(obj, "to_dict"):
+        return obj.to_dict(mode="json")
+    if hasattr(obj, "model_dump"):
+        return obj.model_dump(mode="json")
+    return dict(obj)
```
omby8888: afaik the antropic package uses pydantic v2, so `model_dump` should be safe to use. i dont think this function is required

integrations/claude-managed-agents/clients/anthropic_client.py
```diff
@@ -0,0 +1,302 @@
...
+        if created_at_gt is not None:
+            list_kwargs["created_at_gt"] = created_at_gt
+        if created_at_gte is not None:
+            list_kwargs["created_at_gte"] = created_at_gte
+        if created_at_lt is not None:
+            list_kwargs["created_at_lt"] = created_at_lt
+        if created_at_lte is not None:
+            list_kwargs["created_at_lte"] = created_at_lte
+        if limit is not None:
+            list_kwargs["limit"] = limit
```
omby8888: better to export this to a pydantic model which provides cleaner way to do that

integrations/claude-managed-agents/clients/anthropic_client.py
omby8888, line 1: i think in general that each exporter class should implement it's specific logic.
this client should export the ability to make requests easily through antropic's sdk while considering rate limits

integrations/claude-managed-agents/clients/anthropic_client.py
```diff
@@ -0,0 +1,302 @@
...
+
+        logger.info(
+            f"Creating Claude session for agent '{agent_id}' in environment '{environment_id}'"
+        )
+        try:
+            raw = await self._client.beta.sessions.with_raw_response.create(**payload)
+        except anthropic.APIStatusError as error:
+            self._capture_create_rate_limit_headers(error.response.headers)
+            raise
+        self._capture_create_rate_limit_headers(raw.headers)
```
omby8888: idk exactly how their SDK handle rate limits, but since i expect it to be encapsulated by them (retry with jitter...) should we expect this path to lead to rate limits?
also atm we capture the rate limit only in specific requests, this could lead to some inconsistency

integrations/claude-managed-agents/clients/anthropic_client.py
```diff
@@ -0,0 +1,302 @@
...
+        return event
+
+    def unwrap_webhook(
+        self, payload: str, headers: Mapping[str, str]
+    ) -> UnwrapWebhookEvent:
+        if not self._webhook_signing_secret:
+            raise ValueError("Webhook signing secret is not configured")
+        return self._client.beta.webhooks.unwrap(
+            payload, headers=headers, key=self._webhook_signing_secret
+        )
```
omby8888: let's have specific error classes for those

integrations/claude-managed-agents/webhook_processors/trigger_agent_webhook_processor.py
omby8888, line 1: There's a lot logic in this file that should be moved to ocean core, as it's not specific to this webhook processor

omby8888 (changes requested):

integrations/claude-managed-agents/main.py
talsabagport: You can see a draft for it [here](https://pr-4287.d2ngvl90zqbob8.amplifyapp.com/build-your-software-catalog/sync-data-to-catalog/ai-agents/claude-managed-agents?method=hosted&step=choose-method#webhook-configuration)

integrations/claude-managed-agents/webhook_processors/abstract_webhook_processor.py
talsabagport: `get_event_type` is also called from each processor's `should_process_event` (`vault_webhook_processor.py`, `session_webhook_processor.py`, `trigger_agent_webhook_processor.py`), which runs before validate_payload in the framework's pipeline. So with your suggestion, a genuinely malformed payload hitting the webhook endpoint (missing data/type) will raise a `KeyError` there instead of quietly falling through as "".

integrations/claude-managed-agents/webhook_processors/trigger_agent_webhook_processor.py
talsabagport: I did not literally reuse _INTERACTION_EVENT_TYPES as you suggested, because that constant holds a different, non-overlapping set of values used for a different purpose:

- _INTERACTION_EVENT_TYPES = ["session.error", "session.status_idle", "agent.message"] - types passed to get_session_events() when fetching interaction detail (errors/idle transitions/agent messages) from the Session Events API.
- The set previously inline in should_process_event - {"session.idled", "session.status_idled", "session.status_terminated"} - are top-level webhook delivery types (data.type) that mark a session as terminal.

I extracted the literal into its own clearly-named, documented constant, _TERMINAL_SESSION_WEBHOOK_TYPES, right next to _INTERACTION_EVENT_TYPES.

integrations/claude-managed-agents/actions/utils.py
talsabagport: No, I will create a new param

integrations/claude-managed-agents/actions/trigger_agent_executor.py
talsabagport: I'll make it required for the same session, and not for the same agent (can run multiple sessions in parallel for same agent or different agents)

integrations/claude-managed-agents/actions/abstract_executor.py
talsabagport: Traced through Ocean core:

- PortAppConfigHandler.get_port_app_config() sets event.port_app_config = ... (port_ocean/core/handlers/port_app_config/base.py:81).
- sync_raw.py's _register_resource_raw() (called by register_raw) reads event.event_type directly (sync_raw.py:317).
- event is a LocalProxy that raises EventContextNotFoundError the moment any attribute is accessed with no active event context pushed (port_ocean/context/event.py:114-127).

Since the action execution manager never opens an event context around executor.execute(), both the event_context(...) wrapper and the explicit get_port_app_config(use_cache=False) call are required, or register_raw would crash with EventContextNotFoundError.

integrations/claude-managed-agents/.port/spec.yaml
talsabagport: When you want the agent to checkout a github repo, open a new branch, make changes and push it, you need a github token for that.

integrations/claude-managed-agents/webhook_processors/trigger_agent_webhook_processor.py
talsabagport: I agree, as we discussed we need to extract:

- _complete_port_run (trigger_agent_webhook_processor.py:128-152) - reimplements "complete a run" dispatch between ActionRun/WorkflowNodeRun, purely to add support for merging extra_output into the workflow node's output. The core port_client.report_run_completed(run, success) already does the dispatch but has no extra_output param.

- _post_run_logs + the _WF_LOG_LEVEL mapping (trigger_agent_webhook_processor.py:369-384, 52-61) - reimplements the exact same "WARNING" → "WARN" level-name translation that already lives in port_ocean's post_run_log (port_ocean/clients/port/mixins/actions_and_workflow_runs.py:66-84), just to batch multiple log entries into a single patch_wf_node_run call instead of one call per line — a capability the core post_run_log doesn't expose (it only takes one message at a time).

integrations/claude-managed-agents/actions/abstract_executor.py
talsabagport: Currently it's not possible, a fix should happen in the Ocean core:
```
"port_ocean/core/handlers/actions/execution_manager.py", line 115, in register_executor
    webhook_processor_cls = executor.WEBHOOK_PROCESSOR_CLASS
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'CreateAgentExecutor' object has no attribute 'WEBHOOK_PROCESSOR_CLASS'
Exception ignored in atexit callback: <function shutdown at 0x105cc5120>
```

integrations/claude-managed-agents/.port/spec.yaml
```diff
@@ -0,0 +1,111 @@
...
+        description: Model identifier, for example claude-opus-4-8
+        required: true
+      - name: systemPrompt
+        title: System prompt
+        type: string
+        description: System prompt for the agent
+      - name: config
+        title: Additional configuration
+        type: jqObject
+        description: 'Optional advanced agent configuration passed through to the Anthropic API (for example tools, mcp_servers, skills, metadata, description). Reference the action trigger data with double brackets, for example: {{ .trigger.by.user.email }}.'
```
omby8888, line 79: pls indicate `required=false` specifically

integrations/claude-managed-agents/.port/spec.yaml
```diff
@@ -0,0 +1,111 @@
...
+        required: false
+      - name: prompt
+        title: Prompt
+        type: string
+        description: The user message sent to the agent
+        required: true
+      - name: config
+        title: Additional configuration
+        type: jqObject
+        description: 'Optional session configuration passed through to the Anthropic API when starting a new session (for example title, metadata, vault_ids, resources). Ignored when continuing an existing session. Reference the action trigger data with double brackets, for example: {{ .trigger.by.user.email }}.'
```
omby8888, line 108: same here

integrations/claude-managed-agents/actions/create_agent_executor.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    ACTION_NAME = "create_agent"
+
+    async def execute(self, run: ActionRun | WorkflowNodeRun) -> None:
+        props = run.execution_properties
+        name = props.get("name")
+        model = props.get("model")
+        if not (name and model):
+            raise InvalidActionParametersException("name and model are required")
```
omby8888, line 26:
```suggestion
            raise MissingExecutionPropertyError("name and model are required")
```

integrations/claude-managed-agents/clients/anthropic_client.py
```diff
@@ -0,0 +1,357 @@
...
+                self._set_rate_limit_info(pool, None)
+            else:
+                self._set_rate_limit_info(
+                    pool, replace(info, remaining=info.remaining - 1)
+                )
+
+    async def _on_response(self, response: httpx.Response) -> None:
+        workspace_id = response.headers.get(_WORKSPACE_ID_HEADER)
+        if workspace_id:
+            self._workspace_id = workspace_id
```
omby8888, line 197: what if some func tries to get the workspace_id before the client made a single request?

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,142 @@
...
+        prompt = props.get("prompt")
+        session_id = props.get("sessionId")
+        config = props.get("config")
+        if config is None:
+            config = {}
+        elif not isinstance(config, dict):
+            raise InvalidActionParametersException("config must be an object")
+
+        if not (agent_id and prompt):
+            raise InvalidActionParametersException("agentId and prompt are required")
```
omby8888, line 91:
```suggestion
            raise MissingExecutionPropertyError("agentId and prompt are required")
```

integrations/claude-managed-agents/actions/trigger_agent_executor.py
```diff
@@ -0,0 +1,142 @@
...
+
+        if session_id:
+            session = await self._ensure_session_continuable(session_id)
+            resolved_session_id = session_id
+        else:
+            if not environment_id:
+                raise InvalidActionParametersException(
```
omby8888, line 98:
```suggestion
                raise MissingExecutionPropertyError(
```

integrations/claude-managed-agents/clients/anthropic_client.py
talsabagport: It's not so bad because the type is optional (can be None). We use it now just to build the session link, and we have a fallback. But if you invoke the session, you should have it, so it shouldn't happen

omby8888 (approved):
Awesome work!!!

PR #3471, opened 2026-07-07
[Integration][GitHub] Use workflow run ID on dispatch for concurrent action runs

omby8888:
## Summary
- Use GitHub's `return_run_details` dispatch API to get the `workflow_run_id` directly instead of polling the workflow runs list
- Fetch the dispatched run via `RestWorkflowRunExporter` and track it with a unique external ID, enabling concurrent `dispatch_workflow` action executions
- Remove workflow-based partition key and polling logic; raise `ActionExecutionError` for expected failures
- Bump GitHub integration to `6.2.3` and add unit tests for `DispatchWorkflowExecutor`

## Test plan
- [x] `poetry run pytest tests/github/actions/test_dispatch_workflow_executor.py -n 0`
- [ ] Trigger multiple concurrent `dispatch_workflow` actions against the same workflow and verify each Port run tracks the correct GitHub workflow run
- [ ] Verify webhook completion still updates the correct Port action run via external ID

### DIFF
--- a/integrations/github/github/actions/dispatch_workflow_executor.py
+++ b/integrations/github/github/actions/dispatch_workflow_executor.py
@@ -35,22 +35,21 @@ class DispatchWorkflowExecutor(AbstractGithubExecutor):
     This executor implements the Port action for triggering GitHub Actions workflows.
     It supports:
     - Dispatching workflows with custom inputs
-    - Tracking workflow execution status
+    - Tracking workflow execution status via the returned workflow run ID
     - Reporting workflow completion back to Port
     - Rate limit handling for GitHub API
     - Webhook processing for async status updates
 
-    The executor uses workflow names as partition keys to ensure sequential
-    execution of the same workflow, which is necessary for proper run tracking.
-    It identifies workflow runs by finding the one closest to the trigger time
-    and recording its ID.
+    On dispatch, GitHub returns a workflow run ID when `return_run_details` is set.
+    The executor fetches that run and builds a unique external ID from it, so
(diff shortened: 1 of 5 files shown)


integrations/github/github/actions/dispatch_workflow_executor.py
```diff
@@ -220,30 +169,41 @@ async def execute(self, run: ActionRun | WorkflowNodeRun) -> None:
...
+
+            workflow_run = await self._workflow_run_exporter.get_resource(
+                SingleWorkflowRunOptions(
+                    organization=organization,
+                    repo_name=repo,
+                    run_id=workflow_run_id,
+                )
             )
+            if not workflow_run:
+                raise ActionExecutionError(
```
Hadar-Vaknin, line 190: IMO, a dedicated Exception will be better than the general ActionExecutionError, Something Like WorkflowNotFound

integrations/github/github/actions/dispatch_workflow_executor.py
```diff
@@ -1,39 +1,31 @@
-import asyncio
-from datetime import datetime, timezone
 import json
```
Hadar-Vaknin, line 1: The  docstring on DispatchWorkflowExecutor describes the old behavior, change it to the new one.

integrations/github/github/actions/dispatch_workflow_executor.py
omby8888: I think no need for that

Hadar-Vaknin (approved):

PR #3474, opened 2026-07-07
[Integration][SonarQube] Skip Inaccessible Projects During Enrichment 

emekanwaoma:
# Description

What - Fixed SonarQube project enrichment to handle projects that return no main branch data without crashing the resync.

Why - SonarQube can return stale or inaccessible projects from discovery, then return 403 for branch data. The integration ignored the 403, received an empty branch list, and failed with `list index out of range` when reading the main branch.

How - Set `__branch` to an empty object when no main branch is returned, allowing mappings/selectors to handle the incomplete branch data. Also avoids emitting empty project payloads from project webhooks.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)


oiadebayo (approved):
LGTM

omby8888 (approved):

PR #3476, opened 2026-07-07
[Integration][GitHub] Support member ingestion for GitHub Enterprise teams

jasonnwakaeze-debug:
# Description

What - Enterprise teams (slug prefix ent:*) now have their members populated in Port. Previously they synced as empty shells

Why - GitHub's GraphQL organization.team(slug) resolver silently returns null for enterprise-level teams — they live at the enterprise hierarchy, not the org level, so the resolver can't find them. Customers using  GitHub Enterprise Managed Users (EMU) saw their ent:* teams in Port with no members, despite those teams having members in GitHub.

How -
  - Added enrich_enterprise_teams_with_members on RestTeamExport                                                                                                               
  - Wired in both the resync path (main.py, after _enrich_team_with_extras) and the webhook path (team_webhook_processor.py)
  - SAML email enrichment (enrich_members_with_saml_email) is then called on enterprise team members using the existing cached org-level SAML path, same as regular teams                                            
  - Detection uses slug.startswith("ent:") rather than type == "enterprise" — slug is guaranteed present in both list and single-team GET responses, type field presence on single-team GET was unverified           
   
## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots
Screenshot 2026-07-07 at 12 43 04

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Adds extra REST calls per enterprise team when members are enabled and touches live webhook upserts; scope is limited to `ent:*` teams and existing member/SAML flows.
> 
> **Overview**
> **Enterprise teams** (`ent:*` slugs) can now get **members** in Port when team sync has `members` enabled. GraphQL often returns null for those teams, so the integration **falls back to the org team members REST API** and writes results as `members.nodes`, matching the GraphQL shape.
> 
> `RestTeamExporter.enrich_enterprise_teams_with_members` only runs for `ent:*` slugs; org teams are unchanged. It is wired into **team resync** (after `_enrich_team_with_extras`) and **team webhooks**, with early exits when enrichment yields nothing. For enterprise teams, **SAML email enrichment** still runs on the REST member list when configured.
> 
> Release **6.2.15** and changelog/tests cover the new path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7968f2a9d49b3eb44004fb11ca90599ab06a325. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/github/github/webhook/webhook_processors/team_webhook_processor.py
```diff
@@ -84,6 +88,17 @@ async def handle_event(
                 ),
             )
             data_to_upsert = extras_result[0]
+            enriched = await exporter.enrich_enterprise_teams_with_members(
```
danamsalem-ux, line 95: Add here more validations on data_to_upsert and where you use it as it might be null or undefined

jasonnwakaeze-debug: The null guard for data_to_upsert already exists before the if selector.members: block (lines 70-73) if the REST GET returns nothing, we return early. By the time we reach the enrichment code, data_to_upsert  is guaranteed non-null. Also it's actually pre-existing code or WDYT?

integrations/github/tests/github/webhook/webhook_processors/test_team_webhook_processor.py
```diff
@@ -203,6 +203,93 @@ def create_client_side_effect(client_type: GithubClientType) -> AsyncMock:
...
+        assert isinstance(result, WebhookEventRawResults)
+        assert result.updated_raw_results == [enriched_team]
+        assert result.deleted_raw_results == []
+        mock_enrich_enterprise.assert_awaited_once()
+        mock_enrich_saml.assert_awaited_once_with(
+            mock_graphql_client,
+            "test-org",
+            enterprise_members,
+            False,
+        )
```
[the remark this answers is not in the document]
danamsalem-ux: @jasonnwakaeze-debug  what do you think about that?

integrations/github/github/webhook/webhook_processors/team_webhook_processor.py
danamsalem-ux: i think we should add anyway

danamsalem-ux (approved):

PR #3480, opened 2026-07-08
[Integration][Aikido] Fix premature pagination on open-issue-groups by honoring x-has-next-page header

oiadebayo:
# Description

What - The `get_paginated_resource` method stopped fetching the next page when the current page was smaller than `page_size`. The Aikido API can return a partial page while still having more data, causing missing entities after a resync.

Why - Aikido introduced a permanent fix in their v1 API: the `x-has-next-page: true|false` response header on the `open-issue-groups` endpoint. When present, this header is the authoritative signal for whether to continue paginating.

How - Updated `get_paginated_resource` to check the `x-has-next-page` header first. If the header is present, it drives the decision; if absent, the existing count-vs-page-size heuristic is used as a fallback (preserving backward compatibility for all other endpoints).

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [x] Handled pagination
- [ ] Support Multi account

## Screenshots

N/A

## API Documentation

https://apidocs.aikido.dev/reference/listopenissuegroups


mk-armah (approved):
LGTM

matan84 (approved):

PR #3495, opened 2026-07-12
[Integration][GitHub] Auth foundation for multi-org App

omby8888:
## Summary
- Introduce typed auth providers (`_PatAuthProvider`, `_AppAuthProvider`) resolved lazily via `get_auth_provider()` at call time, avoiding import-time Ocean context access.
- Add GitHub App installation registry for multi-org discovery and per-installation authenticators/rate-limit scopes.
- Add async `create_github_client_for_org()` for the new multi-org auth stack (webhooks, future callers).
- **Keep legacy `create_github_client()`** on a separate sync path: `GitHubAuthenticatorFactory` + `LegacyOrgGitHubAppAuthenticator` with lazy installation ID/token fetch on first API call — no thread pool, no event-loop blocking at client creation.
- Remove legacy `github/context/auth.py` and route actor/authenticator lookups through the new provider API.
- Update auth, rate-limiter, GraphQL page-reduction, and integration tests for the new flow.

Port task: task_thwvuq

## Why two auth paths?

| Path | Used by | Purpose |
|------|---------|---------|
| **Legacy** — `create_github_client()` → `GitHubAuthenticatorFactory` → `LegacyOrgGitHubAppAuthenticator` | Resync handlers, actions, test fixtures, and all existing sync callers | Preserves pre-PR production behavior for single-org App/PAT setups. Auth resolves lazily on first HTTP call. |
| **New** — `create_github_client_for_org()` → `get_auth_provider()` → installation registry | Webhook startup (`on_start`), future multi-org rollout | Multi-org App discovery, per-installation rate-limit scopes, explicit org binding. |

The legacy path is intentionally left in place so this PR does not change runtime behavior for the majority of callers. It will be removed once nothing uses it.

## Next PR steps (migration plan)

1. **Migrate sync callers** — Replace `create_github_client()` with `await create_github_client_for_org(org)` in resync handlers, actions, and any other call sites that know their target organization.
2. **Audit remaining `create_github_client()` usage** — Confirm zero production callers (tests may keep a thin wrapper or switch to explicit org fixtures).
3. **Delete legacy stack** — Remove `GitHubAuthenticatorFactory`, `LegacyOrgGitHubAppAuthenticator`, and the sync `create_github_client()` shim once no callers remain.
4. **Optional follow-up** — Collapse `_clients` caching to a single code path if the factory indirection is no longer needed.

## Test plan
- [x] `make test` in `integrations/github`


integrations/github/github/clients/auth/github_app/installation_authenticator.py
```diff
@@ -0,0 +1,80 @@
...
+
+    async def get_headers(self) -> GitHubHeaders:
+        token_response = await self.get_token()
+        return GitHubHeaders(
+            Authorization=f"Bearer {token_response.token}",
+            Accept="application/vnd.github+json",
+            X_GitHub_Api_Version="2022-11-28",
+        )
+
+    async def _fetch_installation_token(self, installation_id: str) -> GitHubToken:
```
VenfiOranai: Is there a reason to have `installation_id` injectable if it's always gonna be `self._get_installation_id()`? If it was public then I'd get it more, but if it's protected I think it'll make it a bit easier to follow the flow

integrations/github/github/clients/auth/github_app/installation_authenticator.py
```diff
@@ -0,0 +1,80 @@
...
+    async def _get_installation_id(self) -> str:
+        if self.installation_id:
+            return self.installation_id
+
+        try:
+            url = f"{self.app_auth.github_host}/users/{self.organization}/installation"
+            headers = await self.app_auth.get_headers()
+            response = await self.client.get(url, headers=headers.as_dict())
+            response.raise_for_status()
+            return str(response.json()["id"])
```
VenfiOranai: Do we not want to set `self.installation_id` here?

integrations/github/github/clients/auth/github_app/installation_authenticator.py
```diff
@@ -0,0 +1,80 @@
...
+    ):
+        self.app_auth = app_auth
+        self.organization = organization
+        self.installation_id = installation_id
+        self.cached_installation_token: Optional[GitHubToken] = None
+        self.installation_token_lock = asyncio.Lock()
+
+    @property
+    def rate_limit_scope(self) -> str:
+        # TODO: replace to installation id based rate limit scope after create_github_client is deprecated
```
VenfiOranai, line 28: Is this something that can be tied to some existing task, just to make sure it doesn't get lost?

integrations/github/github/clients/auth/github_app/app_authenticator.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @classmethod
+    def from_config(cls) -> "GitHubAppAuthenticator":
+        return cls(
+            app_id=ocean.integration_config["github_app_id"],
+            private_key=ocean.integration_config["github_app_private_key"],
+            github_host=ocean.integration_config["github_host"],
+        )
+
+    def _generate_jwt(self) -> GitHubToken:
```
VenfiOranai: Redundant func? Only called once and the func that calls it is a one-liner

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -0,0 +1,76 @@
...
+        installation_id=installation_id,
+    )
+
+
+async def _fetch_installations() -> dict[str, GitHubAppInstallationAuthenticator]:
+    app_auth = GitHubAppAuthenticator.from_config()
+    index: dict[str, GitHubAppInstallationAuthenticator] = {}
+    async for page in app_auth.iter_app_installations():
+        for installation in page:
+            login = (installation.get("account") or {}).get("login")
```
VenfiOranai: Minor sugar syntax thing, but is there a reason not to do 
`installation.get("account", {}).get("login")`?

Like, can the `account` key exist but be None?

integrations/github/github/clients/auth/auth.py
```diff
@@ -0,0 +1,116 @@
...
+        return True
+
+
+_providers: tuple[type[_GitHubAuthProvider], ...] = (
+    _PatAuthProvider,
+    _AppAuthProvider,
+)
+
+
+def _resolve_provider() -> type[_GitHubAuthProvider]:
```
VenfiOranai: Redundant func?

integrations/github/github/clients/auth/personal_access_token_authenticator.py
```diff
@@ -1,24 +1,44 @@
...
-from loguru import logger
 
 
 class PersonalTokenAuthenticator(AbstractGitHubAuthenticator):
     def __init__(self, token: str):
         self._token = GitHubToken(token=token)
 
-    async def get_token(self, **kwargs: Any) -> GitHubToken:
+    @property
+    def rate_limit_scope(self) -> str:
```
VenfiOranai, line 17: I think it might be a bit safer to have this as a StrEnum/literal rather than any string

integrations/github/github/clients/auth/github_app/installation_authenticator.py
omby8888: your'e correct, made lot of changes in the final version and that slipped 👑

integrations/github/github/clients/auth/github_app/installation_authenticator.py
omby8888: we should only do that call once, but i'll set it in any case

integrations/github/github/clients/auth/github_app/installation_authenticator.py
omby8888: after this PR i'll imidiatly work on a followup PR that will drop this

integrations/github/github/clients/auth/personal_access_token_authenticator.py
omby8888: in installation_authenticator it's not a literal

integrations/github/github/clients/client_factory.py
```diff
@@ -1,35 +1,34 @@
...
-if hasattr(os, "register_at_fork"):
-    os.register_at_fork(after_in_child=_reset_clients_after_fork)
+_CLIENT_CLASSES: Dict[GithubClientType, Type[AbstractGithubClient]] = {
+    GithubClientType.REST: GithubRestClient,
+    GithubClientType.GRAPHQL: GithubGraphQLClient,
+}
+_clients: Dict[tuple[str, GithubClientType], AbstractGithubClient] = {}
 
 
 class GitHubAuthenticatorFactory:
```
raskinmaya, line 34: We now have two credential resolvers (`GitHubAuthenticatorFactory` here vs `get_auth_provider()` in `auth.py`) with different App rules (factory still requires `organization`). Almost all callers still hit this path. Shouldn't we migrate callers onto one resolver in this PR?

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -0,0 +1,76 @@
...
+async def _discover_installations() -> None:
+    global _authenticators_by_org
+    async with _discovery_lock:
+        if _authenticators_by_org:
+            return
+
+        _authenticators_by_org = await _fetch_installations()
+
+
+async def list_installations_authenticators() -> (
```
raskinmaya, line 74: This is the actual multi-org discovery path, but nothing in `on_start`/resync uses it — org listing still goes through `RestOrganizationExporter`, which rejects non-classic-PAT when org is absent. As shipped, this doesn't unlock multi-org app end-to-end

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -0,0 +1,76 @@
...
+            index[login] = _authenticator(
+                organization=login, installation_id=str(installation["id"])
+            )
+    return index
+
+
+async def _discover_installations() -> None:
+    global _authenticators_by_org
+    async with _discovery_lock:
+        if _authenticators_by_org:
```
raskinmaya: Once populated, this never refreshes for the process lifetime. New installs, uninstalls, or suspensions mid-run will keep serving a stale map. Needs invalidation (webhook), TTL, or re-discover-on-miss

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -0,0 +1,76 @@
...
+    authenticators = list(_authenticators_by_org.values())
+    logger.info(f"Discovered {len(authenticators)} GitHub App installation(s)")
+    return authenticators
+
+
+async def get_installation_authenticator_for_organization(
+    organization: str,
+) -> GitHubAppInstallationAuthenticator:
+    await _discover_installations()
+    if organization not in _authenticators_by_org:
```
raskinmaya, line 90: One installation missing `account.login` fails the entire discovery. For one failed installation, I think I'd prefer skip + warn, but let me know what do you think. We can use a threshold (< %k failed installs) as well. If we do that then also filter suspended installs (`suspended_at`/`suspended_by`) so they don't land in the map and blow up later on token mint

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -0,0 +1,76 @@
...
+    async for page in app_auth.iter_app_installations():
+        for installation in page:
+            login = installation.get("account", {}).get("login")
+            if not login:
+                raise AuthenticationException(
+                    f"No login found for installation {installation}"
+                )
+            index[login] = _authenticator(
+                organization=login, installation_id=str(installation["id"])
+            )
```
raskinmaya, line 60: Every `account.login` is treated as an organization, including user installs. Is this the intended behavior? Can you filter by `account.type`?

integrations/github/github/clients/auth/github_app/app_authenticator.py
```diff
@@ -0,0 +1,108 @@
...
+            private_key=ocean.integration_config["github_app_private_key"],
+            github_host=ocean.integration_config["github_host"],
+        )
+
+    async def get_token(self) -> GitHubToken:
+        now = datetime.now(timezone.utc)
+        expires_at = now + timedelta(minutes=self._JWT_EXPIRY_MINUTES)
+        payload = {
+            "iss": self.app_id,
+            "iat": now,
```
raskinmaya: GitHub recommends setting `iat` ~60s in the past to tolerate clock drift (see [here](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app#about-json-web-tokens-jwts)). Worth fixing now that app jwt is used for discovery, actor, and token exchange

integrations/github/github/clients/auth/github_app/installation_authenticator.py
```diff
@@ -0,0 +1,75 @@
...
+            Accept="application/vnd.github+json",
+            X_GitHub_Api_Version="2022-11-28",
+        )
+
+    async def _fetch_installation_token(self) -> GitHubToken:
+        try:
+            installation_id = await self._get_installation_id()
+            url = f"{self.app_auth.github_host}/app/installations/{installation_id}/access_tokens"
+            headers = await self.app_auth.get_headers()
+            response = await self.client.post(url, headers=headers.as_dict())
```
raskinmaya: Token exchange is app-authenticated but goes through `self.client` (installation-scoped limiter). Can't we use `self.app_auth.client`? This way app api rate limits aren't attributed to the installation scope

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -0,0 +1,76 @@
...
+    organization: str, installation_id: str
+) -> GitHubAppInstallationAuthenticator:
+    return GitHubAppInstallationAuthenticator(
+        app_auth=GitHubAppAuthenticator.from_config(),
+        organization=organization,
+        installation_id=installation_id,
+    )
+
+
+async def _fetch_installations() -> dict[str, GitHubAppInstallationAuthenticator]:
```
raskinmaya, line 44: Wasn't sure where to put it but `.port/spec.yaml` still says `githubAppInstallationId` is mandatory when app id is set. That conflicts with discovery-based multi-org. We should update the spec when this becomes user-facing. Let's leave a TODO for that

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -0,0 +1,76 @@
...
+from loguru import logger
+
+from github.clients.auth.github_app.app_authenticator import GitHubAppAuthenticator
+from github.clients.auth.github_app.installation_authenticator import (
+    GitHubAppInstallationAuthenticator,
+)
+from github.helpers.exceptions import AuthenticationException
+
+_authenticators_by_org: dict[str, GitHubAppInstallationAuthenticator] = {}
+_discovery_lock = asyncio.Lock()
```
raskinmaya, line 16: `_reset_after_fork` clears the dict but not `_discovery_lock`. Same issue the rate-limiter registry already guards against - reset/recreate the lock in the child or you risk post-fork deadlock

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -0,0 +1,76 @@
...
+def reset_authenticators_by_org() -> None:
+    global _authenticators_by_org
+    _authenticators_by_org = {}
+
+
+def _authenticator(
+    organization: str, installation_id: str
+) -> GitHubAppInstallationAuthenticator:
+    return GitHubAppInstallationAuthenticator(
+        app_auth=GitHubAppAuthenticator.from_config(),
```
raskinmaya: This creates a new `GitHubAppAuthenticator` per installation (and again in `_fetch_installations`). One shared app auth instance is enough

integrations/github/github/clients/utils.py
```diff
@@ -16,16 +17,20 @@ def integration_config(authenticator: AbstractGitHubAuthenticator) -> Dict[str,
     }
 
 
-def get_github_organizations() -> ListOrganizationOptions:
+def get_github_organizations(
```
raskinmaya, line 20: `get_github_organization` uses `.get("github_organization")` but `get_mono_repo_organization` still `KeyError`s

integrations/github/github/clients/auth/github_app/installation_registry.py
omby8888: I know, the plan is to remove both of the github_organization and the github app installation id params once the feature is fully released

integrations/github/github/clients/auth/github_app/installation_authenticator.py
omby8888: Nice catch!

integrations/github/github/clients/auth/github_app/installation_registry.py
omby8888: why should i filter that?

integrations/github/github/clients/client_factory.py
omby8888: This is the point, so this PR won't be blowed. In the followup PR i will change the callers to use the new credential resolver and the multi org discovery

integrations/github/github/clients/auth/github_app/installation_registry.py
omby8888: followup PR goal

integrations/github/github/clients/auth/github_app/installation_registry.py
omby8888: implemented TTL that should handle new app installations/uninstallation, but this still leaves a path for getting errors from app uninstallation/suspensions until the TTL.
I think the best way to handle this is in the followup PR to have dedicated error for installation token fetch, and have a wrapper in main that will continue in case of such error and revoke it from the installation cache so other kinds won't iterate that as well. wdyt?

raskinmaya: let's just call it out in the code if it's a מסע

integrations/github/github/clients/auth/github_app/installation_registry.py
raskinmaya: let's clarify with alina

raskinmaya (approved):
approving with those 2 open comments

integrations/github/github/clients/auth/github_app/installation_registry.py
omby8888: closing as we agreed on skipping suspended orgs, so reconcile would delete their data

PR #3513, opened 2026-07-13
[Core] Fail unregistered Ocean action runs instead of skipping

omby8888:
## Summary
- When `ExecutionManager` claims a run for an action type with no registered executor, acknowledge it and report it as failed instead of skipping until the visibility timeout expires
- Extract `_ack_and_fail_run_without_executor` helper and update unit test coverage

Port task: https://app.getport.io/taskEntity?identifier=task_ti3va0

## Test plan
- [x] `poetry run pytest port_ocean/tests/core/handlers/actions/test_execution_manager.py::TestExecutionManager::test_poll_action_runs_should_ack_and_fail_unregistered_actions`
- [ ] Verify an integration with a missing executor fails the run in Port with a clear error message

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+## 0.44.12 (2026-07-13)
+
+### Improvements
+
+- Fail unregistered Ocean action runs immediately by acknowledging and reporting them as failed instead of skipping them until the claim visibility timeout expires.
+
 ## 0.44.11 (2026-07-13)
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,11 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
-## 0.44.12 (2026-07-13)
+## 0.44.12 (2026-07-14)
 
 ### Improvements
 
-- Fail unregistered Ocean action runs immediately by acknowledging and reporting them as failed instead of skipping them until the claim visibility timeout expires.
+- Fail unregistered Ocean integration runs immediately by acknowledging and reporting them as failed instead of skipping them until the claim visibility timeout expires.
 
 ## 0.44.11 (2026-07-13)
 
--- a/port_ocean/core/handlers/actions/execution_manager.py
+++ b/port_ocean/core/handlers/actions/execution_manager.py
@@ -209,10 +209,8 @@ class ExecutionManager:
                     try:
                         action_type = run.action_type
                         if action_type not in self._actions_executors:
-                            logger.warning(
-                                "No Executors registered to handle this action, skipping run...",
-                                action_type=action_type,
-                                run_id=run.id,
+                            await self._ack_and_fail_run_without_executor(
+                                run, action_type
                             )
                             continue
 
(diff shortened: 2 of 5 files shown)


port_ocean/core/handlers/actions/execution_manager.py
```diff
@@ -256,6 +254,35 @@ async def _poll_action_runs(self) -> None:
...
+            action_type=action_type,
+            run_id=run.id,
+        )
+        try:
+            await ocean.port_client.acknowledge_run(run)
+            await ocean.port_client.report_run_completed(
+                run,
+                success=False,
+                message=f"No executor registered for action type '{action_type}'",
+                should_raise=False,
```
raskinmaya: After `acknowledge_run` succeeds, the run is no longer reclaimable. If `report_run_completed` fails (network, 5xx's, auth), the broad `except Exception` logs and swallows it. The run stays `IN_PROGRESS` in Port with no retry path. This is even worse than the old "skip until visibility timeout" behavior - the run just simply won't come back.
I think you should either catch only `RunAlreadyAcknowledgedError` around ack, and report failures on complete separately with retry or re-raise, or add a port-client helper that treats ack + complete as one transactional unit with compensating behavior

port_ocean/core/handlers/actions/execution_manager.py
```diff
@@ -209,10 +209,8 @@ async def _poll_action_runs(self) -> None:
                     try:
                         action_type = run.action_type
                         if action_type not in self._actions_executors:
-                            logger.warning(
-                                "No Executors registered to handle this action, skipping run...",
-                                action_type=action_type,
-                                run_id=run.id,
+                            await self._ack_and_fail_run_without_executor(
```
raskinmaya, line 210: You fixed unregistered executors, but two other `continue` paths still claim runs and never ack:
- [Here](https://github.com/port-labs/ocean/pull/3513/changes#diff-8b1d1015f09431178d8ea583903aa066b5740fa3ee127783e5a303f4610c3ba5L224) (lines 217-222): run already queued/in-flight locally, but Port re-delivered it -> silent skip, visibility timeout loop.
- And [here](https://github.com/port-labs/ocean/pull/3513/changes#diff-8b1d1015f09431178d8ea583903aa066b5740fa3ee127783e5a303f4610c3ba5L236) (lines 234-240): `PartitionKeyNotFoundError` - malformed payload -> skip without fail

If the goal is "never leave a claim run unacked", the PR is incomplete.
`PartitionKeyNotFoundError` is especially similar - the integration cannot process the run and should fail fast with a clear message, same as unregistered executor

port_ocean/core/handlers/actions/execution_manager.py
```diff
@@ -209,10 +209,8 @@ async def _poll_action_runs(self) -> None:
                     try:
                         action_type = run.action_type
```
raskinmaya: Why do you need this param? It's always `run.action_type`. Passing it separately invites drift and adds noise. Let's drop the parameter, you can read from `run` directly

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 <!-- towncrier release notes start -->
 
+## 0.44.12 (2026-07-13)
+
+### Improvements
+
+- Fail unregistered Ocean action runs immediately by acknowledging and reporting them as failed instead of skipping them until the claim visibility timeout expires.
```
raskinmaya: The changelog says "action runs" only, but the fix applies to both `ActionRun` and `WorkflowNodeRun` - `claim_pending_runs` returns both, and the unregistered-executor check uses `run.action_type` on the shared `IntegrationRun` interface. Consider wording like "unregistered integration runs" or explicitly mention workflow node runs. The test at `test_execution_manager` ([here](https://github.com/port-labs/ocean/pull/3513/changes#diff-04627f909687304790a248ca1c739a8cb827de264f54b287c809d2aa526a4b79R582)) has the same gap: it only uses `generate_mock_action_run`, not `generate_mock_wf_node_run`

port_ocean/core/handlers/actions/execution_manager.py
omby8888: first point it's valid continuing in the dedup mechanism, removed the PartitionKeyNotFoundError handling as it's not raised anywhere lol

raskinmaya (approved):

PR #3527, opened 2026-07-14
[Core] Task ti44hp/local cache for bp in ocean core

VenfiOranai:
# Description

What -

Why -

How -

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Cached blueprint metadata can be stale for up to the TTL if the blueprint changes outside these client mutations; invalidation on local writes limits but does not eliminate that window for external updates.
> 
> **Overview**
> Adds an in-process **TTL cache** for Port blueprint fetches so repeated `get_blueprint` calls during resync and entity processing avoid redundant API traffic.
> 
> `BlueprintClientMixin` now stores successful GET results in `BlueprintCache` (default **120s** TTL, configurable via `PortSettings.blueprint_cache_ttl_seconds` and wired through `PortClient` / `Ocean`). Callers can bypass caching with `use_cache=False`. **Create, patch, and delete** blueprint paths invalidate the cached entry for that identifier.
> 
> Unit tests cover TTL expiry, invalidation, cache hits on second GET, and refetch after patch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8819069ccc4a6bad64ae87b9b836d5a47023683a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


port_ocean/clients/port/blueprint_cache.py
raskinmaya, line 1: The overall direction makes sense - `get_blueprint` is hit repeatedly during resync and a short in-process cache is justified. I think this can be slimmed though, let me know what do you think:

We can align with the existing feature-flags pattern instead of introducing a dedicated `BlueprintCache`/ `BlueprintCacheEntry` module: keep a small TTL map on the `mixin` (or make it event/resync-scoped), return `Blueprint | None`, and keep invalidation private. Then we can drop the create-path invalidate (it’s effectively a no-op), and either bound cache size or accept that lazy TTL eviction can retain unused entries indefinitely. We should also document the staleness tradeoff (external Port/UI/other-replica updates won’t invalidate for up to TTL)

port_ocean/clients/port/mixins/blueprints.py
```diff
@@ -12,21 +14,48 @@
...
     async def get_blueprint(
-        self, identifier: str, should_log: bool = True
+        self, identifier: str, should_log: bool = True, *, use_cache: bool = True
     ) -> Blueprint:
+        if (
+            use_cache
+            and (entry := self._blueprint_cache.get(identifier)) is not None
+            and time.monotonic() - entry.cached_at < self._blueprint_cache_ttl_seconds
+        ):
+            return entry.blueprint
```
raskinmaya, line 41: Correct me if I'm wrong, but expired entries are never removed - we only skip them. The earlier `BlueprintCache.get` deleted expired keys; that was lost when the cache was inlined. Please evict on miss (`pop`/`del`) so this is a real TTL cache, not “lazy ignore forever”

port_ocean/clients/port/mixins/blueprints.py
```diff
@@ -12,21 +14,48 @@
...
+
         logger.info(f"Fetching blueprint with id: {identifier}")
         response = await self.client.get(
             f"{self.auth.api_url}/blueprints/{identifier}",
             headers=await self.auth.headers(),
         )
         handle_port_status_code(response, should_log=should_log)
-        return Blueprint.parse_obj(response.json()["blueprint"])
+        blueprint = Blueprint.parse_obj(response.json()["blueprint"])
+        self._blueprint_cache[blueprint.identifier] = BlueprintCacheEntry(
```
raskinmaya: Lookup uses the request identifier, but we store under `blueprint.identifier`. Is this intentional? If Port ever returns a canonicalized/different id, we’ll miss on the next call with the original key and either double-fetch or keep two entries. Prefer storing under the requested identifier (optionally also under `blueprint.identifier` if they differ)

port_ocean/clients/port/mixins/blueprints.py
```diff
@@ -12,21 +14,48 @@
...
+    ):
         self.auth = auth
         self.client = client
+        self._blueprint_cache_ttl_seconds = blueprint_cache_ttl_seconds
+        self._blueprint_cache: dict[str, BlueprintCacheEntry] = {}
 
     async def get_blueprint(
-        self, identifier: str, should_log: bool = True
+        self, identifier: str, should_log: bool = True, *, use_cache: bool = True
     ) -> Blueprint:
```
raskinmaya, line 37: No production caller passes `use_cache=False`. If freshness after local writes is the concern, invalidate on create/patch/delete instead of adding an unused flag. Keep it only if we’re deliberately mirroring `get_port_app_config`; otherwise this looks like you are not gonna need this

port_ocean/clients/port/mixins/blueprints.py
```diff
@@ -12,21 +14,48 @@
...
         )
         handle_port_status_code(response, should_log=should_log)
-        return Blueprint.parse_obj(response.json()["blueprint"])
+        blueprint = Blueprint.parse_obj(response.json()["blueprint"])
+        self._blueprint_cache[blueprint.identifier] = BlueprintCacheEntry(
+            blueprint=blueprint
+        )
+        return blueprint
+
+    def invalidate_all_cached_blueprints(self) -> None:
```
raskinmaya: Sorry for nitpicking but I think renaming to `clear_blueprint_cache` would align better with `cache_provider.clear()`. Also: with TTL + resync-start clear, is a public "invalidate all" the right API, or should clearing stay an internal detail of resync?

port_ocean/clients/port/mixins/blueprints.py
```diff
@@ -12,21 +14,48 @@
...
+        self._blueprint_cache.clear()
 
     async def create_blueprint(
```
raskinmaya, line 56: I see you removed per-identifier invalidation after successful create/patch/delete. I think dropping it is incorrect:
- At `default_origin_setup.py`: `get_blueprint` -> `create/patch` in the same process
- At `clean.py`: `delete_blueprint` after blueprints may already be cached
- Stale schema feeds `get_related_entities` (relation targets) and `_resolve_required_fields` (`ERROR` vs `WARNING`)

A one-liner `self._blueprint_cache.pop(identifier, None)` after success would be the right fix. TTL alone only covers external Port changes, it does not cover our writes

port_ocean/clients/port/mixins/blueprints.py
raskinmaya, line 1: Please explicitly validate/document these Port-side cases with a 120s cache and no live-event invalidation (only resync start/end):

- Relation added/renamed/removed in Port UI while Ocean is processing live events -> related-entity resolution against stale `blueprint.relations`
- Property requiredness changed -> misconfiguration severity wrong until TTL/resync
- Blueprint deleted in Port -> cached `get_blueprint` still returns success until TTL/resync

Write-path invalidation covers ocean’s own mutations; TTL is the only guard for external changes. If that’s accepted, say so in the PR description and/or changelog

port_ocean/tests/clients/port/test_blueprint_cache.py
```diff
@@ -0,0 +1,113 @@
...
+                many=False, required=True, target="team", title="Owner"
+            )
+        },
+    )
+
+
+def _seed_cache(blueprint_client: BlueprintClientMixin, blueprint: Blueprint) -> None:
+    blueprint_client._blueprint_cache[blueprint.identifier] = BlueprintCacheEntry(
+        blueprint=blueprint
+    )
```
raskinmaya: Prefer not seeding private `_blueprint_cache`. Drive hits via a real `get_blueprint` (as in `test_get_blueprint_uses_cache_on_second_call`) and only monkeypatch time for expiry

port_ocean/tests/clients/port/test_blueprint_cache.py
raskinmaya, line 1: Missing coverage worth adding:

- `use_cache=False` (if the flag stays)
- create/patch/delete leave cache consistent (once write invalidation is restored)
- failed HTTP responses are not cached
- request identifier != response `blueprint.identifier` store/lookup behavior

port_ocean/clients/port/mixins/blueprints.py
VenfiOranai: If an entry is expired and this if returns false, we override it in the cache, so adding a pop/del is redundant

port_ocean/clients/port/mixins/blueprints.py
VenfiOranai: Done

port_ocean/clients/port/mixins/blueprints.py
VenfiOranai: Removed

port_ocean/clients/port/mixins/blueprints.py
VenfiOranai: I renamed it as you suggested, but not sure I understand the second half of your comment

port_ocean/clients/port/mixins/blueprints.py
VenfiOranai: Added them back

port_ocean/tests/clients/port/test_blueprint_cache.py
VenfiOranai: Added cases

port_ocean/tests/clients/port/test_blueprint_cache.py
VenfiOranai: Done

port_ocean/clients/port/mixins/blueprints.py
VenfiOranai: Played around with required mid-sync, it doesn't really have too much of an effect since the flow here doesn't block upserts regardless, only dictates whether the logs should be warning or error.

raskinmaya: amazing

raskinmaya (approved):

PR #3528, opened 2026-07-14
[Integration] Add skill and plugin kinds for GitHub and GitLab v2

matanAlltra:
- **Add skill and plugin kinds to GitHub and GitLab Ocean exporters**
- **Add validated skill/plugin Port mapping fragments for GitHub and GitLab**
- **Rename Port plugin blueprint examples to agentPlugin**

# Description

What -

Why -

How -

## Type of change

Please leave one option from the following and delete the rest:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] New Integration (non-breaking change which adds a new integration)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Non-breaking change (fix of existing functionality that will not change current behavior)
- [ ] Documentation (added/updated documentation)

<h4> All tests should be run against the port production environment(using a testing org). </h4>

### Core testing checklist

- [ ] Integration able to create all default resources from scratch
- [x] Resync finishes successfully
- [ ] Resync able to create entities
- [ ] Resync able to update entities
- [ ] Resync able to detect and delete entities
- [ ] Scheduled resync able to abort existing resync and start a new one
- [ ] Tested with at least 2 integrations from scratch
- [ ] Tested with Kafka and Polling event listeners
- [ ] Tested deletion of entities that don't pass the selector

### Integration testing checklist

- [ ] Integration able to create all default resources from scratch
- [ ] Completed a full resync from a freshly installed integration and it completed successfully
- [ ] Resync able to create entities
- [ ] Resync able to update entities
- [ ] Resync able to detect and delete entities
- [ ] Resync finishes successfully
- [ ] If new resource kind is added or updated in the integration, add example raw data, mapping and expected result to the `examples` folder in the integration directory.
- [ ] If resource kind is updated, run the integration with the example data and check if the expected result is achieved
- [ ] If new resource kind is added or updated, validate that live-events for that resource are working as expected
- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Implemented the code in async
- [ ] Support Multi account

## Screenshots

Include screenshots from your environment showing how the resources of the integration will look.

## API Documentation

Provide links to the API documentation used for this integration.

---

> [!NOTE]
> **Medium Risk**
[...]

PR #3532, opened 2026-07-15
[Integration][Github] Add skill and plugin kinds

matanAlltra:
## Summary
- Adds GitHub Ocean `skill` and `plugin` kinds to catalog Agent Skills (`SKILL.md`) and agent plugin packages
- Includes push webhooks, example Port mapping fragments (`agentPlugin` blueprint), and unit tests
- Opt-in only — not added to the default `port-app-config.yml`

## Test plan
- [ ] Resync with `kind: skill` and `kind: plugin` mappings against a repo with skills + plugin manifests
- [ ] Confirm skill-only repos (e.g. `.cursor/skills`) do not create `agentPlugin` stubs when using the search relation example
- [ ] Push webhook upsert/delete for skill and plugin paths
- [ ] CI lint/tests for `integrations/github`

Split from https://github.com/port-labs/ocean/pull/3528 (GitHub-only).

image
image
image
image


integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
```diff
@@ -0,0 +1,179 @@
...
+                    "description": "",
+                    "instructions": None,
+                    "frontmatter": {},
+                    "path": str(Path(file_info["filename"]).parent),
+                    "skillMdPath": file_info["filename"],
+                    "root": (
+                        Path(file_info["filename"]).parts[0]
+                        if Path(file_info["filename"]).parts
+                        else ""
+                    ),
```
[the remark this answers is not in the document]
matanAlltra: Fixed — delete stubs now use `match_skill_root(...)` (same as upserts) and normalize `path` with forward slashes.

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
```diff
@@ -0,0 +1,179 @@
...
+    ) -> bool:
+        if selector.repos is None:
+            return current_branch == default_branch
+        for mapping in selector.repos:
+            if mapping.name == repo_name and (
+                mapping.branch == current_branch
+                or (mapping.branch is None and current_branch == default_branch)
+            ):
+                return True
+        return False
```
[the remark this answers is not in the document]
matanAlltra: Acknowledged / deferring. Same pattern already exists in `FileWebhookProcessor`; extracting a shared helper is a follow-up cleanup, not blocking this PR.

integrations/github/github/core/exporters/file_exporter/core.py
```diff
@@ -398,11 +407,16 @@ async def get_tree_recursive(
...
                 f"Did not retrieve tree from {repo}@{branch} from {organization}"
             )
-            return []
+            return [], False
+
+        truncated = bool(response.get("truncated"))
+        if truncated:
+            logger.warning(
+                "Git tree was truncated by GitHub; plugin/file detection may be incomplete"
+            )
```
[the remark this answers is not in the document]
matanAlltra: Deferring — `truncated` here matches the existing file-exporter naming (`tree_truncated` / similar). Renaming would be a wider style cleanup outside this change.

integrations/github/github/core/exporters/plugin_exporter/core.py
```diff
@@ -0,0 +1,137 @@
...
+        manifests: dict[str, Any] = {}
+        for path in present:
+            file_data = await self._file_exporter.get_resource(
+                FileContentOptions(
+                    organization=organization,
+                    repo_name=repo_name,
+                    file_path=path,
+                    branch=branch,
+                )
+            )
```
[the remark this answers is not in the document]
matanAlltra: Deferring for now — manifests per repo are usually few; happy to parallelize with `asyncio.gather` in a follow-up if this shows up in resync latency.

integrations/github/github/webhook/webhook_processors/plugin_webhook_processor.py
```diff
@@ -0,0 +1,154 @@
...
+    async def handle_event(
+        self, payload: EventPayload, resource_config: ResourceConfig
+    ) -> WebhookEventRawResults:
+        organization = self.get_webhook_payload_organization(payload)["login"]
+        repository = payload["repository"]
+        before_sha = payload["before"]
+        after_sha = payload["after"]
+        repo_name = repository["name"]
+        default_branch = repository["default_branch"]
+        current_branch = payload["ref"].split("/")[-1]
```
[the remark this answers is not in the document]
matanAlltra: Fixed — both skill and plugin webhooks now use `payload["ref"].removeprefix("refs/heads/")` so slash-containing branch names are preserved.

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
```diff
@@ -0,0 +1,179 @@
...
+        path: str,
+        roots: list[str],
+        extra_paths: list[str],
+        patterns: list[str],
+    ) -> bool:
+        if not path.endswith("SKILL.md"):
+            return False
+        if path_under_roots_or_extra(path, roots, extra_paths):
+            return True
+        return any(matches_glob_pattern(path, pattern) for pattern in patterns)
```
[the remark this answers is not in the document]
matanAlltra: Fixed — skill matching is now case-insensitive on `SKILL.md` and uses `wcmatch` `DOTGLOB` (same as resync discovery).

integrations/github/.port/examples/skill/mapping.yml
omby8888, line 1: This directory contains raw data examples only, This is more of a docs related file so let's i expect it to be n Port docs rather than here

integrations/github/.port/examples/skill/README.md
omby8888, line 1: same, this shouldn't be here

integrations/github/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "github-ocean"
-version = "6.2.21"
+version = "6.2.22"
```
omby8888: bump minor version instead

integrations/github/github/core/exporters/plugin_exporter/core.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+        if not plugin:
+            return PluginBuildResult(plugin_item=None, tree_truncated=truncated)
+        return PluginBuildResult(
+            plugin_item={
+                "plugin": plugin,
+                "repository": repository,
+                "branch": branch,
+                "organization": organization,
+            },
```
omby8888: our convention is to list root attributes that aren't native to Github with `__` prefix, so that would be
```suggestion
            plugin_item={
                "__plugin": plugin,
                "__repository": repository,
                "__branch": branch,
                "__organization": organization,
            },
```

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
```diff
@@ -0,0 +1,159 @@
...
+        organization = self.get_webhook_payload_organization(payload)["login"]
+        repository = payload["repository"]
+        before_sha = payload["before"]
+        after_sha = payload["after"]
+        repo_name = repository["name"]
+        default_branch = repository["default_branch"]
+        current_branch = payload["ref"].removeprefix("refs/heads/")
+
+        selector = cast(GithubSkillResourceConfig, resource_config).selector
+        effective_roots = selector.roots or list(DEFAULT_SKILL_ROOTS)
```
omby8888: no need for the default fallback here. this populates by the pydantic model already, and changing this would require changes in several places

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
```diff
@@ -0,0 +1,159 @@
...
+                        "path": skill_dir,
+                        "skillMdPath": filename,
+                        "root": match_skill_root(filename, effective_roots)
+                        or skill_dir.split("/")[0],
+                    },
+                    "repository": repository,
+                    "branch": current_branch,
+                    "organization": organization,
+                }
+            )
```
omby8888: shouldnt this use the util?

integrations/github/main.py
```diff
@@ -1023,6 +1038,84 @@ async def resync_files(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
...
+
+    rest_client = create_github_client()
+    org_exporter = RestOrganizationExporter(rest_client)
+    repo_exporter = RestRepositoryExporter(rest_client)
+    plugin_exporter = PluginExporter(rest_client)
+
+    config = cast(GithubPluginResourceConfig, event.resource_config)
+    app_config = cast(GithubPortAppConfig, event.port_app_config)
+    selector = config.selector
+    providers = selector.providers or list(DEFAULT_PLUGIN_PROVIDERS)
```
omby8888: use directly without the default fallback as it's populated by pydantic already

integrations/github/integration.py
```diff
@@ -281,6 +283,83 @@ class GithubFileResourceConfig(ResourceConfig):
     )
 
 
+class GithubSkillSelector(Selector, RepositorySourceModel):
+    content: Literal["frontmatter", "skill.md"] = Field(
```
omby8888: we fetch the skill.md body in any case so I don't see a reason to not include it in the raw data always.
if the user not interested in it he'll not refer it in the mapping

integrations/github/integration.py
```diff
@@ -281,6 +283,83 @@ class GithubFileResourceConfig(ResourceConfig):
...
+            "`frontmatter` returns name/description only; "
+            "`skill.md` also includes the markdown body as instructions."
+        ),
+    )
+    roots: list[str] = Field(
+        title="Roots",
+        default_factory=lambda: list(DEFAULT_SKILL_ROOTS),
+        description=(
+            "Skill parent directories to scan recursively for SKILL.md. "
+            "Defaults cover .agents, Antigravity (.agent), Cursor, Claude, "
```
omby8888: no need to describe the defaults as they should appear in the UI

integrations/github/integration.py
```diff
@@ -281,6 +283,83 @@ class GithubFileResourceConfig(ResourceConfig):
     )
 
 
+class GithubSkillSelector(Selector, RepositorySourceModel):
```
omby8888: since `RepositorySourceModel` provides single org filtration only, in order to limit the lookup on several orgs i would have to duplicate the kind.
pls check out how it's done in the file kind selector as i think it's better structure that supports multi org filtration

integrations/github/integration.py
```diff
@@ -281,6 +283,83 @@ class GithubFileResourceConfig(ResourceConfig):
...
+        ),
+    )
+    paths: list[str] = Field(
+        title="Paths",
+        default_factory=list,
+        description=(
+            "Optional extra glob patterns for SKILL.md files outside the "
+            "configured roots (e.g. '**/legacy-skills/*/SKILL.md')."
+        ),
+    )
```
omby8888: the only diff between those 2 is that the roots is later being templated to glob pattern. lets use only one of them and make the defaults use glob in the first place

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+def roots_to_globs(roots: list[str]) -> list[str]:
+    """Convert skill roots to GitHub tree globs that match SKILL.md files."""
+    globs: list[str] = []
+    for root in roots:
+        clean = root.strip().strip("/")
+        if not clean:
+            continue
+        globs.append(f"{clean}/**/{SKILL_MD_FILENAME}")
+        globs.append(f"{clean}/{SKILL_MD_FILENAME}")
```
omby8888: You can safely remove the second line, `wcmatch` handles top-level files perfectly with the ** recursive pattern.

integrations/github/github/core/exporters/skill_exporter/core.py
```diff
@@ -0,0 +1,88 @@
...
+        super().__init__(client)
+        self._file_exporter = RestFileExporter(client)
+
+    async def get_resource(self, options: Any) -> RAW_ITEM:
+        raise NotImplementedError("SkillExporter does not support get_resource")
+
+    def get_paginated_resources(
+        self, options: List[ListFileSearchOptions]
+    ) -> ASYNC_GENERATOR_RESYNC_TYPE:
+        raise NotImplementedError("Use get_paginated_skills")
```
omby8888: I think we should inherit RestFileExporter, that way you'll have get_resource populated and u could override get_paginated_resources to operate as `get_paginated_skills` does more natively using `self.super.get_paginated_resources`

integrations/github/github/core/exporters/file_exporter/core.py
```diff
@@ -385,11 +385,20 @@ async def fetch_commit_diff(
...
         self, organization: str, repo: str, branch: str
     ) -> List[Dict[str, Any]]:
         """Retrieve the full recursive tree for a given branch."""
+        tree_items, _truncated = await self.get_tree_recursive_meta(
+            organization, repo, branch
+        )
+        return tree_items
+
+    @cache_coroutine_result()
+    async def get_tree_recursive_meta(
```
omby8888: Lets simply return that tuple (or better - pydantic model) in `get_tree_recursive` as i didn't saw it has many refs. this is a super rare edge case that i doubt we'll face in the near future but i would like to add a warning log in the file exporter as well for that so no need for 2 functions, we'll check if the response was truncated in either case

integrations/github/github/core/exporters/skill_exporter/core.py
```diff
@@ -0,0 +1,88 @@
...
+                        branch=file_obj["branch"],
+                        organization=file_obj.get("organization"),
+                        roots=roots,
+                    )
+                )
+            if skills:
+                yield skills
+
+
+def build_skill_file_patterns(
```
omby8888: this should be a builder class much like `FilePatternMappingBuilder` (could inherit from it as well), in any case this is not the exporter's responsibility imo

integrations/github/github/webhook/webhook_processors/base_repository_webhook_processor.py
```diff
@@ -11,11 +11,28 @@
 from github.webhook.webhook_processors.github_abstract_webhook_processor import (
     _GithubAbstractWebhookProcessor,
 )
-from integration import GithubPortAppConfig, RepoSearchSelector
+from integration import GithubPortAppConfig, RepoSearchSelector, RepositorySourceModel
 from loguru import logger
 
 
 class BaseRepositoryWebhookProcessor(_GithubAbstractWebhookProcessor):
+    def _is_applicable_to_repo_branch(
```
omby8888: this logic exists exactly in `integrations/github/github/webhook/webhook_processors/file_webhook_processor.py` so u can inherit this instead of `BaseRepositoryWebhookProcessor`

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
```diff
@@ -0,0 +1,159 @@
...
+
+        rest_client = create_github_client()
+        exporter = RestFileExporter(rest_client)
+        diff_data = await exporter.fetch_commit_diff(
+            organization, repo_name, before_sha, after_sha
+        )
+        changed_files = diff_data.get("files") or []
+
+        skill_changes = [
```
omby8888: the logic here should be much like in the main.py file imo.
u should build the file patterns and then use `integrations/github/github/webhook/webhook_processors/file_webhook_processor.py` `_get_matching_patterns` to verify if they match against the commit's changed files easily

integrations/github/github/core/exporters/skill_exporter/__init__.py
omby8888, line 1: not sure all of this really worth exporting as they should be encapsulated logic

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Returns (frontmatter_dict, body). On missing/invalid frontmatter returns
+    ({}, full content as body).
+    """
+    text = content.replace("\r\n", "\n")
+    if not text.startswith("---"):
+        return {}, text
+
+    parts = text.split("---", 2)
+    if len(parts) < 3:
```
omby8888, line 85: this if will never be entered so u can remove it

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+    from github.helpers.utils import matches_glob_pattern
+
+    for pattern in extra_paths:
+        if matches_glob_pattern(skill_md_path, pattern, flags=glob.DOTGLOB):
+            return True
+    return False
+
+
+def parse_skill_markdown(content: str) -> tuple[dict[str, Any], str]:
```
omby8888: this could be private fn

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -0,0 +1,162 @@
...
+        "description": description,
+        "instructions": None,
+        "frontmatter": frontmatter,
+        "path": skill_dir,
+        "skillMdPath": skill_md_path,
+        "root": root,
+    }
+
+    if content_mode == "skill.md":
+        skill["instructions"] = body
```
omby8888: let's have pydantic model for that

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -0,0 +1,162 @@
...
+        logger.warning(f"Failed to parse skill frontmatter: {exc}")
+        return {}, body
+
+    if isinstance(parsed, dict):
+        return parsed, body
+    logger.warning("Skill frontmatter did not parse to a mapping; ignoring")
+    return {}, body
+
+
+def build_skill_object(
```
omby8888: can be private

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+    name = frontmatter.get("name")
+    if not isinstance(name, str) or not name.strip():
+        name = path_name
+
+    description = frontmatter.get("description")
+    if not isinstance(description, str):
+        description = ""
+
+    root = match_skill_root(skill_md_path, roots) or skill_dir.split("/")[0]
```
omby8888: isn't that contained in the path already? why is that attr needed?

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -0,0 +1,162 @@
...
+    for root in roots:
+        clean = root.strip().strip("/")
+        if not clean:
+            continue
+        globs.append(f"{clean}/**/{SKILL_MD_FILENAME}")
+        globs.append(f"{clean}/{SKILL_MD_FILENAME}")
+    return globs
+
+
+def match_skill_root(skill_md_path: str, roots: list[str]) -> str | None:
```
omby8888: not sure we need that fn after we'll have dedicated builder

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -0,0 +1,162 @@
...
+    ".opencode/skills",
+    "skills",
+]
+
+SKILL_MD_FILENAME = "SKILL.md"
+
+SkillContentMode = Literal["frontmatter", "skill.md"]
+
+
+def roots_to_globs(roots: list[str]) -> list[str]:
```
omby8888: we could have the roots as globs by default imo and this will not be needed

omby8888 (changes requested):

integrations/github/.port/examples/skill/mapping.yml
matanAlltra: Done — removed `mapping.yml` from `.port/examples`. Kept only the raw `a.json` example; mapping docs can live in Port docs.

integrations/github/.port/examples/skill/README.md
matanAlltra: Done — removed the README from examples.

integrations/github/pyproject.toml
matanAlltra: Done — bumped to a minor version. After merging main (which was already at 6.3.1), this lands as **6.4.0**.

integrations/github/github/core/exporters/plugin_exporter/core.py
matanAlltra: Done with the folder-kind convention: keep the kind object unprefixed (`plugin` / `skill`), and prefix context fields as `__repository`, `__branch`, `__organization`.

We intentionally did **not** use `__plugin` / `__skill` — `folder` kind uses the same pattern (`folder` + `__repository`/…).

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
matanAlltra: Done — removed the `or list(DEFAULT_*)` fallbacks. Selector defaults come from the pydantic models now.

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
matanAlltra: Done — delete stubs now use `infer_skill_root(...)` (same root resolution as upserts).

integrations/github/main.py
matanAlltra: Done — using `selector.providers` directly (pydantic default_factory fills it).

integrations/github/integration.py
matanAlltra: Agreed — removed the `content` selector. We always include the markdown body as `instructions`; mapping can ignore it if unused.

integrations/github/integration.py
matanAlltra: Done — dropped the long “defaults cover…” description as part of collapsing to glob `paths` only.

integrations/github/integration.py
matanAlltra: Done — skills now use a `paths: list[{path, organization?, repos?}]` selector (same shape as file kind). Plugins use `repositories: list[{organization?, repos?}]` for the same multi-org reason.

integrations/github/integration.py
matanAlltra: Done — single `paths` field with glob defaults (e.g. `.cursor/skills/**/SKILL.md`). Removed `roots`.

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Done — defaults are only the `**/SKILL.md` globs now (no separate top-level `root/SKILL.md` entry).

integrations/github/github/core/exporters/skill_exporter/core.py
matanAlltra: Deferring for now — kept composition with `RestFileExporter` (same as folder exporter). Inheritance would couple skill to file’s `get_resource`/`get_paginated_resources` surface without much gain in this PR; happy to revisit if you still prefer inheritance.

integrations/github/github/core/exporters/file_exporter/core.py
matanAlltra: Done — collapsed `_meta` into `get_tree_recursive` returning `(tree, truncated)`, with the truncate warning logged for all callers (file + plugin). Light footprint: one unpack in the file exporter path.

integrations/github/github/core/exporters/skill_exporter/core.py
matanAlltra: Simplified differently: removed `build_skill_file_patterns` from the exporter. `main.py` now builds `GithubFilePattern`s and reuses `FilePatternMappingBuilder` directly (no new builder class).

integrations/github/github/webhook/webhook_processors/base_repository_webhook_processor.py
matanAlltra: Partially addressed — skill webhook now builds file patterns and uses `get_matching_files` / the same branch applicability helper as file webhooks. Kept extending `BaseRepositoryWebhookProcessor` (plugin needs the repo-branch helper too) rather than inheriting the full file processor.

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
matanAlltra: Done — skill webhook converts selector `paths` to `GithubFilePattern`s, filters with `_is_applicable_to_repo_branch`, then uses `get_matching_files` on the commit diff (same approach as file webhook).

integrations/github/github/core/exporters/skill_exporter/__init__.py
matanAlltra: Done — `__init__.py` now only exports what resync/webhooks need.

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Kept — this path is reachable for an opening `---` without a closing fence (`"---\nname: x".split("---", 2)` has length 2). Added a unit test for that case.

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Done — renamed to `_parse_skill_markdown`.

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Deferring — file kind uses a TypedDict/`dict` raw shape rather than a pydantic model for emitted items. Kept a plain dict for consistency; we can introduce a model later if we standardize across file/folder/skill.

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Left public — still used by webhook/resync helpers and tests. Encapsulated the parser as private instead (`_parse_skill_markdown`).

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Kept `root` — `path` is the skill folder (e.g. `.cursor/skills/hello`), while `root` is the matched convention prefix (`.cursor/skills`). Useful as a stable mapping/filter field without parsing globs in JQ.

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Replaced with `infer_skill_root(...)` against the configured path globs (no separate roots list / builder).

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Done — defaults are globs up front (`DEFAULT_SKILL_PATHS`); `roots_to_globs` is gone.

integrations/github/.port/examples/skill/a.json
omby8888, line 1: can we have few more examples? also for plugins?

integrations/github/github/core/exporters/file_exporter/core.py
```diff
@@ -134,7 +134,9 @@ async def collect_matched_files(
             logger.debug(
                 f"Processing pattern '{pattern}' on branch '{branch}' for {repo_name} from {organization}"
             )
-            tree = await self.get_tree_recursive(organization, repo_name, branch)
+            tree, _truncated = await self.get_tree_recursive(
```
omby8888:
```suggestion
            tree, _ = await self.get_tree_recursive(
```

integrations/github/github/core/exporters/skill_exporter/core.py
omby8888: it would be same coupling u have now imo just cleaner

integrations/github/integration.py
```diff
@@ -281,6 +283,91 @@ class GithubFileResourceConfig(ResourceConfig):
...
+
+
+def _default_skill_paths() -> list[GithubSkillPattern]:
+    return [GithubSkillPattern(path=path) for path in DEFAULT_SKILL_PATHS]
+
+
+class GithubSkillSelector(Selector):
+    paths: list[GithubSkillPattern] = Field(
+        title="Paths",
+        default_factory=_default_skill_paths,
```
omby8888:
```suggestion
        default=[GithubSkillPattern(path=path) for path in DEFAULT_SKILL_PATHS],
```
although what u did works, we use those models to generate json schemas to reflect the mapping schema in the UI. since `default_factory` is runtime code, its omitted from the schema. use default instead so it'll be reflected

integrations/github/integration.py
```diff
@@ -281,6 +283,91 @@ class GithubFileResourceConfig(ResourceConfig):
...
+        title="Skill selector",
+        description="Selector for discovering and ingesting Agent Skills.",
+    )
+
+
+class GithubPluginRepoSelector(RepositorySourceModel):
+    """Org/repo scope to scan for agent plugins."""
+
+    class Config:
+        extra = "forbid"
```
omby8888: use `RepositorySourceModel` directly instead

integrations/github/integration.py
```diff
@@ -281,6 +283,91 @@ class GithubFileResourceConfig(ResourceConfig):
...
+            "codex",
+            "agents",
+            "kimi",
+            "opencode",
+            "pi",
+            "antigravity",
+        ]
+    ] = Field(
+        title="Providers",
+        default_factory=lambda: list(DEFAULT_PLUGIN_PROVIDERS),
```
omby8888:
```suggestion
        default=list(DEFAULT_PLUGIN_PROVIDERS),
```

integrations/github/integration.py
```diff
@@ -281,6 +283,91 @@ class GithubFileResourceConfig(ResourceConfig):
...
+            "claude",
+            "cursor",
+            "codex",
+            "agents",
+            "kimi",
+            "opencode",
+            "pi",
+            "antigravity",
+        ]
+    ] = Field(
```
omby8888:
```suggestion
    providers: list[Literal[PluginProvider]] = Field(
```

integrations/github/integration.py
```diff
@@ -281,6 +283,91 @@ class GithubFileResourceConfig(ResourceConfig):
...
+        title="Providers",
+        default_factory=lambda: list(DEFAULT_PLUGIN_PROVIDERS),
+        description=(
+            "Agent plugin providers to detect. A repository is treated as a "
+            "plugin when any matching manifest/dir exists."
+        ),
+    )
+    repositories: list[GithubPluginRepoSelector] = Field(
+        title="Repositories",
+        default_factory=lambda: [GithubPluginRepoSelector()],
```
omby8888:
```suggestion
        default=[GithubPluginRepoSelector()],
```

omby8888 (changes requested):

integrations/github/github/core/exporters/plugin_exporter/core.py
```diff
@@ -0,0 +1,157 @@
...
+        super().__init__(client)
+        self._file_exporter = RestFileExporter(client)
+
+    async def get_resource(self, options: Any) -> RAW_ITEM:
+        raise NotImplementedError("PluginExporter does not support get_resource")
+
+    def get_paginated_resources(self, options: Any) -> ASYNC_GENERATOR_RESYNC_TYPE:
+        raise NotImplementedError("Use get_paginated_plugins")
+
+    async def get_paginated_plugins(
```
omby8888: i think inheritance and sticking to `AbstractGithubExporter`'s contract applies here as well

integrations/github/github/core/exporters/plugin_exporter/core.py
```diff
@@ -0,0 +1,157 @@
...
+                self._build_plugin_item(
+                    organization=organization,
+                    repository=repo,
+                    branch=branch,
+                    manifest_paths=manifest_paths,
+                    providers=providers,
+                )
+                for repo, branch in repositories
+            )
+        )
```
omby8888: worth having hard set throttling here, as if repo page will increase that will affect the concurrency here directly, since here we have relatively heavy op worth having a semaphore, u can check out `semaphore_async_iterator` in `port_ocean/utils/async_iterators.py` for that

integrations/github/github/core/exporters/plugin_exporter/core.py
```diff
@@ -0,0 +1,157 @@
...
+            directory_supports=directory_supports,
+        )
+        if not plugin:
+            return PluginBuildResult(plugin_item=None, tree_truncated=truncated)
+        return PluginBuildResult(
+            plugin_item={
+                "plugin": plugin,
+                "__repository": repository,
+                "__branch": branch,
+                "__organization": organization,
```
omby8888: Saw this returning in the webhook processor, wdyt about having a pydantic model that will be the one true source of constructing plugin item and export them as dict?

integrations/github/github/core/exporters/plugin_exporter/core.py
```diff
@@ -0,0 +1,157 @@
...
+                providers=providers,
+            )
+        except Exception as exc:
+            logger.warning(
+                f"Failed to process plugin manifests for {repository.get('name')}: {exc}"
+            )
+            return None
+        return result.plugin_item
+
+    async def build_plugin_for_repo(
```
omby8888: should be private

integrations/github/github/core/exporters/plugin_exporter/core.py
```diff
@@ -0,0 +1,157 @@
...
+                    providers=providers,
+                )
+                for repo, branch in repositories
+            )
+        )
+        batch = [item for item in results if item]
+        if batch:
+            yield batch
+
+    async def _build_plugin_item(
```
omby8888: this just calls the next function with try catch, is this really needed?

integrations/github/github/core/exporters/plugin_exporter/utils.py
```diff
@@ -0,0 +1,229 @@
...
+
+def _str_field(data: dict[str, Any], *keys: str) -> Optional[str]:
+    for key in keys:
+        val = data.get(key)
+        if isinstance(val, str) and val.strip():
+            return val
+    return None
+
+
+def normalize_plugin(
```
omby8888, line 169: not sure i quite got the purpose of this function, if it's for the raw data normalization let's move it to the dedicated model.
also this is super long and each time we add support of new providers this function will need to be changed as well, could u generalize it to iterate all available providers?

integrations/github/github/core/exporters/skill_exporter/utils.py
omby8888: but now we have 2 building entry points: `build_skill_raw_item` and `build_skill_object`.
can we extract single entry point?

integrations/github/github/webhook/webhook_processors/plugin_webhook_processor.py
```diff
@@ -0,0 +1,118 @@
...
+        return WebhookEventRawResults(
+            updated_raw_results=[],
+            deleted_raw_results=[
+                {
+                    "plugin": empty_plugin(name=repo_name),
+                    "__repository": repository,
+                    "__branch": current_branch,
+                    "__organization": organization,
+                }
+            ],
```
omby8888, line 103: can we unify the way of building plugin raw object via a pydantic model maybe?

integrations/github/integration.py
```diff
@@ -281,6 +283,91 @@ class GithubFileResourceConfig(ResourceConfig):
...
+        ]
+    ] = Field(
+        title="Providers",
+        default_factory=lambda: list(DEFAULT_PLUGIN_PROVIDERS),
+        description=(
+            "Agent plugin providers to detect. A repository is treated as a "
+            "plugin when any matching manifest/dir exists."
+        ),
+    )
+    repositories: list[GithubPluginRepoSelector] = Field(
```
omby8888: not sure if repository is the best name, as this will look like:
```
repositories:
  - org: ...
    repo: ...
```

maybe paths?

omby8888 (changes requested):

integrations/github/github/core/exporters/skill_exporter/core.py
matanAlltra: Done — `SkillExporter` now subclasses `RestFileExporter` and overrides `get_paginated_resources`. `PluginExporter` implements the `AbstractGithubExporter` contract (`get_resource` / `get_paginated_resources`) but stays on composition with `RestFileExporter` rather than inheritance, because narrowing the options type would break Liskov/mypy.

integrations/github/github/core/exporters/skill_exporter/utils.py
matanAlltra: Done — single public entry point `build_skill_raw_item` (pydantic `SkillRawItem` → `model_dump`). Delete stubs and upserts both go through it; `build_skill_object` is private.

integrations/github/.port/examples/skill/a.json
matanAlltra: Done — added more examples under `.port/examples/skill/{b,c}.json` and `.port/examples/plugin/{b,c}.json`.

integrations/github/github/core/exporters/file_exporter/core.py
matanAlltra: Done — `tree, _ = await self.get_tree_recursive(...)`.

integrations/github/integration.py
matanAlltra: Done — skill `paths` now uses `default=[GithubSkillPattern(path=path) for path in DEFAULT_SKILL_PATHS]` so the UI schema reflects defaults.

integrations/github/integration.py
matanAlltra: Done — plugin scopes use `RepositorySourceModel` directly (removed the empty subclass).

integrations/github/integration.py
matanAlltra: Done — `default=list(DEFAULT_PLUGIN_PROVIDERS)`.

integrations/github/integration.py
matanAlltra: Done — `providers: list[PluginProvider]` importing the shared `PluginProvider` type.

integrations/github/integration.py
matanAlltra: Done — `default=[RepositorySourceModel()]` (after dropping the empty subclass).

integrations/github/github/core/exporters/plugin_exporter/core.py
matanAlltra: Done — plugin exporter now implements the `AbstractGithubExporter` contract with real `get_resource` / `get_paginated_resources` (composition with `RestFileExporter` for tree/content helpers).

integrations/github/github/core/exporters/plugin_exporter/core.py
matanAlltra: Done — hard-capped with `semaphore_async_iterator` (10 concurrent repos) inside the plugin exporter.

integrations/github/github/core/exporters/plugin_exporter/core.py
matanAlltra: Done — `Plugin` / `PluginRawItem` pydantic models are the single construction path; exporter and webhook both use `build_plugin_raw_item`.

integrations/github/github/core/exporters/plugin_exporter/core.py
matanAlltra: Done — `build_plugin_for_repo` is now private (`_build_plugin_for_repo`).

integrations/github/github/core/exporters/plugin_exporter/core.py
matanAlltra: Done — removed the thin try/except wrapper; error handling is inlined at the call site.

integrations/github/github/core/exporters/plugin_exporter/utils.py
matanAlltra: Done — `normalize_plugin` now iterates providers via shared manifest/dir maps into the pydantic `Plugin` model (marketplace / name-precedence edge cases preserved).

integrations/github/github/webhook/webhook_processors/plugin_webhook_processor.py
matanAlltra: Done — unified via `PluginRawItem` pydantic model + `build_plugin_raw_item`.

integrations/github/integration.py
matanAlltra: Done — renamed plugin selector field `repositories` → `paths`.

integrations/github/main.py
```diff
@@ -1092,6 +1109,97 @@ async def resync_files(
...
+                # The exporter caps how many repositories it scans concurrently,
+                # so the whole organization can be handed over in one call.
+                repositories = [
+                    PluginRepositoryOptions(
+                        organization=org_login, repository=repo_obj, branch=branch
+                    )
+                    async for _repo_name, branch, repo_obj in repo_selector.select_repos(
+                        path, repo_exporter, org_login, org["type"]
+                    )
+                ]
```
omby8888:
```suggestion
                repositories: list[PluginRepositoryOptions] = []
                async for _repo_name, branch, repo_obj in repo_selector.select_repos(
                    path, repo_exporter, org_login, org["type"]
                ):
                  repositories.append(PluginRepositoryOptions(
                      organization=org_login, repository=repo_obj, branch=branch
                  ))
```

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
```diff
@@ -0,0 +1,137 @@
...
+            "repository", {}
+        )
+
+    async def _should_process_event(self, event: WebhookEvent) -> bool:
+        is_push_event = event.headers.get("x-github-event") == "push"
+        is_github_private_repo = (
+            event.payload.get("repository", {}).get("name") == ORG_CONFIG_REPO
+        )
+        has_branch_name = event.payload.get("ref", "").startswith("refs/heads/")
+        return is_push_event and not is_github_private_repo and has_branch_name
```
omby8888: both of those returns across file kind related webhook processors, can they inherit from it instead of duplicating the code?

integrations/github/github/webhook/webhook_processors/base_repository_webhook_processor.py
omby8888: still think inheritance is the way to go here, as this is copied directly from the file webhook processor instead of using it ootb via inheritence

integrations/github/github/webhook/webhook_processors/plugin_webhook_processor.py
```diff
@@ -0,0 +1,116 @@
...
+            "repository", {}
+        )
+
+    async def _should_process_event(self, event: WebhookEvent) -> bool:
+        is_push_event = event.headers.get("x-github-event") == "push"
+        is_github_private_repo = (
+            event.payload.get("repository", {}).get("name") == ORG_CONFIG_REPO
+        )
+        has_branch_name = event.payload.get("ref", "").startswith("refs/heads/")
+        return is_push_event and not is_github_private_repo and has_branch_name
```
omby8888: same as https://github.com/port-labs/ocean/pull/3532/changes#r3658244918

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
matanAlltra: Fixed — `SkillWebhookProcessor` now inherits from `FileWebhookProcessor`, so push validation / `_should_process_event` come from there instead of being duplicated.

integrations/github/github/webhook/webhook_processors/plugin_webhook_processor.py
matanAlltra: Fixed — same as above; `PluginWebhookProcessor` now inherits from `FileWebhookProcessor`.

integrations/github/github/webhook/webhook_processors/base_repository_webhook_processor.py
matanAlltra: Fixed — removed the duplicated `_is_applicable_to_repo_branch` from `BaseRepositoryWebhookProcessor`. Skill/plugin webhooks now use `FileWebhookProcessor._is_pattern_applicable_to_branch` via inheritance.

integrations/github/main.py
matanAlltra: Fixed — `resync_plugins` already collects `PluginRepositoryOptions` into a list for the whole org before calling `get_paginated_resources` (async list comprehension).

matanAlltra: Fixed — switched to the explicit `async for` + `append` collection style as suggested.

integrations/github/main.py
```diff
@@ -1097,6 +1114,100 @@ async def resync_files(
...
+    repo_selector = CompositeRepositorySelector(app_config.repository_type)
+
+    for path in paths:
+        async for orgs in org_exporter.get_paginated_resources(
+            ListOrganizationOptions(organization=path.organization)
+        ):
+            for org in orgs:
+                org_login = org["login"]
+                # The exporter caps how many repositories it scans concurrently,
+                # so the whole organization can be handed over in one call.
```
omby8888: remove this comment

omby8888 (approved):
Nice job!!

integrations/github/main.py
matanAlltra: Fixed — removed the comment.

PR #3547, opened 2026-07-17
[Integration][Snyk] Fix serialization of array query parameters

eriport:
# Description

What - 

Fix query parameter serialization for array values. Array query parameters are now sent as comma-separated strings instead of repeated query parameters.

Why - 

Snyk REST API expects non-exploded array query parameters (for example, `status=open,resolved`). 
We were sending repeated parameters (for example, `status=open&status=resolved`), which caused the API to return HTTP 400 and prevented vulnerability resyncs when multiple filter values were configured.

How -

Update `GenerateQueryParamMixin.generate_query_params()` to convert list values into comma-separated strings before sending requests with `httpx`. This applies to all Snyk API query parameter models that use the mixin.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Changes query-string shape for all Snyk API filters using the mixin; wrong serialization for booleans or edge cases could still break requests. Existing unit tests in `test_overrides.py` still assert raw lists/bools and may need updating to match the new output.
> 
> **Overview**
> Fixes Snyk REST calls that returned **HTTP 400** when selectors used multi-value filters (e.g. `status`, `effective_severity_level`).
> 
> `GenerateQueryParamMixin.generate_query_params()` no longer passes Python lists through to `httpx` (which produced repeated keys like `status=open&status=resolved`). **List values are now joined with commas** (`status=open,resolved`), and non-list values are sent as strings. That applies to every API query-param model that uses the mixin (projects, vulnerabilities, policies, targets).
> 
> Release **0.5.52** with a changelog entry; integration version bumped in `pyproject.toml`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff71658c441975e22c97c0e105437c733096f31e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/snyk/snyk/overrides.py
```diff
@@ -12,7 +12,10 @@
 class GenerateQueryParamMixin(BaseModel):
     def generate_query_params(self) -> dict[str, Any]:
         params = self.dict(exclude_none=True, exclude_unset=True)
-        return params
+        return {
+            key: ",".join(map(str, value)) if isinstance(value, list) else str(value)
```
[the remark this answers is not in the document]
danamsalem-ux: I agree with this comment, is that case handled?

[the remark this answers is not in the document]
eriport: yes it is now

integrations/snyk/snyk/overrides.py
```diff
@@ -12,7 +12,11 @@
 class GenerateQueryParamMixin(BaseModel):
     def generate_query_params(self) -> dict[str, Any]:
         params = self.dict(exclude_none=True, exclude_unset=True)
-        return params
+        return {
+            key: ",".join(map(str, value)) if isinstance(value, list) else value
+            for key, value in params.items()
+            if not isinstance(value, list) or value
```
raskinmaya, line 18: This only joins lists coming out of the selector models. Pagination still goes through `parse_next_page_params`, which rebuilds repeated keys into lists (look at utils.py around lines 13-15; tests even assert `status=open&status=resolved → ["open","resolved"]`). Those lists get passed straight to httpx on page 2+ and explode again, which leads to the same 400.

Prefer normalizing list -> comma once in `_send_api_request` (or right before `params=`). That covers first page, webhooks, and `links.next` in one place. Mixin-only is the fragile half-fix

integrations/snyk/snyk/overrides.py
```diff
@@ -12,7 +12,11 @@
...
+        }
```
raskinmaya, line 20: If you keep the mixin join, also join (or reject) list values at the client boundary. Right now transport encoding is split across config models and httpx defaults, which is why this bug existed. One serializer at the request edge is the simpler long-term fix (same pattern as joining in azure-devops / github dependabot, but centralized)

integrations/snyk/tests/snyk/test_overrides.py
```diff
@@ -25,8 +29,8 @@ def test_generate_query_params_preserves_list_values() -> None:
         environment=["frontend", "backend"],
```
raskinmaya: `test_generate_query_params_preserves_list_values` no longer preserves lists - could you rename it? It also overlaps `test_generate_query_params_joins_list_values_with_commas` and `test_policy_…_preserves_list_review`. Keep one parametrized test covering project + vuln + policy list fields; drop the duplicates

integrations/snyk/tests/snyk/test_overrides.py
```diff
@@ -1,6 +1,10 @@
 import pytest
+from typing import Literal
 from pydantic.v1 import ValidationError
-from snyk.overrides import SnykTargetAPIQueryParams, TargetSelector
+from snyk.overrides import (
+    SnykTargetAPIQueryParams,
+    TargetSelector,
+)
 
 from snyk.overrides import (
```
raskinmaya, line 4: Two separate `from snyk.overrides import (...)` blocks. Let's merge into one

integrations/snyk/snyk/overrides.py
```diff
@@ -12,7 +12,11 @@
 class GenerateQueryParamMixin(BaseModel):
     def generate_query_params(self) -> dict[str, Any]:
         params = self.dict(exclude_none=True, exclude_unset=True)
-        return params
+        return {
+            key: ",".join(map(str, value)) if isinstance(value, list) else value
```
raskinmaya: `map(str, value)` isn’t needed. every list field here is already `list[str]`/`list[Literal[...]]`. `",".join`(value) is enough. The `if not isinstance(value, list)` or value filter is correct but hard to read; a short loop (omit empty lists, join non-empty) would be clearer for the same behavior

integrations/snyk/snyk/overrides.py
```diff
@@ -12,7 +12,11 @@
 class GenerateQueryParamMixin(BaseModel):
```
raskinmaya, line 12: `tags` is `key=value` and snyk uses commas as the multi-value delimiter. Joining is correct for their contract, but any tag/name value with an embedded `,` becomes ambiguous. Worth a one-line note in the field description that values must not contain commas

integrations/snyk/snyk/overrides.py
raskinmaya: @eriport

integrations/snyk/snyk/overrides.py
raskinmaya: @eriport

integrations/snyk/snyk/overrides.py
raskinmaya: Resolved — addressed in later commits:

1. `_send_api_request` now normalizes list values to comma-joined strings right before `params=` (covers first page, pagination, webhooks).
2. `parse_next_page_params` also joins repeated keys (`status=open&status=resolved` → `"open,resolved"`) so page 2+ no longer rebuilds exploded lists.

Covered by `test_parse_next_page_params_joins_repeated_keys_with_commas`.

integrations/snyk/snyk/overrides.py
raskinmaya: Resolved — list → comma serialization now happens at the request edge in `_send_api_request`, in addition to the mixin. That keeps transport encoding centralized regardless of whether params come from selectors or `links.next`.

integrations/snyk/snyk/overrides.py
raskinmaya: Resolved — `tags` field description now includes: \"Tag values must not contain commas.\"

integrations/snyk/snyk/overrides.py
```diff
@@ -12,7 +12,10 @@
...
     def generate_query_params(self) -> dict[str, Any]:
         params = self.dict(exclude_none=True, exclude_unset=True)
-        return params
+        return {
+            key: ",".join(map(str, value)) if isinstance(value, list) else str(value)
+            for key, value in params.items()
+        }
```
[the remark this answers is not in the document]
raskinmaya: Resolved — scalars are no longer stringified; only non-empty lists are joined. Booleans stay as bools (see `test_generate_query_params_preserves_boolean_values_not_stringified`).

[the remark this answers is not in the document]
raskinmaya: Resolved — empty lists are omitted via \`if not isinstance(value, list) or value\` in the mixin, covered by \`test_generate_query_params_excludes_empty_lists\`.

raskinmaya (approved):
Re-reviewed after the follow-up commits. Previous comments are addressed:

- List params are normalized at the request edge in `_send_api_request`
- `parse_next_page_params` joins repeated keys so pagination no longer re-explodes arrays
- Empty lists are omitted; booleans stay as bools
- `tags` field notes that values must not contain commas
- Tests cover the serialization behavior

LGTM.

raskinmaya (approved):

PR #3548, opened 2026-07-17
[Integration][Gitlab-v2] Remove dependency on unstable Gitlab error message

melodyogonna:
# Description

What - Gitlab error messages can apparently change without warning so isn't a good candidate to base behaviour on.

Why - Changed error message broke group search fallback.

How - Just depend on only the status code.

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Broader fallback on all 400s can hide some invalid requests and add extra API work via project-level search, but it restores reliable file discovery when GitLab changes error text.
> 
> **Overview**
> **Group file search** no longer inspects GitLab’s **400 response body** to decide whether blob scope is unsupported. **`_is_blob_search_unavailable`** was removed; **`_search_files_in_group`** now treats **any HTTP 400** from the group search API as a trigger to **fall back to per-project search** via **`_search_files_in_group_projects`**. Other status codes are still re-raised.
> 
> Tests were updated so a generic 400 from group search **invokes the fallback** instead of failing immediately; errors surface only if the fallback path also fails.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f91648bf14b210f6810c2dcf8f953567418b30e3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/gitlab-v2/gitlab/clients/gitlab_client.py
```diff
@@ -967,7 +967,7 @@ async def _search_files_in_group(
                 if processed_batch:
                     yield processed_batch
         except httpx.HTTPStatusError as e:
-            if not self._is_blob_search_unavailable(e):
+            if e.response.status_code != 400:
```
raskinmaya: This change makes us fall back on every `400`, not just the known gitlab `"Scope 'blobs' is not available for this search"` case. That can hide real bad-request errors and turn one failed group search into an expensive project-by-project scan on large groups. Can we keep the fallback scoped to the specific gitlab limitation instead of all `400`s?

integrations/gitlab-v2/tests/clients/test_gitlab_client.py
```diff
@@ -857,11 +857,16 @@ async def test_search_files_in_group_other_400_raises(
             "get_paginated_resource",
             side_effect=error,
         ):
-            with pytest.raises(httpx.HTTPStatusError):
-                async for _ in client._search_files_in_group(
-                    "my-group", "blobs", "test.json"
-                ):
-                    pass
+            with patch.object(
```
raskinmaya, line 860: This test doesn't actually prove the fallback happened. It would still pass if `_search_files_in_group()` re-raised the original `400` before ever calling `_search_files_in_group_projects()`. Can we assert that the fallback method was called, or make it return a sentinel result so the intended behavior is pinned down?

integrations/gitlab-v2/gitlab/clients/gitlab_client.py
melodyogonna: @raskinmaya yeah that is the issue I was trying to avoid by scoping to an error message. We got reports that the search no longer worked, when I looked into it the error message had changed to something else, which broke the fallback. We could continue scoping to specific error messages but with the full knowledge that that is brittle and we'd have to continually update the error message guard as Gitlab evolves.

melodyogonna: I updated the code to continue depending on error messages, the old error message is maintained for Gitlab customers on self-hosted who may not be getting the updated error message.

integrations/gitlab-v2/tests/clients/test_gitlab_client.py
melodyogonna: With the revert to error message checking this test case is no longer necessary and has been removed.

raskinmaya (approved):

PR #3572, opened 2026-07-20
[Integration][GitHub] Migrate callers to create_github_client_for_org

omby8888:
## Summary
- Use authenticator-scoped `create_github_client()` for resyncs, with `create_github_client_for_org()` retained for webhook and action requests.
- Discover GitHub App installations as organization-scoped authenticators and run each resync against its allowed organization; PATs use `github_organization` as their optional scope.
- Centralize file and folder selector access filtering per authenticator, keep organization cache entries isolated by rate-limit scope, and make installation lookups case-insensitive.

Follow-up to #3495 (task_tgtuzt).

## Test plan
- [x] `make test` in `integrations/github` (895 passed)


integrations/github/main.py
```diff
@@ -161,40 +159,73 @@ async def on_start() -> None:
...
+        async for organizations in org_exporter.get_paginated_resources():
+            logger.info(
+                f"Subscribing to GitHub webhooks for {len(organizations)} organizations"
+            )
+
+            for org in organizations:
+                await _create_webhooks_for_organization(org["login"], base_url)
+
+
+def _resync_per_authenticator(
```
raskinmaya, line 170: I’m wondering if we can simplify the org-scoping story a bit:
right now resolution seems split across `authenticator.organization`, integration config, `can_access_organization`, and the pattern-builder filters. Wouldn't it be cleaner if every authenticator owned its scope (PAT from `github_organization` when set; app from the installation), `resync` only used `create_github_client(authenticator)`, and `create_github_client_for_org` stayed as the webhook/action entrypoint?
also regarding saas: with oauth no longer publishing installation/org, are we intentionally discovering every install on the shared app, or should discovery stay scoped to the oauth installation?

integrations/github/.port/spec.yaml
```diff
@@ -89,8 +84,6 @@ saas:
       integrationSpec:
         githubHost: "'https://api.github.com'"
         githubAppId: .oauthData.appId
```
raskinmaya, line 86: Not sure if we spoke about it, but removing both `githubAppInstallationId` and `githubOrganization` from oauth `valuesOverride` means a shared github app (`appId` + PEM) will discover every installation via `GET /app/installations`. That means that one customer’s integration can sync/create webhooks for other orgs that installed the same app. IMO you should keep publishing `installationId` (or at least `githubOrganization`) from `oauthData`, and scope discovery to that installation, and not all installs for this app

integrations/github/github/clients/auth/github_app/installation_registry.py
```diff
@@ -53,11 +55,21 @@ async def _fetch_installations() -> dict[str, GitHubAppInstallationAuthenticator
                     f"Installation {installation} skipped: organization login missing or installation suspended."
                 )
                 continue
+            if (
+                configured_organization
+                and login.casefold() != configured_organization.casefold()
```
raskinmaya, line 60: Fetch filters with `casefold`, but lookup is exact `organization not in _authenticators_by_org`. Config/`create_github_client_for_org("MyOrg")` fails when github's login is `myorg`. Index and resolve with a casefold key (or normalize once on insert). TBH I even think you should add a UT for this case

integrations/github/github/core/exporters/organization_exporter.py
```diff
@@ -1,87 +1,90 @@
...
             )
+            if organization_resource:
+                yield [organization_resource]
+            return
 
         async for batch in self._stream_selected_organizations(
-            allowed_multi_organizations, include_authenticated_user
+            allowed_multi_organizations, include_authenticated_user, _auth_scope
         ):
             yield batch
```
raskinmaya: Dropping the classic-PAT/`OrganizationRequiredException` path regresses fine-grained PATs; spec still says `githubOrganization` is for fine-grained PAT; without it this falls through to `GET /user/orgs`, which often returns `[]` for fine-grained tokens (see [here](https://docs.github.com/en/rest/orgs/orgs#list-organizations-for-the-authenticated-user)), so resync succeeds with no data instead of failing. IMO you should keep an explicit "org required" failure when the token can’t list orgs (or when `/user/orgs` is empty and no org is configured)

integrations/github/github/core/exporters/organization_exporter.py
```diff
@@ -1,87 +1,90 @@
...
+            and requested_organization.casefold()
+            != authenticated_organization.casefold()
+        ):
+            raise AuthenticationException(
+                f"Authenticator for '{authenticated_organization}' cannot access "
+                f"requested organization '{requested_organization}'"
+            )
 
-        if organization := options.get("organization"):
+        if organization := (
```
raskinmaya: Org scope is resolved in three places: `authenticator.organization`, `ocean.integration_config["github_organization"]`, and `can_access_organization`/pattern-builder filters. PAT never sets `authenticator.organization`, so PAT vs app behave differently. Can't we use a single source? e.g. set organization on the PAT authenticator from config in the auth provider, and drop the config fallback + duplicate filters in exporters/builders

integrations/github/github/core/exporters/folder_exporter/utils.py
```diff
@@ -30,26 +28,33 @@ async def build(self, folders: List[FolderSelector]) -> List[ListFolderOptions]:
...
-                    key = (org_login, repo_name)
-                    repo_map[key].append(
-                        FolderSearchOptions(
-                            organization=org_login,
-                            branch=branch,
-                            path=folder_sel.path,
-                            repo=repo_obj,
+                for org in batch:
+                    org_login = org["login"]
+                    if (
```
raskinmaya: You already pass `ListOrganizationOptions(organization=...)`, and the exporter either returns that org or raises. The extra `casefold()` continue is dead. Let's delete it. Applies [here](https://github.com/port-labs/ocean/pull/3572/changes#diff-e59a74032646eb667f6d0895139e9aea79edd81ff6031d3190375fc0dcfdc280R158) as well

integrations/github/github/core/exporters/organization_exporter.py
```diff
@@ -1,87 +1,90 @@
...
+        ):
+            yield batch
 
     @cache_iterator_result()
-    async def get_paginated_resources(
-        self, options: ListOrganizationOptions
+    async def _get_paginated_resources(
+        self,
+        options: ListOrganizationOptions | None,
+        _auth_scope: str,
```
raskinmaya, line 30: the dummy `_auth_scope` only exists because `cache_iterator_result` drops `self` from the hash - easy to "clean up" and reintroduce cross-installation cache bleed. Ideally I'd want a first-class cache key (or stop caching org listing across authenticators this way)

integrations/github/CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- towncrier release notes start -->
 
+
+## 6.5.0 (2026-07-23)
+
+
+### Improvements
+
+- Migrated all GitHub API callers from legacy `create_github_client()` to installation-scoped `create_github_client_for_org()`.
```
raskinmaya: this mentions callers moved to `create_github_client_for_org()`, but resync still uses `create_github_client(authenticator)` - the old public name, new signature. That’s a bit confusing. Let's rephrase it: "authenticator-scoped `create_github_client` + org helper for webhooks/actions", wdyt?

integrations/github/github/clients/auth/github_app/installation_authenticator.py
raskinmaya: `installation_id` is required. `_get_installation_id`’s fallback to `get_installation_id_for_organization` is effectively dead. Could you trim the fallback and fix the docstring?

integrations/github/github/actions/dispatch_workflow_executor.py
```diff
@@ -125,7 +125,15 @@ async def _get_partition_key(self, run: IntegrationRun) -> str | None:
 
         return f"{organization}/{repo}/{workflow}"
 
-    async def _get_default_ref(self, organization: str, repo_name: str) -> str:
+    async def _get_execution_client(self, run: IntegrationRun) -> GithubRestClient:
+        organization = run.execution_properties.get("org")
+        if not isinstance(organization, str):
+            raise InvalidActionParametersException("organization is required")
```
raskinmaya: this says "organization is required" but [here](https://github.com/port-labs/ocean/pull/3572/changes#diff-9ff2bea38b82b34bf809e5465478c97da96f38342b9e4c1aeae5c45937536064R51) it's "org is required". let's align them

integrations/github/github/core/options.py
```diff
@@ -5,12 +5,8 @@
 from pydantic.v1 import BaseModel, Field
 
 
-class ListOrganizationOptions(TypedDict):
-    """Options for listing organizations."""
-
-    organization: NotRequired[str]
-    allowed_multi_organizations: NotRequired[List[str]]
-    include_authenticated_user: NotRequired[bool]
+class ListOrganizationOptions(BaseModel):
```
raskinmaya, line 8: Every options type here is a `TypedDict`; only `ListOrganizationOptions` became a pydantic model. That forces `.organization` vs `["organization"]` and doesn’t buy validation. Why not use a `TypedDict`?

integrations/github/main.py
```diff
@@ -161,40 +159,73 @@ async def on_start() -> None:
...
+    return wrapper
 
 
 @ocean.on_resync(ObjectKind.ORGANIZATION)
-async def resync_organizations(kind: str) -> ASYNC_GENERATOR_RESYNC_TYPE:
+@_resync_per_authenticator
+async def resync_organizations(
+    kind: str, authenticator: AbstractGitHubAuthenticator
+) -> ASYNC_GENERATOR_RESYNC_TYPE:
     """Resync all organizations the Personal Access Token user is a member of."""
```
raskinmaya: This isn't PAT-only now

integrations/github/main.py
omby8888: in Saas (both OAuth or not) we should use installations discovery

integrations/github/.port/spec.yaml
omby8888: why not? The purpose is that we'll always use the discovery path, if customer wants to opt in multi org he can simply install the app on another org, rather than delete those attributes now in the integration.
Do u think we should still have a path in App authentication for scoped single org rather than discovery? seem redundant to me, i'm also consider removing the existing github_org ref in the installation registry

integrations/github/github/core/exporters/organization_exporter.py
omby8888: i'm fine with it return 401, this wasn't handle before as well

integrations/github/github/core/exporters/organization_exporter.py
omby8888: I'll add a UT that verifies we cache properly but this is a common pattern in integrations, considering self in the hashing will introduce performance degradation in some integrations as they rely on different instances to cache similarly

integrations/github/github/core/options.py
omby8888: BaseModel much more cleaner, ideally i'de migrate all to basemodel but it's out of scope

integrations/github/github/core/exporters/organization_exporter.py
raskinmaya: spoke f2f

raskinmaya (approved):
50 ILS pls

PR #3585, opened 2026-07-21
[Integration][Github][Gitlab-v2][Claude-Managed-Agents][Core]Pass run context to action rate limit checks

omby8888:
## Summary
- require action executors to accept the integration run when checking rate limits
- pass the run from the execution manager to each rate-limit check
- migrate GitHub, GitLab v2, and Claude Managed Agents executors to the new interface
- bump core to 0.45.8 without changing integration versions

## Testing
- `. .venv/bin/activate && pytest port_ocean/tests/core/handlers/actions/test_execution_manager.py`
- `cd integrations/claude-managed-agents && PYTHONPATH=../.. .venv/bin/python -m pytest tests/test_actions.py`
- Ruff and Black on affected files

### DIFF
--- a/integrations/github/github/actions/abstract_github_executor.py
+++ b/integrations/github/github/actions/abstract_github_executor.py
@@ -1,5 +1,6 @@
 from github.clients.client_factory import create_github_client
 from port_ocean.core.handlers.actions.abstract_executor import AbstractExecutor
+from port_ocean.core.models import IntegrationRun
 
 MIN_REMAINING_RATE_LIMIT_FOR_EXECUTE_WORKFLOW = 20
 
@@ -8,14 +9,16 @@ class AbstractGithubExecutor(AbstractExecutor):
     def __init__(self) -> None:
         self.rest_client = create_github_client()
 
-    async def is_close_to_rate_limit(self) -> bool:
+    async def is_close_to_rate_limit(self, run: IntegrationRun) -> bool:
         info = self.rest_client.get_rate_limit_status()
         if not info:
             return False
 
         return info.remaining < MIN_REMAINING_RATE_LIMIT_FOR_EXECUTE_WORKFLOW
 
-    async def get_remaining_seconds_until_rate_limit(self) -> float:
+    async def get_remaining_seconds_until_rate_limit(
+        self, run: IntegrationRun
+    ) -> float:
         info = self.rest_client.get_rate_limit_status()
         if not info:
             return 0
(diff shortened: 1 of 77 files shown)


integrations/github/github/actions/abstract_github_executor.py
```diff
@@ -8,14 +9,16 @@ class AbstractGithubExecutor(AbstractExecutor):
     def __init__(self) -> None:
         self.rest_client = create_github_client()
 
-    async def is_close_to_rate_limit(self) -> bool:
+    async def is_close_to_rate_limit(self, run: IntegrationRun) -> bool:
         info = self.rest_client.get_rate_limit_status()
```
raskinmaya, line 13: This still builds a legacy `create_github_client()` and reads that cache. App rate limits are per installation; actions already pass org in `execution_properties`. Multi-org can gate (or fail to gate) on the wrong bucket and then hit 403/secondary limits mid-execute. If run-aware is indeed the goal, I think you should resolve the client from run before checking status

port_ocean/tests/core/handlers/actions/test_execution_manager.py
```diff
@@ -311,6 +311,12 @@ async def test_execute_run_should_sleep_if_rate_limited(
             level="WARNING",
             should_raise=False,
         )
+        mock_test_executor.is_close_to_rate_limit.assert_called_with(
+            mock_test_action_run
+        )
+        mock_test_executor.get_remaining_seconds_until_rate_limit.assert_called_once_with(
+            mock_test_action_run
+        )
```
raskinmaya, line 319: `assert_called_once_with` on a method inside a wall-clock while backoff loop can still flake if the loop runs twice. Prefer deterministic `side_effect=[True, False]` + mocked `asyncio.sleep`. Claude tests passing bare `MagicMock()` also underline that nothing asserts real run-aware behavior

integrations/github/github/actions/abstract_github_executor.py
```diff
@@ -8,14 +9,16 @@ class AbstractGithubExecutor(AbstractExecutor):
     def __init__(self) -> None:
         self.rest_client = create_github_client()
 
-    async def is_close_to_rate_limit(self) -> bool:
+    async def is_close_to_rate_limit(self, run: IntegrationRun) -> bool:
```
raskinmaya, line 12: `run` is required but unused here. Intentional?

omby8888: yeah, will be used in the multi org PR

integrations/github/github/actions/abstract_github_executor.py
omby8888: bruh this is not the same PR, multi org references will be on that PR, this PR only enables the per org logic for this case

port_ocean/tests/core/handlers/actions/test_execution_manager.py
omby8888: because there's no run aware logic there, this PR only enables to treat the run in the logic

raskinmaya (approved):

PR #3587, opened 2026-07-21
[Integration][Cursor-Cloud-Agents] Add Cursor Cloud Agents integration

talsabagport:
# Description

## What

Add a new Ocean SaaS integration for [Cursor Cloud Agents](https://cursor.com/docs/cloud-agents) that syncs agents and runs into Port, supports workflow actions to create and trigger agents, and completes Port runs via v0 webhooks when `reportCompletion` is enabled.

## Why

Port users need to manage Cursor cloud agents in the catalog and drive them from workflows (launch agents, send follow-ups, track completion) without building custom glue. This mirrors the Cursor agent flow in workflow-service and exposes it as a first-class Ocean integration with live events and action processing.

## How

- **Catalog sync** - Resync exporters for `agent` and `run` via the Cursor v1 API. Runs are enriched with per-run token usage from `GET /v1/agents/{id}/usage` (fetched in parallel with run details in webhook/catalog paths).
- **Actions** - `create_agent` supports `apiVersion` `v0` or `v1` (default `v1`). `trigger_agent` always uses the v1 follow-up API. Request bodies are validated with Pydantic models aligned to the Cursor API, with Port-only policy where needed (`reportCompletion`, repository requirements, etc.).
- **`reportCompletion`** - On v0 create with `reportCompletion: true`, registers a per-run webhook (`OCEAN__BASE_URL` required) and keeps the Port run `IN_PROGRESS` until a terminal webhook. On trigger, completion tracking only applies when the agent was originally created with v0 + `reportCompletion`. v1 create has no webhooks and always completes the Port run after launch.
- **Webhooks** - v0 agent status callbacks are verified with HMAC. The signing secret is derived per run from `OCEAN__PORT__CLIENT_SECRET`, org id, and the Port run id in the callback URL (no customer webhook secret). Terminal statuses: `FINISHED`, `ERROR`, `CANCELLED`, `EXPIRED`. Catalog entities are upserted even when no in-progress Port run is tracked.
- **Run correlation** - Webhook payloads carry no run id; the processor resolves the Cursor run from List Runs at or before the webhook timestamp, then matches the Port run by Cursor run id (follow-up) or agent id (create).
- **Defaults** - Blueprints, port-app-config mappings, and `.port/spec.yaml` for SaaS install with `cursorApiKey` and live events enabled.
- **Tests** - 110 tests covering actions, validation, exporters, webhooks, catalog normalization, and signing.

## Type of change

- [x] Docs PR link [here](https://github.com/port-labs/port-docs/pull/4528)

### Preflight checklist

- [x] Handled rate limiting
- [x] Handled pagination
- [x] Support Multi account - N/A.

## Screenshots

image
image
image

## API Documentation
https://cursor.com/docs/cloud-agent/api/endpoints
https://cursor.com/docs/cloud-agent/api/v0

---

> [!NOTE]
> **Medium Risk**
> New integration with webhook HMAC tied to Port client secret and workflow run completion logic; mistakes could mis-report run status or mishandle callbacks, but it is additive and does not change existing integrations.
> 
> **Overview**
> Adds a new **Port Ocean** integration under `integrations/cursor-cloud-agents` that syncs **Cursor Cloud Agents** into Port and exposes workflow actions to launch and follow up on agents.
> 
> **Catalog sync** maps `agent` and `run` kinds to `cursor_agent` / `cursor_run` blueprints (status, repos, PR URLs, token usage, agent relation). Resync uses the v1 List Agents API and fans out List Runs per agent, merging usage from `GET /v1/agents/{id}/usage`. Pagination failures raise `CursorAgentsPaginationError` so Ocean skips delete-on-partial-sync.
> 
> **Workflow actions** `create_agent` (v0 or v1 create APIs) and `trigger_agent` (v1 follow-ups, partitioned by `agentId`) validate bodies with Pydantic, shallow-merge `config`, and best-effort upsert catalog entities after each API call. **v0 + `reportCompletion`** registers a per-run webhook (`OCEAN__BASE_URL`), derives HMAC secrets from Port client secret + org id + run id, and leaves the Port run in progress until `CursorAgentWebhookProcessor` handles terminal `FINISHED`/`ERROR` callbacks. **v1 create** rejects `reportCompletion`; follow-up completion tracking only applies when the agent was v0-created with a registered webhook.
> 
> SaaS live events, Port resource definitions (`.port/spec.yaml`, blueprints, mappings), client with retryable HTTP, and broad pytest coverage are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc147205f46ce6d766e20bc825668461f8a0c824. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/cursor-cloud-agents/.port/spec.yaml
```diff
@@ -0,0 +1,123 @@
...
+  - name: cursorApiKey
+    type: string
+    required: true
+    sensitive: true
+    description: Cursor API key used to authenticate with the Cloud Agents API. Generate one from Cursor Dashboard -> API Keys.
+  - name: webhookSigningSecret
+    type: string
+    required: false
+    sensitive: true
+    description: Optional secret used to sign and verify Cursor v0 agent status webhooks. When set, the integration derives a per-run signing secret and validates the X-Webhook-Signature header on incoming webhooks. When unset, webhook signature verification is skipped. Generate a random string (for example with openssl rand -hex 32) and store it as an organization secret during installation.
```
omby8888:
```suggestion
    description: Optional secret used to sign and verify Cursor v0 agent status webhooks. When set, the integration derives a per-run signing secret and validates the X-Webhook-Signature header on incoming webhooks. When unset, webhook signature verification is skipped.
```
This sentence is not true for self hosted installations as we don't host that data

integrations/cursor-cloud-agents/.port/spec.yaml
```diff
@@ -0,0 +1,123 @@
...
+    sensitive: false
+    default: https://api.cursor.com
+    description: The host of the Cursor API. If not specified, the default will be https://api.cursor.com.
+  - name: cursorConsoleHost
+    type: url
+    required: false
+    sensitive: false
+    default: https://cursor.com
+    description: The host of the Cursor console, used to build agent links in action run output. If not specified, the default will be https://cursor.com.
+  - name: cursorPageSize
```
omby8888: Users don’t need to choose pagination size, we should manage it to prevent inefficient or unstable syncs.

integrations/cursor-cloud-agents/core/exporters/abstract_exporter.py
```diff
@@ -0,0 +1,18 @@
...
+
+class AbstractCursorExporter(ABC):
+    """Abstract base class for Cursor Cloud Agents resource exporters."""
+
+    def __init__(self, client: CursorAgentsClient) -> None:
+        self.client = client
+
+    @abstractmethod
+    def get_paginated_resources(
+        self, *, include_archived: bool = False
```
omby8888: can we make the signature with generic options model so that it'll be exendable?
look in github for example, each exporter has corresponding options model containing the selector possibilities

integrations/cursor-cloud-agents/core/exporters/runs_exporter.py
```diff
@@ -0,0 +1,50 @@
...
+    async def get_paginated_resources(
+        self, *, include_archived: bool = False
+    ) -> ASYNC_GENERATOR_RESYNC_TYPE:
+        query_params = {"includeArchived": include_archived}
+        async for agents_batch in self.client.paginate_by_cursor(
+            V1_AGENTS, "items", params=query_params
+        ):
+            for agent in agents_batch:
+                agent_id = agent.get("id")
+                if not agent_id:
```
omby8888: this should use the agent exporter instead imo

integrations/cursor-cloud-agents/core/exporters/agents_exporter.py
```diff
@@ -0,0 +1,19 @@
...
+from port_ocean.core.ocean_types import ASYNC_GENERATOR_RESYNC_TYPE
+
+from clients.endpoints import V1_AGENTS
+from core.exporters.abstract_exporter import AbstractCursorExporter
+
+
+class AgentsExporter(AbstractCursorExporter):
+    """Syncs the `agent` kind from the v1 List Agents API."""
+
+    async def get_paginated_resources(
```
omby8888, line 16: since those yielded batches are required to fetch the runs or any other future kinds, lets cache their results using `@cache_iterator_result`

integrations/cursor-cloud-agents/core/exporters/runs_exporter.py
```diff
@@ -0,0 +1,50 @@
...
+                    continue
+                usage_by_run_id = await fetch_usage_by_run_id(self.client, agent_id)
```
omby8888: i wonder whether this should be configurable selector like `enrich_runs_with_usage=true`, is there a value for only the runs without the token usage? asking in terms of rate limit budget saving

integrations/cursor-cloud-agents/core/exporters/runs_exporter.py
```diff
@@ -0,0 +1,50 @@
...
+                async for runs_batch in self.client.paginate_by_cursor(
```
omby8888, line 84: we would bring tons of data over time as i guess this those runs are often, wdyt about having a selector of absolute utc date that specifies the oldest date to get runs on. since this api is sorted desc by creation date, when we'll run upon the selector date we will stop the pagination

integrations/cursor-cloud-agents/webhook_processors/utils.py
```diff
@@ -0,0 +1,109 @@
...
+    except Exception as error:
+        logger.warning(
+            f"Failed to list Cursor runs for agent {agent_id} "
+            f"while resolving webhook run id: {error}"
+        )
+        return None
+    return resolve_cursor_run_id_from_runs(runs, webhook_time)
+
+
+async def resolve_tracked_run(
```
omby8888: i think it would be cleaner to do this check in place instead of having this multi responsibility function.

integrations/cursor-cloud-agents/core/catalog.py
```diff
@@ -0,0 +1,178 @@
...
+    raw: dict[str, Any],
+    *,
+    console_host: str | None = None,
+) -> None:
+    try:
+        await upsert_raw_entity(kind, raw, console_host=console_host)
+    except Exception as error:
+        logger.warning(
+            f"Failed to upsert {kind} into the catalog (continuing): {error}"
+        )
```
omby8888: i think that at least `try_upsert_entity` is not needed and it will be cleaner to use ocean.register_raw in place, also the usage of async with event_context(EventType.HTTP_REQUEST, trigger_type="machine") is not needed as i added that part in core already so you'll have it

integrations/cursor-cloud-agents/core/catalog.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+async def fetch_usage_by_run_id(
```
omby8888: this should sit in the agents exporter, also the name is missleading by making me think that it's for specific run

integrations/cursor-cloud-agents/core/catalog.py
```diff
@@ -0,0 +1,178 @@
...
+        if not isinstance(run_usage, dict):
+            continue
+        run_id = run_usage.get("id")
+        usage = run_usage.get("usage")
+        if isinstance(run_id, str) and run_id and usage is not None:
+            usage_by_run_id[run_id] = usage
+    return usage_by_run_id
+
+
+async def fetch_run_raw_for_catalog(
```
omby8888: this should be implemented as get_resource on the runs exporter (see similar patterns in datadog/github integrations), my fear is there are now several places incharged of fetching an agent run and constructing it as raw entity

integrations/cursor-cloud-agents/core/webhook_signing.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+async def _get_org_id() -> str:
+    global _org_id_cache
+    if _org_id_cache is None:
+        _org_id_cache = await ocean.port_client.get_org_id()
+    return _org_id_cache
+
+
+def get_webhook_signing_secret() -> str | None:
```
omby8888: this can be in place or become private, as `derive_webhook_secret` also returns None if this not being set

integrations/cursor-cloud-agents/webhook_processors/utils.py
omby8888: those use only once and together, either both should be united or one become private, also the name should explain better what we want to get here (closest run event to the webhook time, but why?) and consider moving that to the run exporter

integrations/cursor-cloud-agents/webhook_processors/cursor_agent_webhook_processor.py
```diff
@@ -0,0 +1,121 @@
...
+        agent_id = str(payload.get("id"))
+        status = payload.get("status")
+
+        webhook_time = parse_webhook_timestamp(payload)
+        cursor_run_id = await resolve_cursor_run_id_for_webhook(agent_id, webhook_time)
+        run = await resolve_tracked_run(agent_id, cursor_run_id)
+        success = status == "FINISHED"
+        summary = payload.get("summary")
+
+        if run is None:
```
omby8888, line 96: if run is none we only do the log and then nothing, would be cleaner to exit after that and avoid the 2 more similar checks afterwards

integrations/cursor-cloud-agents/clients/run_reads.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+
+from typing import Any
+
+from clients.cursor_agents_client import CursorAgentsClient
+from clients.endpoints import v1_agent_runs
+
+
+async def list_first_runs_page(
```
omby8888: should sit in the runs exporter

integrations/cursor-cloud-agents/actions/create_agent_executor.py
omby8888, line 1: maybe split each api version handling to a different file? or event different actions overall? as each will validate the data differently and the functionality differs by little, wdyt?

integrations/cursor-cloud-agents/.port/spec.yaml
talsabagport: Done — removed the org-secret sentence; kept the openssl example.

integrations/cursor-cloud-agents/.port/spec.yaml
talsabagport: Removed cursorPageSize from spec; pagination uses internal DEFAULT_PAGE_SIZE only.

integrations/cursor-cloud-agents/core/exporters/abstract_exporter.py
talsabagport: Added core/options.py with ListAgentOptions / ListRunOptions / GetRunOptions and from_resource_config factories; abstract exporter now takes options with async generator signatures.

integrations/cursor-cloud-agents/core/exporters/runs_exporter.py
talsabagport: RunsExporter now iterates cached AgentsExporter.get_paginated_resources instead of paginating V1_AGENTS directly.

integrations/cursor-cloud-agents/core/catalog.py
talsabagport: Moved to AgentsExporter.get_usage_by_run_id.

integrations/cursor-cloud-agents/core/exporters/agents_exporter.py
talsabagport: Added @cache_iterator_result on AgentsExporter.get_paginated_resources.

integrations/cursor-cloud-agents/core/catalog.py
talsabagport: Implemented as RunsExporter.get_resource(GetRunOptions); removed fetch_run_raw_for_catalog from catalog.

integrations/cursor-cloud-agents/core/exporters/runs_exporter.py
talsabagport: Added RunSelector.enrichRunsWithUsage (default true); RunsExporter skips usage fetch when false.

integrations/cursor-cloud-agents/core/webhook_signing.py
talsabagport: Renamed to _get_webhook_signing_secret (private); authenticate still checks it once at the top.

integrations/cursor-cloud-agents/core/exporters/runs_exporter.py
talsabagport: Added RunSelector.oldestRunDate; pagination stops per agent once createdAt is before the cutoff.

integrations/cursor-cloud-agents/webhook_processors/utils.py
talsabagport: Collapsed into resolve_newest_run_id_at_or_before (uses RunsExporter.list_first_page internally).

talsabagport: Inlined tracked-run lookup in the webhook processor (_find_tracked_run: cursor run id first, then agent id).

integrations/cursor-cloud-agents/webhook_processors/cursor_agent_webhook_processor.py
talsabagport: Restructured handle_event to return early when no tracked run (catalog upsert only).

integrations/cursor-cloud-agents/core/catalog.py
talsabagport: Removed try_upsert_entity/upsert_raw_entity; webhook and executors call ocean.register_raw directly (no nested event_context).

integrations/cursor-cloud-agents/clients/run_reads.py
talsabagport: Moved to RunsExporter.list_first_page; deleted clients/run_reads.py.

integrations/cursor-cloud-agents/actions/create_agent_executor.py
talsabagport: Kept one create_agent action (apiVersion is a v0 bridge). Split implementation into strategy handlers under actions/create_agent/ (v0_handler, v1_handler) with a thin CreateAgentExecutor facade.

integrations/cursor-cloud-agents/actions/create_agent/context.py
```diff
@@ -0,0 +1,44 @@
...
+    model: object | None
+    auto_create_pr: bool | None
+    config: dict[str, object]
+
+    @classmethod
+    def from_run(cls, run: IntegrationRun) -> CreateAgentContext:
+        props = run.execution_properties
+        config = props.get("config")
+        if config is None:
+            config = {}
```
omby8888:
```suggestion
        config = props.get("config", {})
```

integrations/cursor-cloud-agents/actions/config_validation.py
omby8888, line 1: This whole file should be in the context model in some way, for instance `validate_report_completion_policy` should be validator of the pydantic model and `parse_api_version` should be private fun in that file

integrations/cursor-cloud-agents/clients/client_factory.py
```diff
@@ -0,0 +1,22 @@
...
+    global _cursor_agents_client
+    if _cursor_agents_client is not None:
+        return _cursor_agents_client
+
+    integration_config: dict[str, Any] = ocean.integration_config
+    _cursor_agents_client = CursorAgentsClient(
+        api_host=integration_config["cursor_api_host"],
+        api_key=integration_config["cursor_api_key"],
+        console_host=integration_config["cursor_console_host"],
+        page_size=DEFAULT_PAGE_SIZE,
```
omby8888: let's remove this param all together and just set the page size internally in the client

omby8888 (changes requested):

omby8888 (approved):
Nice!!

PR #3588, opened 2026-07-21
[Core] add relative time helpers in core

raskinmaya:
# Description
- Add shared lookback helpers in `port_ocean.utils.relative_time`: `days_ago`, `months_ago`, and `to_rfc3339`
- Covering the common day/month → UTC datetime → RFC3339 cases integrations reimplement today
- Leave vendor-specific behavior local (start-of-day/month snaps, 30d strings, search prefixes, window partitioning, API clamps)

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Low Risk**
> Additive utility module and tests only; no changes to resync, auth, or existing integration paths.
> 
> **Overview**
> Introduces **`port_ocean.utils.relative_time`** for shared UTC lookback math in integration selectors: **`days_ago`** and **`months_ago`** (optional injectable `now`, negative offsets treated as future), plus **`to_rfc3339`** for API-friendly UTC strings (naive datetimes assumed UTC; `timespec` seconds or microseconds).
> 
> Adds **`port_ocean/tests/utils/test_relative_time.py`** with frozen-clock coverage for day/month edges, negative offsets, and RFC3339 formatting. Release notes and **`pyproject.toml`** version bump to **0.45.8**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ac0a2761e9c8bab8f3b4ffac9f512c7dace51aa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/port_ocean/tests/utils/test_relative_time.py
+++ b/port_ocean/tests/utils/test_relative_time.py
@@ -38,12 +38,6 @@ def test_months_ago_calendar(monkeypatch: pytest.MonkeyPatch) -> None:
     assert months_ago(1) == datetime(2025, 12, 31, 12, 0, 0, tzinfo=timezone.utc)
 
 
-def test_months_ago_negative_is_future(freeze_now: datetime) -> None:
-    assert months_ago(-2) == datetime(
-        2026, 9, 21, 12, 30, 45, 123456, tzinfo=timezone.utc
-    )
-
-
 def test_to_rfc3339_seconds() -> None:
     assert to_rfc3339(FROZEN_NOW) == "2026-07-21T12:30:45Z"
 
--- a/port_ocean/tests/utils/test_relative_time.py
+++ b/port_ocean/tests/utils/test_relative_time.py
@@ -1,52 +1,47 @@
 from datetime import datetime, timedelta, timezone
+from types import SimpleNamespace
 
-from dateutil.relativedelta import relativedelta
+import pytest
 
+import port_ocean.utils.relative_time as relative_time
 from port_ocean.utils.relative_time import days_ago, months_ago, to_rfc3339
 
 FROZEN_NOW = datetime(2026, 7, 21, 12, 30, 45, 123456, tzinfo=timezone.utc)
 
 
-def test_days_ago() -> None:
-    assert days_ago(7, now=FROZEN_NOW) == FROZEN_NOW - timedelta(days=7)
-
-
-def test_days_ago_negative_is_future() -> None:
-    assert days_ago(-3, now=FROZEN_NOW) == FROZEN_NOW + timedelta(days=3)
-
-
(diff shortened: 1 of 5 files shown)


port_ocean/utils/relative_time.py
```diff
@@ -0,0 +1,41 @@
...
+    return value.astimezone(timezone.utc)
+
+
+def days_ago(days: int, *, now: datetime | None = None) -> datetime:
+    """UTC datetime ``days`` ago (negative ``days`` = in the future)."""
+    current = _as_utc(now) if now is not None else datetime.now(timezone.utc)
+    return current - timedelta(days=days)
+
+
+def months_ago(months: int, *, now: datetime | None = None) -> datetime:
```
VenfiOranai: I think this is probably just a naming thing, but it's a bit weird to allow `now` to be injected. Like, it should only ever have one value, that being datetime.now (which is the default you set when None is injected), meaning that if a different value is injected, it's not really now.
If the plan here really is to call now outside and inject it, then I'm not sure that's really worth the added complexity just to save the duplicate call (unless this function is supposed to be called a ton in a short span of time)

port_ocean/tests/utils/test_relative_time.py
```diff
@@ -0,0 +1,65 @@
+from datetime import datetime, timedelta, timezone
+
+from dateutil.relativedelta import relativedelta
+
+from port_ocean.utils.relative_time import days_ago, months_ago, to_rfc3339
+
+FROZEN_NOW = datetime(2026, 7, 21, 12, 30, 45, 123456, tzinfo=timezone.utc)
+
+
+def test_days_ago() -> None:
```
VenfiOranai: Some of these tests don't really check anything, they run the exact same logic as the function they're checking. I think it might be preferable to assert on static values rather than calculated ones (the ideal would be to create a static comparison datetime object, but if that doesn't work out then comparing specific attributes should also be great imo)

port_ocean/utils/relative_time.py
raskinmaya: Agreed — dropped the injectable `now`. The helpers always use `datetime.now(timezone.utc)`. Tests freeze the clock via monkeypatch instead.

port_ocean/tests/utils/test_relative_time.py
raskinmaya: Updated — assertions now use static expected datetimes (e.g. `datetime(2026, 7, 14, ...)`) rather than re-deriving the lookback with `timedelta`/`relativedelta`.

VenfiOranai (approved):
Looks great

PR #3604, opened 2026-07-23
[Core] remove non-transient client errors from retryable requests by default

eriport:
# Description

What - 

remove `HTTPStatus.BAD_REQUEST` (400) from the default retry status codes in `RetryConfig`. `HTTPStatus.UNAUTHORIZED` (401) is retained, OAuth integrations rely on it to trigger the token refresh healing mechanism via `_ON_RETRY_CALLBACK`.

Why - 

a 400 is a non-transient client error. The request is malformed and will keep failing. The framework already supports opt-in retries for specific status codes via `additional_retry_status_codes`, which is the right mechanism for integrations with specific needs.

How - 

removes the two status codes from, `default_status_codes` and provided a function to help provide visibility when we hit this non-retryable errors. If your integration relies on the default retry behaviour for 400 or specific statuses, opt in explicitly:

```python
RetryConfig(
    additional_retry_status_codes=[HTTPStatus.BAD_REQUEST]
)
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Breaking behavior for any integration that relied on default retries for 400/401; core HTTP retry path is widely used but the change is intentional and opt-in recovery exists.
> 
> **Overview**
> **Breaking:** Default HTTP retries no longer include **400** or **401**. Those responses fail on the first attempt instead of going through the retry/backoff loop. Transient codes (429, 502, 503, 504) are unchanged.
> 
> Integrations that still need retries on auth or bad-request edge cases can opt in with `RetryConfig(additional_retry_status_codes=[HTTPStatus.UNAUTHORIZED])` (or 400). Port’s own client can keep handling 401 via `TokenRetryTransport` / `before_retry_async` separately from these defaults.
> 
> When a final response is **401** or **400**, `RetryTransport` now logs targeted warnings (credentials vs payload) via `_log_non_retryable_client_error` in both sync and async paths. Release **0.46.0** and changelog document the breaking change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55c26d62152d2a07b3151678563a545896b2b576. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


port_ocean/helpers/retry.py
```diff
@@ -419,6 +417,22 @@ def _log_before_retry(
                 f" {type(error).__name__} - {str(error) or 'No error message'}, retrying in {sleep_time} seconds."
             )
 
+    def _log_non_retryable_client_error(
```
omby8888: why we need to log that specifically? isn't this already logged upon failure?

port_ocean/helpers/retry.py
```diff
@@ -101,8 +101,6 @@ def __init__(
                 HTTPStatus.BAD_GATEWAY,
                 HTTPStatus.SERVICE_UNAVAILABLE,
                 HTTPStatus.GATEWAY_TIMEOUT,
-                HTTPStatus.UNAUTHORIZED,
```
omby8888, line 104: This is a problem as OAuth clients rely on the first failure to know they need to read a refreshed token from  `ocean.app.config.oauth_access_token_file_path` as healing mechanism.

omby8888 (changes requested):

omby8888 (approved):
LGTM

PR #3655, opened 2026-08-03
[Integration][AWS-v3] Add CloudTrail live events webhook processor

Hadar-Vaknin:
# Description

- Add a unified `CloudTrailWebhookProcessor` on `/webhook` that ingests EventBridge CloudTrail events.

- Support S3 buckets (`CreateBucket`, `DeleteBucket`) and Lambda functions (`CreateFunction`, `UpdateFunctionConfiguration`, `UpdateFunctionCode`, `DeleteFunction`) via a CloudTrail event-name mapping registry and per-kind `LiveEventFactories` wired into `ExporterMetadata`.

- Gate live events behind the org feature flag `AWS_V3_LIVE_EVENTS_ENABLED`.

- Improve S3 bucket live-event reliability: `head_bucket` before inspection so missing buckets raise instead of returning stub resources; treat `NoSuchBucket` / `404` as not-found; pass `AccountId` through exporter context.

[Design doc for the feature](https://www.notion.so/live-events-in-aws-v3-integration-3a69be8ff7e381d58be5e62814ff8616?source=copy_link)

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> New externally exposed webhook path and API-key auth affect entity sync correctness; misconfiguration or weak keys could allow spoofed updates, and access-denied fetch paths delete Port entities.
> 
> **Overview**
> Adds **live event ingestion** for AWS-V3 via a `/webhook` endpoint that accepts EventBridge/CloudTrail payloads, secured with `live_events_api_key` and the `x-port-aws-ocean-api-key` header.
> 
> **CloudTrail pipeline:** A parser normalizes envelopes into kind, account, region, identifier, and upsert/delete action. Registries map CloudTrail `eventName` values to **S3 buckets** (`CreateBucket` / `DeleteBucket`) and **Lambda functions** (versioned create/update/delete API names). A single `CloudTrailWebhookProcessor` routes events through `kind_to_export_metadata`: deletes emit ARN/name properties; upserts resolve an account session, fetch the resource via existing exporters (honoring selector `include_actions`), or treat access denied / not found as delete.
> 
> **Exporter metadata:** `ExporterMetadata` moves to `metadata/types` and gains optional `LiveEventFactories` (request + delete property builders). S3 and Lambda register live-event hooks and shared ARN helpers. `get_session_for_account` looks up sessions by account ID (documented as POC-linear scan).
> 
> **S3 tweaks:** Single-bucket fetch passes `AccountId` into `ResourceInspector`; list-buckets only sets `CreationDate` when present.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4fbfef793f39a3a31cd28eba65acd3f752a903e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Hadar-Vaknin:
@copilot review

port_ocean/core/models.py
```diff
@@ -142,6 +142,7 @@ class IntegrationFeatureFlag(StrEnum):
     )
```
Hadar-Vaknin, line 142: Will be separated into a different PR

integrations/aws-v3/aws/core/exporters/live_events/__init__.py
```diff
@@ -0,0 +1,3 @@
+from aws.core.exporters.live_events.arn import regional_service_arn, s3_bucket_arn
+
+__all__ = ["regional_service_arn", "s3_bucket_arn"]
```
VenfiOranai: You can add `# mypy: implicit_reexport` at the top of the file, saves the need for the dupe declaration

integrations/aws-v3/aws/core/exporters/live_events/arn.py
```diff
@@ -0,0 +1,17 @@
+from aws.core.exporters.metadata.types import LiveEventContext
```
VenfiOranai: Not sure how much I dig this file. Both the position (it'll get lost between all the exporters), and the grouping of functions feels like each one should be in its own relevant file.

integrations/aws-v3/aws/core/exporters/s3/bucket/exporter.py
```diff
@@ -23,12 +23,18 @@ async def get_resource(self, options: SingleBucketRequest) -> dict[str, Any]:
         async with AioBaseClientProxy(
             self.session, options.region, self._service_name
         ) as proxy:
+            # ListBucketsAction only shapes the input dict and does not call AWS.
+            # Verify the bucket exists first so missing buckets raise (e.g.
+            # NoSuchBucket) instead of returning a synthesized stub resource.
```
VenfiOranai: I feel like the comment here could be made to be a bit more explicit about the goal, while dropping the method which is a bit over-explained. Like, I assume that it's in order to enable delete LE to ingest correctly, but it's not 100% clear (and also I might've misunderstood)

integrations/aws-v3/aws/core/exporters/exporter_metadata.py
```diff
@@ -72,20 +78,20 @@
...
-    exporter: type[IResourceExporter[Any]]
-    paginated_request_model: type[ResourceRequestModel]
-    regional: bool = True
 
+__all__ = [
+    "ExporterMetadata",
+    "LiveEventContext",
+    "LiveEventFactories",
+    "kind_to_export_metadata",
+]
```
VenfiOranai: Why?

integrations/aws-v3/aws/webhook/event_name_mappings/lambda_function.py
```diff
@@ -0,0 +1,33 @@
...
+from aws.core.helpers.types import ObjectKind
+from aws.webhook.event_name_mappings.mapping import CloudTrailEventAction, EventNameMapping
+
+
+def _extract_lambda_function_name(detail: dict[str, Any]) -> str | None:
+    function_name = detail.get("requestParameters", {}).get("functionName")
+    return function_name if isinstance(function_name, str) and function_name else None
+
+
+MAPPINGS: dict[str, EventNameMapping] = {
```
VenfiOranai: Any reason not to call this `LAMBDA_MAPPINGS`? Might just be a personal thing, but I'd prefer to minimize usage of `as` in imports, can harm readability

integrations/aws-v3/aws/webhook/event_name_mappings/lambda_function.py
```diff
@@ -0,0 +1,33 @@
+from typing import Any
+
+from aws.core.helpers.types import ObjectKind
+from aws.webhook.event_name_mappings.mapping import CloudTrailEventAction, EventNameMapping
+
+
+def _extract_lambda_function_name(detail: dict[str, Any]) -> str | None:
+    function_name = detail.get("requestParameters", {}).get("functionName")
+    return function_name if isinstance(function_name, str) and function_name else None
```
VenfiOranai: Are the checks here necessary? Like, can a function's name be an empty string?
Also in the one place I see this ends up being called (cloudtrail_parser.py, line 95), it seems like we return None on any falsy value, so None and '' should end up functioning the same

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
```diff
@@ -0,0 +1,181 @@
...
+from aws.webhook.feature_flags import is_aws_v3_live_events_enabled
+
+
+class CloudTrailWebhookProcessor(AbstractWebhookProcessor):
+    """Handles CloudTrail live events for all registered resource kinds."""
+
+    async def authenticate(
+        self, payload: EventPayload, headers: dict[str, Any]
+    ) -> bool:
+        if not await is_aws_v3_live_events_enabled():
```
VenfiOranai, line 67: Not 100% familiar with how standard live events work in Ocean, but should we ever get here when live events aren't enabled? Like, if this line runs it means we've received a live event (so either they are enabled, or they're disabled and we're receiving a huuuuuuge amount of spam)

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
```diff
@@ -0,0 +1,181 @@
...
+            logger.warning("AWS-v3 live events are disabled by organization feature flag")
+            return False
+
+        expected_api_key = ocean.integration_config.get("live_events_api_key")
+        if not expected_api_key:
```
VenfiOranai, line 72: Similar to above, is there no global way to block it before we enter this flow assuming the ocean instance isn't geared for LE?

integrations/aws-v3/aws/core/exporters/metadata/types.py
```diff
@@ -0,0 +1,31 @@
...
+class LiveEventContext:
+    identifier: str
+    account_id: str
+    region: str
+
+
+@dataclass(frozen=True)
+class LiveEventFactories:
+    request_factory: Callable[[LiveEventContext, list[str]], ResourceRequestModel]
+    delete_properties_factory: Callable[[LiveEventContext], dict[str, str]]
```
VenfiOranai: Might just be the naming, but I really don't get what this is actually used for. Like, going off the name I'd assume it's properties you want to delete (not sure why you'd want that). Then looking at the logic it seems like properties you send on delete (in which case I'm not 100% sure I dig the method too much, feel like identifier keys for objects should be declared in a less specific function since it makes sense they'd be used for other things as well)

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
```diff
@@ -0,0 +1,181 @@
...
+            return WebhookEventRawResults(
+                updated_raw_results=[], deleted_raw_results=[]
+            )
+
+        live_event_context = LiveEventContext(
+            identifier=parsed.identifier,
+            account_id=parsed.account_id,
+            region=parsed.region,
+        )
+        live_events = metadata.live_events
```
VenfiOranai: Really super personal nitpick, but could you pretty please bump this up one line, having a populated line that's also a one-liner right after a closed and indented section looks a smidge bit off

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
```diff
@@ -0,0 +1,181 @@
...
+        return is_supported_cloudtrail_event(event.payload)
+
+    async def get_matching_kinds(self, event: WebhookEvent) -> list[str]:
+        parsed = parse_cloudtrail_event(event.payload)
+        return [parsed.kind] if parsed is not None else []
+
+    async def validate_payload(self, payload: EventPayload) -> bool:
+        return parse_cloudtrail_event(payload) is not None
+
+    async def handle_event(
```
VenfiOranai: This function feels a bit gargantuan, thoughts on splitting it up a bit? My immediate thought goes to creating an abstract class for handling events by type (and then two instances of the class, one for upsert one for delete), but let me know what you think

integrations/aws-v3/aws/webhook/cloudtrail_parser.py
```diff
@@ -0,0 +1,109 @@
...
+
+    mapping = EVENT_NAME_MAPPINGS.get(event_name)
+    if mapping is None:
+        return None
+
+    identifier = mapping.extract_identifier(detail)
+    account_id = payload.get("account") or detail.get("recipientAccountId")
+    region = payload.get("region") or detail.get("awsRegion")
+
+    if not identifier or not account_id or not region:
```
VenfiOranai, line 87: Sure region is mandatory? Kinds like S3 aren't always region-specific, so does that always get sent? Even if for S3 the answer is yes, worth looking into it for other future kinds as well just to make sure this isn't a potential isse.

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
```diff
@@ -0,0 +1,181 @@
...
+        self, payload: EventPayload, resource_config: ResourceConfig | None
+    ) -> WebhookEventRawResults:
+        parsed = parse_cloudtrail_event(payload)
+        if parsed is None:
```
VenfiOranai: You do this in `validate_payload`, sure it's necessary here as well?

integrations/aws-v3/aws/webhook/cloudtrail_parser.py
```diff
@@ -0,0 +1,109 @@
+"""CloudTrail-via-EventBridge parser for aws-v3 live events.
```
VenfiOranai, line 1: Not sure this is the best place for the example payload, any thoughts as to where might be more relevant? Or honestly do we even need this at all?

integrations/aws-v3/aws/webhook/cloudtrail_parser.py
```diff
@@ -0,0 +1,109 @@
...
+class NormalizedEvent:
+    kind: str
+    identifier: str
+    account_id: str
+    region: str
+    action: CloudTrailEventAction
+    event_name: str
+
+
+def _get_detail(payload: dict[str, Any]) -> dict[str, Any]:
```
VenfiOranai: Some of these functions feel a bit like overkill. Like, assuming detail exists, can it ever not be a dict?
Same thing in `get_event_name` with the str check

integrations/aws-v3/aws/webhook/cloudtrail_parser.py
```diff
@@ -0,0 +1,109 @@
...
+    return event_name if isinstance(event_name, str) else None
+
+
+def _has_error_code(detail: dict[str, Any]) -> bool:
+    """Return True when the CloudTrail event records a failed API call."""
+    error_code = detail.get("errorCode")
+    return isinstance(error_code, str) and bool(error_code)
+
+
+def is_supported_cloudtrail_event(payload: dict[str, Any]) -> bool:
```
VenfiOranai: Kinda relates to the comment at the top of the file, but what do you think about declaring a TypedDict and using that rather than `dict[str, Any]`? I think it might make following the typing easier, and it's a solid way to document the API without having it in a comment.

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
```diff
@@ -0,0 +1,181 @@
...
+        parsed = parse_cloudtrail_event(event.payload)
+        return [parsed.kind] if parsed is not None else []
+
+    async def validate_payload(self, payload: EventPayload) -> bool:
+        return parse_cloudtrail_event(payload) is not None
+
+    async def handle_event(
+        self, payload: EventPayload, resource_config: ResourceConfig | None
+    ) -> WebhookEventRawResults:
+        parsed = parse_cloudtrail_event(payload)
```
VenfiOranai: You call the parse function in three places (with two of them barely using the contents). Would it be possible to have this only be called once in order to avoid the redundancy?

integrations/aws-v3/aws/utils/feature_flags.py
```diff
@@ -0,0 +1,16 @@
+from loguru import logger
```
VenfiOranai, line 1: Thoughts on moving this file to the general utils under aws?
That way we have one dedicated feature flags file for the whole integration rather than spreading them out

integrations/aws-v3/aws/webhook/feature_flags.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+async def is_aws_v3_live_events_enabled() -> bool:
+    """Return whether AWS-v3 CloudTrail live events are enabled for this org."""
+    try:
+        flags = await ocean.port_client.get_organization_feature_flags()
+        return IntegrationFeatureFlag.AWS_V3_LIVE_EVENTS_ENABLED in flags
+    except Exception as error:
+        logger.warning(
+            f"Failed to check AWS-v3 live events feature flag, assuming disabled: {error}"
```
VenfiOranai: I'm not 100% familiar with our logging settings, but do we support having the error not as a straight string? I know in default python it's with `exc_info=True`, and I saw that in loguru it's done with `.opt(exception=True)`, just makes the error within the log much cleaner
If you're not sure yourself then nevermind

integrations/aws-v3/aws/webhook/event_name_mappings/lambda_function.py
```diff
@@ -0,0 +1,33 @@
+from typing import Any
```
VenfiOranai: Why here and not in the exporters dir? As is we'll need a full duplicate of all the kinds which I feel we should rather minimize

integrations/aws-v3/aws/core/exporters/live_events/__init__.py
Hadar-Vaknin: Removed the whole file.

integrations/aws-v3/aws/core/exporters/s3/bucket/exporter.py
Hadar-Vaknin: Close, this is about making sure delete live events are handled correctly, but not on the explicit DeleteBucket path. It matters for create/update live events, when we re-fetch the bucket after the event, get_resource only has the bucket name and would otherwise synthesize a stub resource without checking AWS. head_bucket ensures a missing bucket raises NoSuchBucket, which the webhook processor catches and turns into a delete ingestion instead of upserting a phantom bucket.

I will update the comment to be a bit more explicit

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
Hadar-Vaknin: We can hit this with live events disabled. The route is always registered at startup, and AWS keeps delivering if infra was set up before the FF was turned off. The org FF gates processing, not delivery. 
I tried gating route registration, but add_webhook_processor runs synchronously at import and the AWS side isn’t tied to the FF. 
If u can come up with a better solution, I would love to implement it. 100% agree with u btw - if customers set up LE AWS CloudFormation, this will be spamming if the FF is off.

integrations/aws-v3/aws/webhook/cloudtrail_parser.py
Hadar-Vaknin: Shortened the docstring a bit, I guess the tests are enough to figure out the payload

integrations/aws-v3/aws/core/exporters/metadata/types.py
Hadar-Vaknin: delete_properties_factory returns the Properties dict Port uses to match and delete the entity in deleted_raw_results.

Renamed to `deletion_identifier_properties_factory`and added a short docstring. If you prefer, we could also pull identifier derivation out of the live-events module into a shared helper on the exporter, though right now delete is the only consumer. 
wdyt?

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
Hadar-Vaknin: Agree this should be split by action, I would rather do that with private methods on the processor rather than an ABC tbh...
But give me an ACK on this before I implement.

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
Hadar-Vaknin: This function will be changed because of the comment below :)

integrations/aws-v3/aws/webhook/cloudtrail_parser.py
Hadar-Vaknin: I kept the dict guard on _get_detail since we still want malformed payloads  (detail as a string or something) to be rejected gracefully rather than crash

integrations/aws-v3/aws/webhook/cloudtrail_parser.py
Hadar-Vaknin: Yes, AWS always sends region (payload.region from EventBridge + detail.awsRegion from CloudTrail). S3 is global in the catalog but still regional at the API layer so we need region for the upsert fetch client

integrations/aws-v3/aws/webhook/feature_flags.py
Hadar-Vaknin: Changed it to the common pattern

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
VenfiOranai: I think that's probably solid as well (just prefer having a dict mapping the type to function rather than an if/elif/else chain)

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
VenfiOranai: Mmm, my only real idea involves changing the `add_webhook_processor` function, which is probably definitely out of scope, so nvm

integrations/aws-v3/aws/core/exporters/metadata/types.py
VenfiOranai: I think it might be preferable to pull it out to a dedicated section. There's at least one other thing I can think of that might be useful for (action merging in the exporter for example), so having it ready as a standalone bit will help there later on.

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
Hadar-Vaknin: Sounds good

integrations/aws-v3/aws/core/exporters/live_events/__init__.py
VenfiOranai: Not sure that's best practice, we generally do have init files

integrations/aws-v3/aws/core/exporters/s3/bucket/exporter.py
VenfiOranai: Ok then maybe now I'm a bit more confused. When will we receive a create/update for a bucket that's gone? Wouldn't that be a delete event, or is their api a bit silly?

integrations/aws-v3/aws/utils/feature_flags.py
```diff
@@ -0,0 +1,17 @@
...
+
+from port_ocean.context.ocean import ocean
+from port_ocean.core.models import IntegrationFeatureFlag
+
+
+async def is_aws_v3_live_events_enabled() -> bool:
```
VenfiOranai, line 7: Thoughts on making this a general `is_flag_enabled` with the flag as an input?

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+class CloudTrailWebhookProcessor(AbstractWebhookProcessor):
+    """Handles CloudTrail live events for all registered resource kinds."""
+
+    _cached_parsed_event: NormalizedEvent | None = None
+    _parsed_event_loaded: bool = False
+
+    def _get_parsed_event(self, payload: EventPayload) -> NormalizedEvent | None:
+        if not self._parsed_event_loaded:
```
VenfiOranai, line 57: Would it be possible to `if self._cached_parsed_event`? Saves the need for another param

integrations/aws-v3/aws/core/exporters/metadata/types.py
```diff
@@ -0,0 +1,77 @@
...
+
+@dataclass
+class ExporterMetadata:
+    exporter: type[IResourceExporter[Any]]
+    paginated_request_model: type[ResourceRequestModel]
+    regional: bool = True
+    live_events: LiveEventFactories | None = None
+
+    @property
+    def supports_live_events(self) -> bool:
```
VenfiOranai: Unused?

integrations/aws-v3/aws/webhook/webhook_processors/cloudtrail_webhook_processor.py
Hadar-Vaknin: So _cached_parsed_event can be in two None states: one when we haven't called the parser yet and the second when it parsed with error or invalid 
so after a failed parse you'd re-parse on every call to _get_parsed_event

VenfiOranai (approved):
Looking swell

PR #3686, opened 2026-08-07
[Integration][Azure-DevOps] Prevent duplicate webhook subscriptions from transient lookup failures

jasonnwakaeze-debug:
# Description

What -  Webhook reconciliation was creating unbounded duplicate service hook subscriptions in ADO

Why -

How -
  Removed return_exceptions=True from get_filtered_webhook_subscriptions — any lookup failure now raises through asyncio.gather, aborting webhook setup entirely. No partial data, no phantom creates.          
  Added _find_duplicate_subscriptions (static, pure) — identifies duplicates by (publisherId, eventType, url, projectId) key, returns kept + duplicates.                                                          
  Added _delete_subscriptions / _bounded_delete_subscription — bounded concurrent deletion with per-subscription failure logging.                       
  create_webhook_subscriptions feeds duplicates into subs_to_delete before reconciliation. Guard skips dedup when subscription count is ≤ the 23 defined types (healthy orgs skip the loop entirely).             
     
```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Changes live-event webhook provisioning and can delete duplicate subscriptions in ADO; behavior is safer but webhook setup may fail entirely until ADO API errors clear.
> 
> **Overview**
> Fixes **unbounded duplicate Azure DevOps service hook subscriptions** caused when subscription inventory lookups failed partially (e.g. 429/timeouts) but reconciliation still ran as if missing hooks needed to be created.
> 
> **Lookup failures now fail closed:** `get_filtered_webhook_subscriptions` no longer swallows per–event-type errors via `return_exceptions=True`; any failed fetch aborts webhook setup instead of reconciling from incomplete data.
> 
> **Startup self-heal for existing duplicates:** When more subscriptions exist than the defined event types, reconciliation deduplicates by `(publisherId, eventType, url, projectId)`, queues extras for deletion, and deletes via new bounded-concurrent `_delete_subscriptions` (replacing unbounded parallel deletes).
> 
> Release **0.11.5** with changelog and tests updated for abort-on-failure and duplicate detection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bb874b2181cfbdea8f0ed55d35590a63b800e0a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/azure-devops/azure_devops/client/azure_devops_client.py
```diff
@@ -2189,18 +2189,49 @@ async def fetch(
...
-        subscriptions: list[WebhookSubscription] = []
-        for result in results:
-            if isinstance(result, asyncio.CancelledError):
-                raise result
-            if isinstance(result, BaseException):
-                logger.warning(f"Failed to fetch webhook subscriptions: {result}")
-                continue
-            subscriptions.extend(result)
-        return subscriptions
+        return [sub for batch in results for sub in batch]
```
danamsalem-ux, line 2194: The problem seems to be related to the changes in this pr - https://github.com/port-labs/ocean/pull/3366/changes.
The problem there is that you skip failed subs generated (due to rate limits etc) and then it appears like they are missing so the integration tries to create them.
Adding another method to dedup them won't solve the issue, it keeps the bad logic either way.
I think you should start be removing the bad logic that does this - i believe its the change here- https://github.com/port-labs/ocean/pull/3366/changes#diff-6d76e1ab85e24b8f643174f9a502ef2dd04e0dc019a5ebd6b1c8e7367624da4fR1571 
but you should validate it in staging or locally against the azure devops sandbox account.

danamsalem-ux: Besides that, due to the faulty logic now we have accounts with a lot of duplicated subscription, we need to find them and delete those duplicated subs - it can be via a script or some method here that does it on the pod startup

jasonnwakaeze-debug: Thanks Dan — that's exactly the change. The return_exceptions=True on the subscription listing gather (our PR #3366) is removed in this PR. Failed lookups now abort webhook setup entirely instead of silently returning partial data.

For existing duplicates: _find_duplicate_subscriptions detects and excludes them from reconciliation so they don't interfere with matching, but doesn't delete them on startup, if we're going for deleting on startup, then deleting thousands of subscriptions has a chance of draining the org's TSTU budget before the sync even starts.

WDYT?

danamsalem-ux: then maybe run a script that does this, we need to clean this up, i think  _find_duplicate_subscriptions won't solve the issue it will just ignore it, but we need to own it since we caused it for our customers.
So lets start with reverting what caused it, and think of a script or any other way to remove the duplicates carefully

jasonnwakaeze-debug: I agree, I've removed _find_duplicate_subscriptions entirely. The PR now only reverts the bad logic (removes return_exceptions=True from the subscription listing gather)

For cleanup of existing duplicates I have the scripts for it. How do we send it out, do I create a new repo is there a process for this please?

danamsalem-ux: No no need to its a one time thing, send it to me privately ill review it

danamsalem-ux (approved):

PR #3711, opened 2026-08-10
[Integration][Komodor] Follow the pagination cursor when syncing services

### LINKED ISSUE #3710: [Bug] [Integration] [Komodor] Service resync only ingests the first page
### Ocean Framework Version

`port_ocean = "^0.48.1"`, as pinned in `integrations/komodor/pyproject.toml`.

### Integration Version

`0.2.87`

### Steps to reproduce

1. Install the Komodor integration against an account that has more than 25 services.
2. Run a resync of the `komodorService` kind.
3. Count the `komodorService` entities that land in Port.

### What did you expect to see?

Every service in the Komodor account, across all clusters.

### What did you see instead?

Exactly 25 of them, which is one page at `SERVICES_PAGE_SIZE`. The resync finishes cleanly, with no error and nothing suspicious in the logs.

The cause is in `integrations/komodor/client.py`. `get_all_services` still paginates by offset:

```python
"pagination": {"pageSize": SERVICES_PAGE_SIZE, "page": current_page},
...
current_page = response.get("meta", {}).get("nextPage", None)
if not current_page:
    break
```

Komodor's `POST /api/v2/services/search` moved to cursor pagination some time ago. The request now takes `pagination.token`, and the response returns `meta.token` instead of `meta.page` / `meta.nextPage` / `meta.prevPage`. The API ignores the unknown `page` field rather than rejecting it, so there is no 400 to notice. Without a token the server keeps returning the first keyset page, `meta.nextPage` is never present, and the loop breaks on its first iteration.

`get_health_monitor` is unaffected. `/health/risks` still uses `offset` and `hasMoreResults`.

`integrations/komodor/examples/service.response.json` also still shows the old meta shape, and both existing tests in `tests/test_client.py` assert the offset contract, so they pass against the broken client.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

shaharduany:
# Description

Fixes #3710

What - `get_all_services` now paginates with Komodor's cursor token instead of an offset page number.

Why - `POST /api/v2/services/search` takes `pagination.token` and returns `meta.token`, where it used to take `pagination.page` and return `meta.nextPage`. The client kept sending `page`, which the API ignores rather than rejects, so every request returned the first keyset page with no `nextPage`, and `if not current_page: break` fired on the first iteration. Resync ended after 25 services with no error.

How - The first request sends `{"pageSize": 25}`. Each response's `meta.token` becomes the next request's `pagination.token`, and a null token ends the loop.

`kind` is dropped from the request. It was pinned client side to the same four kinds the API applies when the field is absent, so the result set is unchanged. An empty `kind` array is a 400 rather than a wildcard, so omitting the key is the only way to get the default.

`get_health_monitor` is untouched, `/health/risks` still uses `offset` and `hasMoreResults`.

Also here: `examples/service.response.json` had the old `meta` shape, and `port_ocean` moves to `^0.48.2` because the file validator requires the pin to be at least the latest PyPI release.

No `meta.nextPage` fallback, so a Komodor deployment old enough to still answer with `nextPage` would stop after one page. The integration points at the hosted API, so I would rather have one code path than a branch for a server version I cannot test. Happy to add it if you would rather be defensive.

`integrations/komodor/.ocean-release/fix-services-pagination.yaml`:

```yaml
bump: patch
changelog-type: bugfix
changelog: Paginate services with the cursor token returned by Komodor so resync fetches every page
```

Release intent file rather than a manual bump, since the integration is on `0.2.87` and gets bumped most days. This looks like the first `.ocean-release` file in the repo, so say the word if you would rather have the manual edit. `validate_ocean_release.py` passes locally on the declarative path.

## Type of change

## Integration testing checklist

Test first. Five tests, three of which were red before the change:

- `test_get_all_services_omits_cursor_on_first_request`
- `test_get_all_services_follows_token_across_pages`
- `test_get_all_services_yields_services_from_every_page`
- `test_get_all_services_stops_when_token_is_null`
- `test_get_all_services_stops_when_token_key_is_absent`

The two previous tests asserted the offset contract, so they passed against the broken client. Replaced, not kept. `make test` and `make lint` pass.

Ran against a live account with 1864 services:

- 1864 over 75 requests at `pageSize=25`, versus 25 on the old client.
- A single `pageSize=5000` request returned the same 1864. Diffing both sets by `kind-cluster-namespace-service` found nothing missing or duplicated, so the cursor neither skips nor repeats across 75 hops.
- Omitting `kind` returned the same 1864 as sending the explicit four.

That run also showed the API sends `"token": null` on the last page rather than dropping the key. `if not next_token` covers both, but the fixtures were modelling a response that never occurs, so they now send the null and a separate test covers the absent key.

I have no Port testing org, so none of this went through a real resync and I have no screenshots. Happy to work with someone who has an org set up if you need that first.

- [ ] Docs PR link [here](#)

No blueprint, mapping, or resource kind changed, so the only `examples` edit is the `meta` block.

## Preflight checklist

- [x] Handled pagination
- [ ] Handled rate limiting - unchanged by this PR
- [ ] Support Multi account - unchanged by this PR

## Screenshots

None, see the note about the testing org.

## API Documentation

`POST /api/v2/services/search` on `https://api.komodor.com` or `https://api.eu.komodor.com`.

Request

```json
{
  "pagination": { "pageSize": 25, "token": "<meta.token from the previous response>" }
}
```

Response meta

```json
{ "pageSize": 25, "token": "<cursor for the next page, null on the last page>" }
```

`kind` is optional and defaults to `Deployment`, `StatefulSet`, `DaemonSet` and `Rollout`. I could not find a stable public URL for the reference docs, so the shapes are written out rather than linked.


matan84 (approved):
Hi! Thank you for the contribution. The version bump looks good, and the code changes make sense. Please pull from main, so we'll be able to merge the changes.

shaharduany:
> Hi! Thank you for the contribution. The version bump looks good, and the code changes make sense. Please pull from main, so we'll be able to merge the changes.

Hi @matan84 
Thanks a lot. 
I rebased main, and it'd be great if you would merge. 😄

PR #3720, opened 2026-08-11
[Core] Treat HTTP 408 Request Timeout as a retryable status code 

yaelibarg:
# Description

What - Treat HTTP 408 Request Timeout as a retryable status code 

Why - DSP could send back 408, and we want ocean to retry in that case

How - Add 408 to the `retry_config`

## Release (integrations / ocean core)

If this PR changes an integration or `port_ocean/`, add a release intent file **or** manually bump `pyproject.toml` and `CHANGELOG.md` (manual takes priority):

- Integration: `integrations/<name>/.ocean-release/<unique-name>.yaml`
- Core: `.ocean-release/core/<unique-name>.yaml`

```yaml
bump: patch
changelog-type: bugfix
changelog: Describe the change
```

## Type of change

Please leave one option from the following and delete the rest:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] New Integration (non-breaking change which adds a new integration)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Non-breaking change (fix of existing functionality that will not change current behavior)
- [ ] Documentation (added/updated documentation)

<h4> All tests should be run against the port production environment(using a testing org). </h4>

### Core testing checklist

- [ ] Integration able to create all default resources from scratch
- [ ] Resync finishes successfully
- [ ] Resync able to create entities
- [ ] Resync able to update entities
- [ ] Resync able to detect and delete entities
- [ ] Scheduled resync able to abort existing resync and start a new one
- [ ] Tested with at least 2 integrations from scratch
- [ ] Tested with Kafka and Polling event listeners
- [ ] Tested deletion of entities that don't pass the selector

---

> [!NOTE]
> **Low Risk**
> Small change to default retry policy on the shared HTTP transport; it only adds retries for 408 and does not alter auth, data handling, or breaking APIs.
> 
> **Overview**
> **Ocean core 0.48.4** extends default HTTP retry behavior so **408 Request Timeout** is retried like other transient status codes (429, 502, etc.), including when DSP returns 408.
> 
> `RetryConfig` now includes `HTTPStatus.REQUEST_TIMEOUT` in its built-in `retry_status_codes`, and `RetryTransport`’s documented default list is updated to `[408, 429, 401, 502, 503, 504]`. Unit tests that assert the default retry set were aligned, and **CHANGELOG** / **pyproject.toml** were bumped for the patch release.
> 
[...]

PR #3724, opened 2026-08-11
[Core] Fix failed Entity Sorting with Search Identifiers

emekanwaoma:
# Description

What - Fixed failed-entity topological sorting when entity identifiers or relation targets are search/combinator objects.

Why - Search identifiers are valid dicts, but the failed-upsert retry sorter used identifiers as graph keys, causing `unhashable type: 'dict'` during reconciliation.

How - Normalize identifiers to stable string keys only inside the sorter, without mutating the entity payload sent to Port. Added regression tests for search identifiers and search-identifier dependencies.

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [x]  Resync able to detect and delete entities
- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Low Risk**
> Localized bugfix to failed-upsert topological sorting with added regression tests; no changes to Port API payloads or auth.
> 
> **Overview**
> Fixes failed-entity retry ordering when identifiers or relation targets are **search/combinator dicts**, which previously crashed reconciliation with `unhashable type: 'dict'` because graph nodes used raw identifiers.
> 
> **`EntityTopologicalSorter`** now normalizes identifiers to stable string keys (`json.dumps` with sorted keys for dicts, `str` otherwise) only for sorting—entity payloads sent to Port are unchanged. Relation matching and graph node identity use the same normalization, and dependency edges compare full `node()` tuples instead of loose blueprint/identifier checks.
> 
> **`order_by_entities_dependencies`** in the Port applier delegates to the shared sorter instead of duplicating logic. Release **0.48.4** with changelog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97d958f42a552d0d417f4a82ef843f44e0b6c26a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mk-armah (approved):
LGTM

CHANGELOG.md
raskinmaya, line 1: Please do not use manual bumping, use the new intents way. You can see an example for how it's done in core [here](https://github.com/port-labs/ocean/pull/3720)

pyproject.toml
raskinmaya, line 1: Please do not use manual bumping, use the new intents way. You can see an example for how it's done in core [here](https://github.com/port-labs/ocean/pull/3720)

port_ocean/core/utils/entity_topological_sorter.py
```diff
@@ -43,9 +44,18 @@ def get_entities(self, sorted: bool = True) -> Generator[Entity, Any, None]:
         for entity in sorted_and_mapped:
             yield entity
 
+    @staticmethod
+    def _identifier_key(identifier: Any) -> str:
```
raskinmaya: This duplicates `_get_entity_key` in utils. `_get_entity_key` also handles `BaseModel` identifiers via `.dict()`. Please reuse or extract a shared `normalize_identifier()` helper instead of a third copy. We fixed unhashable dicts in delete/diff paths before; this is the same class of bug

port_ocean/core/handlers/entities_state_applier/port/order_by_entities_dependencies.py
```diff
@@ -1,49 +1,12 @@
...
-Node = tuple[str, str]
+Node = tuple[object, str]
 
 
 def node(entity: Entity) -> Node:
-    return entity.identifier, entity.blueprint
+    return EntityTopologicalSorter.node(entity)
 
 
 def order_by_entities_dependencies(entities: list[Entity]) -> list[Entity]:
```
raskinmaya: Nothing imports this. `applier` calls `EntityTopologicalSorter` directly. Instead of a 12-line pass-through, delete this file. Keeping it just preserves duplicate `Node`/`node()` surface area

port_ocean/core/utils/entity_topological_sorter.py
```diff
@@ -43,9 +44,18 @@ def get_entities(self, sorted: bool = True) -> Generator[Entity, Any, None]:
...
+        if isinstance(identifier, dict):
```
raskinmaya: Only `dict` is normalized; `IngestSearchQuery`/Pydantic models fall through to `str()`, which won’t match `json.dumps` of the dict form on the relation side. `_get_entity_key` already handles `BaseModel`. As I already mentioned, this is probably worth aligning

port_ocean/tests/core/utils/test_entity_topological_sorter.py
```diff
@@ -76,6 +78,44 @@ def test_handle_failed_with_self_dependencies() -> None:
...
+        {"dep_name_1": "missing_entity"},
+    )
+
+    entity_topological_sort = EntityTopologicalSorter()
+    entity_topological_sort.register_entity(entity_a)
+
+    assert list(entity_topological_sort.get_entities()) == [entity_a]
+
+
+def test_handle_failed_with_search_identifier_dependencies() -> None:
```
raskinmaya, line 103: This test matches when relation value equals the target’s search-identifier dict. In production, a failed entity often has a string identifier while the dependent’s relation is a search query dict. Those still won’t match after normalization, so retry order may be wrong - only the crash is fixed

port_ocean/tests/core/utils/test_entity_topological_sorter.py
```diff
@@ -76,6 +78,44 @@ def test_handle_failed_with_self_dependencies() -> None:
     ], f"Processed order: {processed_order}"
 
 
+def test_handle_failed_with_search_identifier() -> None:
```
raskinmaya, line 89: This test only registers one entity - it mainly proves "no crash". Add a case where relations contain a search dict (the original unhashable type: 'dict' path when flattening relation targets) and multiple entities need ordering

port_ocean/core/utils/entity_topological_sorter.py
```diff
@@ -56,24 +66,27 @@ def order_by_entities_dependencies(entities: list[Entity]) -> list[Entity]:
...
-                    for identifiers in entity.relations.values()
-                    if identifiers is not None
-                ],
-                [],
-            )
+            relation_target_ids = {
+                EntityTopologicalSorter._identifier_key(identifier)
+                for identifier in sum(
+                    [
+                        identifiers if isinstance(identifiers, list) else [identifiers]
```
raskinmaya: Relations can be lists (multi-relation). Add a test with `["entity_a", {search dict}]` to lock in flattening + normalization for integrations that mix plain IDs and search relations

port_ocean/core/utils/entity_topological_sorter.py
```diff
@@ -43,9 +47,33 @@ def get_entities(self, sorted: bool = True) -> Generator[Entity, Any, None]:
...
+            return target_ids
+
+        for rule in search_query.get("rules", []):
+            rule_dict = identifier_to_dict(rule)
+            if rule_dict is None:
+                continue
+            if rule_dict.get("rules"):
+                target_ids.update(EntityTopologicalSorter._relation_target_ids(rule))
+            if (
+                rule_dict.get("property") == "$identifier"
```
raskinmaya: `_relation_target_ids` only extracts `$identifier` + `=`. Integrations like github-copilot use `contains` on custom properties (`member_logins`). Those failed entities still won’t get dependency edges. This should be at least documented as a known limitation

port_ocean/tests/core/utils/test_entity_topological_sorter.py
raskinmaya: I think this test can be either dropped or folded into a multi-entity ordering test so we're not maintaining noise

port_ocean/core/utils/entity_topological_sorter.py
```diff
@@ -43,9 +47,33 @@ def get_entities(self, sorted: bool = True) -> Generator[Entity, Any, None]:
         for entity in sorted_and_mapped:
             yield entity
 
+    @staticmethod
+    def _relation_target_ids(identifier: Any) -> set[str]:
```
raskinmaya: This function is identifier normalization logic, not graph sorting. This can be moved to `entity_identifier.py` so `EntityTopologicalSorter` stays focused on topo sort only

port_ocean/core/utils/entity_topological_sorter.py
```diff
@@ -56,24 +84,26 @@ def order_by_entities_dependencies(entities: list[Entity]) -> list[Entity]:
...
+                ):
+                    relation_target_ids.update(
+                        EntityTopologicalSorter._relation_target_ids(identifier)
+                    )
+
             related_entities = [
                 related
                 for related in entities
-                if related.identifier in relation_target_ids
+                if normalize_identifier(related.identifier) in relation_target_ids
```
raskinmaya, line 80: Correct me if I'm wrong, but since matching is identifier-only, two failed entities with the same normalized id on different blueprints could get a bogus edge. Pre-existing, but worth noting since it will be amplified after this PR is merged

emekanwaoma: Noted this in the helper docstring. I kept the existing identifier-only matching behavior and scoped this PR to fixing the dict/hashability crash.

port_ocean/core/utils/entity_topological_sorter.py
emekanwaoma: documented in docstrings

raskinmaya (approved):

PR #3735, opened 2026-08-12
[Integration][Github] Add git blob SHA to skill kind raw payload

travisgosselin:
## Description

Adds the git blob SHA (content hash) of each `SKILL.md` file to the raw payload emitted by the GitHub integration's `skill` kind.

**Motivation:** downstream consumers building catalogs/blueprints on top of the `skill` kind need a cheap way to detect duplicate/copied `SKILL.md` files across repos, and to detect "did this file change" without diffing full text. A git blob SHA is a natural fit for this, and GitHub's Contents/Git Trees API already compute and return it for every file, so this is essentially free plumbing, not a new hash computation.

**Important:** this is a content-addressed blob SHA, not a commit SHA. Two files with identical content (even across different repos/branches/commits) will have the same `blob_sha`.

## What changed

- `github/core/exporters/skill_exporter/utils.py` — added `blob_sha: Optional[str] = None` to the `Skill` pydantic model; threaded a `blob_sha` kwarg through `_build_skill()` / `build_skill_raw_item()`.
- `github/core/exporters/skill_exporter/core.py` — resync path now passes `blob_sha=file_obj["metadata"]["sha"]` into `build_skill_raw_item()`.
- `github/core/exporters/file_exporter/utils.py` / `core.py` — the REST Contents API response already includes `sha`, so no change was needed there. For the GraphQL path, the tree's blob `sha` (already fetched via the recursive Git Trees API) is now threaded through file metadata, rather than adding a new field to the shared GraphQL batch query.
- `github/webhook/webhook_processors/skill_webhook_processor.py` — create/update events pass `blob_sha` from the Contents API response; delete events now also pass `blob_sha`, sourced from the `sha` GitHub's compare API already includes on removed file entries.
- `.port/examples/skill/{a,b,c}.json` — updated fixtures with realistic `blob_sha` values.
- Tests added/updated in `test_skill_plugin_utils.py`, `test_skill_plugin_exporters.py`, `test_file_exporter.py`, and `test_skill_webhook_processor.py` covering resync (REST + GraphQL), create/update/delete webhooks, and the tree-sha threading.
- Added a release intent file (`.ocean-release/skill-kind-blob-sha.yaml`) per the repo's contributing guidelines, so CI can handle the version bump and CHANGELOG entry on merge rather than doing it manually in this PR.

## Field naming

The exported field on the `skill` kind is named `blob_sha` (renamed from an earlier `sha` per review feedback, to avoid ambiguity with commit SHAs and to be explicit that this is specifically a git blob SHA). Internally, the generic per-file metadata carried by the shared file exporter (used across several kinds) still uses the plain key `sha`, matching the existing convention for other kinds (see `.port/examples/folder/a.json`, `.port/examples/branch/*.json`) — only the skill-specific model field was renamed.

## Deviations

- No hashing algorithm is implemented in Python — the SHA is only ever read from GitHub's existing API responses (Contents API `sha`, Git Trees API `sha`, compare API `sha`).
- The shared GraphQL batch file query is left unchanged; the blob SHA is sourced from the already-fetched git tree instead, to avoid widening a query used by other kinds.
- GitLab's analogous skill helper does not currently expose an equivalent blob-sha field. I left it untouched to keep this PR scoped to GitHub — happy to follow up with a GitLab equivalent in a separate PR if desired.

## Test plan

- `poetry run pytest tests/github/core/exporters/ -k "skill or file_exporter"` — 82 passed
- `poetry run pytest tests/` (full github integration suite) — 954 passed
- `poetry run mypy` on all changed source files — no issues
- `poetry run ruff check` on all changed files — all checks passed
- No live resync was run against a real GitHub org/token — validated with existing fixtures/mocks only.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>


integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
```diff
@@ -98,6 +98,7 @@ async def handle_event(
                     branch=current_branch,
                     organization=organization,
                     path_globs=path_globs,
+                    sha=file_data.get("sha"),
```
raskinmaya: No test covers the webhook update path. We added resync + utils tests, but `SkillWebhookProcessor` is the other half of live events. Please add a test (mirror `test_file_webhook_processor`) asserting `get_resource`’s top-level sha lands on `skill.sha` for updates, and that deletes still emit `sha: null`

integrations/github/github/core/exporters/file_exporter/utils.py
```diff
@@ -250,6 +257,7 @@ def build_batch_file_query(
             ... on Blob {{
                 text
                 byteSize
+                oid
```
raskinmaya: We already fetch the recursive tree and each blob entry includes `sha` (`filter_github_tree_entries_by_pattern` drops it). Threading `entry["sha"]` through `file_info` -> `get_graphql_file_metadata` would avoid extending the batch GraphQL query for every small file across all kinds. `oid` matches fetched content, but for branch-tip resync they should be identical. Was tree `sha` considered?

integrations/github/github/core/exporters/file_exporter/utils.py
```diff
@@ -236,6 +242,7 @@ def get_graphql_file_metadata(
         "url": url,
         "path": file_path,
         "size": size,
+        "sha": sha,
```
raskinmaya, line 246: This changes the shared GraphQL batch query for all file fetching, not just skills. That’s probably fine (as metadata gets sha for free), but the PR frames this as skill-kind work. I think it's worth calling out in the description and intent file, something like "file-kind GraphQL metadata now includes sha too". Any consumer assuming a fixed metadata shape should know

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
raskinmaya, line 116: Motivation is duplicate/change detection, but delete stubs always have `sha: null`. GitHub’s compare API should include `sha` on removed files. We could pass `file_info.get("sha")` with no extra fetch. If identity on delete matters for catalog cleanup (“which skill content left?”), this is a cheap win and I don't see a reason to not wire it there as well

integrations/github/tests/github/core/exporters/test_file_exporter.py
```diff
@@ -1000,6 +1000,76 @@ async def mock_batches(
...
+                        "branch": "main",
+                    }
+                ]
+            ):
+                pass
+
+        process_file_mock.assert_awaited_once()
+        metadata = process_file_mock.await_args.kwargs["metadata"]
+        assert metadata["sha"] == "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391"
+
```
raskinmaya: These are ~70 lines mocking `process_graphql_files` end-to-end for one metadata field. A direct unit test on `_process_retrieved_graphql_files` (or `get_graphql_file_metadata(..., sha=oid)`) would be shorter and less brittle. Fine to keep if this matches nearby patterns, but it’s a lot of setup for assert `metadata["sha"] == oid` IMO

integrations/github/github/core/exporters/file_exporter/core.py
raskinmaya, line 348: Not introduced here, but relevant: GraphQL path skips files with falsy content, so an empty `SKILL.md` never reaches the skill exporter on resync. Webhook path doesn’t have this guard. If empty skills are valid, this is a resync vs webhook parity bug that's worth handling

raskinmaya (changes requested):

omby8888:
can we rename the attribute to `blob_sha` instead of `sha`?

travisgosselin:
Done in 1ff26191f — renamed `sha` to `blob_sha` on the `skill` kind (model field, builder kwargs, fixtures, tests, PR description). The shared file-exporter metadata key used across other kinds is untouched.

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
travisgosselin: Added in 103eb88aa: `tests/github/webhook/webhook_processors/test_skill_webhook_processor.py`, mirroring `test_file_webhook_processor.py`. Covers both the update path (Contents API `sha` -> `skill.blob_sha`) and confirms delete stubs still emit `blob_sha: null`.

integrations/github/github/core/exporters/file_exporter/utils.py
travisgosselin: Good catch... I like that approach better for sure... switched in 29fea8f9d. `filter_github_tree_entries_by_pattern` now carries `sha` from the tree entry through `file_info` -> `extract_file_paths_and_metadata` -> `_process_retrieved_graphql_files`, and `build_batch_file_query` no longer requests `oid`. Tree sha is the branch-tip sha at discovery time (same commit the GraphQL Blob is read from), so it's equivalent here and avoids touching the shared batch query.

integrations/github/github/core/exporters/file_exporter/utils.py
travisgosselin: Resolved by 29fea8f9d — the shared GraphQL batch query is no longer touched at all (build_batch_file_query never gained the oid field, since we switched to threading the tree's own sha through instead — see the reply on the sibling comment).

integrations/github/github/webhook/webhook_processors/skill_webhook_processor.py
travisgosselin: No I had not thought it was important on delete, but as you say, that is interesting easy add.

travisgosselin: Wired up in c7389072c — deleted_raw_results now passes blob_sha=file_info.get("sha") from the compare API's removed-file entry, no extra fetch needed. Updated the delete-path test to assert the sha flows through instead of asserting it's null.

integrations/github/tests/github/core/exporters/test_file_exporter.py
travisgosselin: It was a similar approach in the test above it... reviewing a bit more....

travisgosselin: Agreed — simplified in 6079e4ab8 to a direct unit test on `_process_retrieved_graphql_files` instead of mocking `process_graphql_files` end-to-end. ~70 lines down to ~25, and it no longer depends on the batching/query-building machinery it wasn't actually testing.

integrations/github/github/core/exporters/file_exporter/core.py
travisgosselin: That seems fair - but as an existing bug, probably worth putting into another PR for transparency / discovery. I'll let you coordinate on the existing bug on your side unless it blocks this.

travisgosselin:
Thanks for the detailed review @raskinmaya ... I think I got the items covered. Are you able to review again?

integrations/github/github/core/exporters/skill_exporter/utils.py
```diff
@@ -139,17 +142,20 @@ def build_skill_raw_item(
     branch: str,
     path_globs: list[str],
     organization: Optional[str] = None,
+    blob_sha: Optional[str] = None,
 ) -> dict[str, Any]:
     """Single entry point for building a skill raw item.
 
-    Webhook deletes call this with an empty ``content`` so that deleted entities
-    carry the exact same identifiers as the upserted ones.
+    Webhook deletes call this with an empty ``content`` (and no ``blob_sha``) so
```
emekanwaoma: tiny: this docstring still says delete calls pass no blob_sha, but delete webhooks now pass blob_sha=file_info.get("sha").

emekanwaoma (changes requested):
LGTM. Left tiny suggestion.

integrations/github/github/core/exporters/file_exporter/core.py
raskinmaya: Fair. I've opened a bug for it

raskinmaya (approved):

integrations/github/github/core/exporters/skill_exporter/utils.py
travisgosselin: Great - I see this was fixed. Thanks!

emekanwaoma (approved):
LGTM

PR #3746, opened 2026-08-13
[CI] Add workflow to notify integration support about external contributions

MPTG94:
# Description

What - Adds a GitHub Actions workflow that notifies n8n when an external contributor opens or updates a PR, or when anyone comments on that PR.

Internal PRs (opened by org members, owners, or collaborators) are ignored entirely — including comments on them.

Why - so integration support know an external contribution was made and they can take a look

How - GH workflow that triggers on the correct events and notifies through an n8n workflow, the workflow references secrets that were already added to the repository

## Type of change


.github/workflows/notify-on-external-contribution.yml
```diff
@@ -0,0 +1,130 @@
+name: Notify on external contribution
```
VenfiOranai, line 1: I'm not familiar with n8n so it's possible they have something on their end to handle this, but won't it potentially be a bit spammy? Like, on every push/comment/reply (something which can happen in bursts depending on the dev) seems a bit heavy without some sort of debouncing.
Also how do bot comments figure into this? Amplify/bot review comments will generally come in immediately after the PR open, so that might also be a bit of a burst.

MPTG94: This is just the initial version so it will probably go through some fine tuning to make it less spammy (either here or in n8n).

Yes n8n has its own tooling to control flow or decide what gets executed or not, at the moment it's more about getting the events to n8n and making sure the right events make it.

I don't think it will be spammy when it comes to PR pushes, open source contributions are usually well thought and not just a burst of commits with a message of "small fix"

Comments might be more of an issue but we'll have to see.

I am not sure how bots are categorized, the different options in `author_association` are: Owner, Member, Collaborator, Contributor, First_timer, First_time_contributor, None

Online it says that bots are usually treated as none or contributor, so it probably will send a message for those, if it becomes too spammy we'll figure out a way to filter those out (there aren't that many bot users so it can even be a closed list)

VenfiOranai (approved):

PR #3748, opened 2026-08-14
[Core] Use Lua scripts for atomic Redis stream operations in cluster mode

Hadar-Vaknin:
# Description

Replaces transactional Redis pipelines (pipeline(transaction=True)) with Lua scripts for stream message finalization and PEL requeue operations.

We previously relied on transactional pipelines to make multi-step stream operations atomic.
redis-py explicitly does not support transactional pipelines in cluster mode. In redis.asyncio.cluster.RedisCluster.pipeline():
"Cluster implementation of pipeline does not support transaction or shard_hint."
So pipeline(transaction=True) is not a reliable way to get atomic MULTI/EXEC behavior when running against Redis Cluster.

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Changes core live-events message ack and PEL requeue paths; incorrect Lua or eval args could leave PEL entries stuck or duplicate deliveries, though behavior is intended to match the prior pipeline semantics.
> 
> **Overview**
> Replaces **transactional Redis pipelines** (`pipeline(transaction=True)`) with **Lua scripts** for live-events stream finalize and PEL requeue, because cluster clients do not support transactional pipelines for true atomic multi-step work.
> 
> `ack_and_finalize_stream_entry` now runs **XACK + XDEL** in one `EVAL` instead of a MULTI/EXEC pipeline. A new **`requeue_stream_entry`** helper atomically **XADDs** the updated fields (including incremented `requeue_count`), **ACKs**, and **XDELs** the original id; the PEL requeue worker calls this instead of building its own pipeline.
> 
> Consumer ack paths and PEL tests are updated to assert the correct script bodies and `eval` arguments rather than separate `xadd`/`xack`/`xdel` pipeline calls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68d06730768d928c702f263f0c88be71e79b8f07. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/port_ocean/consumers/redis_stream_utils.py
+++ b/port_ocean/consumers/redis_stream_utils.py
@@ -80,6 +80,12 @@ async def _eval_lua_script(
     numkeys: int,
     *keys_and_args: str,
 ) -> Any:
+    """Run a Lua script on Redis so multiple commands execute atomically.
+
+    Stream ack/finalize and requeue paths issue several commands (e.g. XACK + XDEL).
+    Redis runs each EVAL script as a single atomic unit, so partial updates cannot
+    leave entries half-processed if another client or failure interleaves.
+    """
     return await cast(Awaitable[Any], redis.eval(script, numkeys, *keys_and_args))
 
 
--- a/port_ocean/consumers/redis_stream_utils.py
+++ b/port_ocean/consumers/redis_stream_utils.py
@@ -1,4 +1,6 @@
 import asyncio
+from collections.abc import Awaitable
+from typing import Any, cast
 
 from loguru import logger
 from redis.exceptions import ConnectionError as RedisConnectionError
@@ -72,6 +74,15 @@ return redis.call('XDEL', KEYS[1], ARGV[2])
 """
 
 
+async def _eval_lua_script(
+    redis: RedisClient,
+    script: str,
+    numkeys: int,
+    *keys_and_args: str,
+) -> Any:
+    return await cast(Awaitable[Any], redis.eval(script, numkeys, *keys_and_args))
+
+
 async def ack_and_finalize_stream_entry(
     redis: RedisClient,
     *,
@@ -81,7 +92,8 @@ async def ack_and_finalize_stream_entry(
 ) -> None:
     """Atomically ack a stream entry and delete it using a Lua script."""
     try:
-        await redis.eval(
+        await _eval_lua_script(
+            redis,
             ACK_AND_FINALIZE_STREAM_ENTRY_SCRIPT,
             1,
             stream_key,
@@ -113,7 +125,8 @@ async def requeue_stream_entry(
     for key, value in fields.items():
         field_items.extend((key, value))
 
-    await redis.eval(
+    await _eval_lua_script(
+        redis,
         REQUEUE_STREAM_ENTRY_SCRIPT,
         1,
         stream_key,
--- a/port_ocean/consumers/redis_stream_utils.py
+++ b/port_ocean/consumers/redis_stream_utils.py
@@ -77,8 +77,8 @@ return redis.call('XDEL', KEYS[1], ARGV[2])
 async def _eval_lua_script(
     redis: RedisClient,
     script: str,
-    numkeys: int,
-    *keys_and_args: str,
+    keys: list[str],
+    args: list[str],
 ) -> Any:
     """Run a Lua script on Redis so multiple commands execute atomically.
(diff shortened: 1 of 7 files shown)


port_ocean/consumers/redis_stream_utils.py
```diff
@@ -60,19 +62,44 @@ async def ensure_consumer_group(
...
+"""
+
+REQUEUE_STREAM_ENTRY_SCRIPT = """
+redis.call('XADD', KEYS[1], '*', unpack(ARGV, 3))
+redis.call('XACK', KEYS[1], ARGV[1], ARGV[2])
+return redis.call('XDEL', KEYS[1], ARGV[2])
+"""
+
+
+async def _eval_lua_script(
```
VenfiOranai, line 77: Might be worth adding a docstring explaining that we need Lua for the atomicity

port_ocean/consumers/redis_stream_utils.py
```diff
@@ -60,19 +62,44 @@ async def ensure_consumer_group(
...
+    """Atomically ack a stream entry and delete it using a Lua script."""
     try:
-        async with redis.pipeline(transaction=True) as pipe:
-            await pipe.xack(stream_key, consumer_group, message_id)
-            await pipe.xdel(stream_key, message_id)
-            await pipe.execute()
+        await _eval_lua_script(
+            redis,
+            ACK_AND_FINALIZE_STREAM_ENTRY_SCRIPT,
+            1,
```
VenfiOranai: Possible to make this configurable rather than hard-coded in the function? Like, from ocean_config or something?

port_ocean/consumers/redis_stream_utils.py
```diff
@@ -60,19 +62,44 @@ async def ensure_consumer_group(
         )
 
 
+ACK_AND_FINALIZE_STREAM_ENTRY_SCRIPT = """
+redis.call('XACK', KEYS[1], ARGV[1], ARGV[2])
+return redis.call('XDEL', KEYS[1], ARGV[2])
+"""
+
+REQUEUE_STREAM_ENTRY_SCRIPT = """
+redis.call('XADD', KEYS[1], '*', unpack(ARGV, 3))
```
VenfiOranai, line 71: Does it always have to be `KEYS[1]`? Since users can configure it to their own redis instances, might be nice to have this configurable as well

port_ocean/consumers/redis_stream_utils.py
```diff
@@ -60,19 +62,44 @@ async def ensure_consumer_group(
...
 async def ack_and_finalize_stream_entry(
     redis: RedisClient,
     *,
     stream_key: str,
     consumer_group: str,
     message_id: str,
 ) -> None:
-    """Ack a stream entry and delete it in one transaction."""
+    """Atomically ack a stream entry and delete it using a Lua script."""
     try:
```
VenfiOranai, line 103: Do we want this try/except only on this func, or also the other? If we want for both, might be preferable to move it into the `_eval_lua_script` function

VenfiOranai (approved):

port_ocean/consumers/redis_stream_utils.py
Hadar-Vaknin: Didn't really understand the comment.
KEYS[1] isn’t a fixed Redis key or instance setting, it’s the first entry in the KEYS array passed to EVAL

port_ocean/consumers/redis_stream_utils.py
Hadar-Vaknin: I actually disagree, since the scripts should not be configurable, and I don't want to have the risk of someone changing it to a wrong flow or something.

port_ocean/consumers/redis_stream_utils.py
Hadar-Vaknin: `requeue_stream_entry ` is wrapped in a try-catch in its 1 caller.
I don't want to write the try-catch in the `_eval_lua_script ` since the 2 paths (requeue, ack) have different handling for errors: ack - swallow missing stream, the second recreates.

PR #3750, opened 2026-08-14
[Integration][Github] Add excludeArchived selector to skip archived repositories during discovery

travisgosselin:
## Description

Adds an opt-in `excludeArchived` selector to the GitHub integration, allowing repository discovery to skip archived repositories before any per-repository resources (pull requests, issues, workflows, files, etc.) are fetched.

**Motivation:** large GitHub organizations can accumulate many archived repositories over time. Repository discovery is shared and cached once per resync, but there was previously no way to exclude archived repos from that step — every kind still paid the full per-repository API cost for archived repos even when their data isn't useful. Filtering after discovery doesn't reduce sync time or request volume; filtering before per-repository fetches begin does.

## What changed

- Added `excludeArchived` (bool, default `false`) to the shared selector fields used across per-repository kinds (pull request, issue, repository, workflow, workflow run, release, tag, dependabot alert, code scanning alert, secret scanning alert, deployment, deployment status, branch, collaborator, environment, file, folder, skill, plugin).
- Filtering happens once in `RestRepositoryExporter._fetch_repositories`, covering both the REST list and GitHub Search discovery strategies.
- Explicitly listed `repos` entries are always included regardless of archived status (consistent with existing `repoSearch` behavior) — `excludeArchived` only applies to implicit "all repositories" discovery.
- Added a release-intent changeset (minor/feature).

## Test plan

- `poetry run pytest` (full github integration suite) — 953 passed
- `poetry run mypy` / `ruff check` / `black --check` / `yamllint` — all pass
- New tests cover both discovery strategies (REST list + Search) and confirm explicit `repos` entries are unaffected by the filter.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

---

> [!NOTE]
> **Medium Risk**
> The change touches shared repository discovery used by many resync kinds, but it is opt-in with default false and only narrows which repos are scanned rather than altering auth or data mutation logic.
> 
> **Overview**
> Adds an opt-in **`excludeArchived`** selector (default `false`) on per-repository GitHub kinds so discovery can drop archived repos **before** per-repo API work runs.
> 
> Filtering is centralized in **`RestRepositoryExporter._fetch_repositories`** for both REST list and search discovery; the flag is popped from options so it never hits the GitHub API, and cached discovery keys separately on this setting.
> 
> **`RepoSearchSelector`** and **`RepositorySourceModel`** (files, folders, skills, plugins) inherit the field; resync paths in **`main.py`** and **`AllRepositorySelector`** pass it through. **Explicit `repos` entries are still synced** even when archived, matching existing `repoSearch` behavior.
> 
> Tests cover list/search filtering, selector wiring, and default schema for skill paths; minor feature changeset included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3b16b83270e1a1df02bfd0c9a834700bd2c8bfa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/github/github/core/exporters/repository_exporter.py
```diff
@@ -139,6 +144,10 @@ async def _fetch_repositories(
             search_params,
             incremental_cursor=incremental_cursor,
         ):
+            if exclude_archived:
```
emekanwaoma, line 158: For the Search API path, can we also add archived:false to the GitHub search query when excludeArchived is enabled?

We should keep this client-side filter as a safety net, since we’ve seen cases where GitHub Search still returns repositories with repo.archived == true even when archived:false is present. But pushing the qualifier into the query should reduce wasted search pages/rate limit usage and avoid archived repos occupying the search result window before we filter them locally.

integrations/github/github/core/exporters/repository_exporter.py
```diff
@@ -122,6 +122,11 @@ async def _fetch_repositories(
         search_params = cast(
             Optional[RepoSearchParams], params.pop("search_params", None)
         )
+        # Popped (not left in `params`) so it's never sent as a GitHub API query
```
emekanwaoma: I don’t think this comment should stay. The pop is clear enough, and the cache behavior it describes is a tradeoff rather than an obvious benefit.

integrations/github/integration.py
```diff
@@ -47,7 +47,21 @@
 FILE_PROPERTY_PREFIX = "file://"
 
 
-class RepoSearchSelector(Selector):
+class ExcludeArchivedSelector(BaseModel):
+    exclude_archived: bool = Field(
```
emekanwaoma, line 51: This selector is wired into resync discovery, but I don’t see equivalent handling in the live-event path. Today BaseRepositoryWebhookProcessor gates on repoSearch, but not on excludeArchived.

That means a resource configured with excludeArchived: true can skip archived repos during resync, but still process webhook events from archived repos later. For repository events, an archived repo can also be upserted again instead of being skipped/deleted.

Can we add shared webhook filtering for this selector, similar to the existing repoSearch gate?

integrations/github/tests/test_config_schema.py
```diff
@@ -5,4 +5,6 @@
 def test_skill_path_defaults_omit_unset_repository_filters() -> None:
     defaults = GithubSkillSelector.schema()["properties"]["paths"]["default"]
 
-    assert defaults == [{"path": path} for path in DEFAULT_SKILL_PATHS]
+    assert defaults == [
```
emekanwaoma: The public config field is excludeArchived, and these defaults used to omit unset repository filters.

If the goal is to keep defaults clean, this should probably continue excluding default values, and if serialized, it should use aliases. Otherwise the generated schema/default config exposes internal snake_case for a public selector field.

integrations/github/tests/github/core/exporters/test_repository_exporter.py
```diff
@@ -32,6 +32,21 @@
     },
 ]
 
+TEST_REPOS_WITH_ARCHIVED = [
```
emekanwaoma, line 36: The exporter tests cover resync filtering, but the feature also changes selector semantics for repo-backed kinds. Can we add webhook coverage for excludeArchived: true?

In particular: archived repository payloads should not create/update entities when the selector excludes archived repos, and repository archive/upsert events should not reintroduce resources that resync would skip.

emekanwaoma (changes requested):
Left some comments

integrations/github/github/core/exporters/repository_exporter.py
travisgosselin: Sounds good - removed.

integrations/github/github/core/exporters/repository_exporter.py
travisgosselin: Yep that makes sense. I wasn't sure how you wanted to handle the situation where repoSearch was set to a conflicting value for archive status? My assumption is the root level excludeArchived selector takes precedence. Added some tests around this.

integrations/github/tests/github/core/exporters/test_repository_exporter.py
travisgosselin: I wasn't entirely sure what that looked like - I put something together for you to review when you have a min.

integrations/github/integration.py
travisgosselin: Added!

integrations/github/tests/test_config_schema.py
travisgosselin: thats a good catch, I believe copilot was calling that out as well. Definitely don't want any leaks or contract breaks on this.

integrations/github/tests/test_config_schema.py
```diff
@@ -5,4 +5,6 @@
...
+        {"path": path, "exclude_archived": False} for path in DEFAULT_SKILL_PATHS
+    ]
```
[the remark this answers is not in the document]
travisgosselin: Resolved.

travisgosselin:
Thanks for all the review @emekanwaoma - I've made some updates - I wasn't entirely confident on some of the fixes from your comments. Would appreciate another review, and/or further direction!

integrations/github/tests/test_config_schema.py
emekanwaoma: I think I may have mixed two separate points, so it could have been clearer.

The comment had two concerns:

1. **Bad / real concern:**
The schema was exposing internal snake_case:

     `"exclude_archived": false
     `
     That is a public config contract issue.

2. **Cleanliness concern:**
The schema was also adding a default field that used to be omitted:

     `"excludeArchived": false
     `
     That is valid, but noisy.


The expected change is here:

[integrations/github/integration.py
](https://github.com/port-labs/ocean/pull/3750/changes#diff-352618668ce736511a5de564f4e299286cc93f0445a9f0300368ba6c93452eef)
Inside `GithubSkillSelector.Config.schema_extra`:

```
schema["properties"]["paths"]["default"] = [
    path.dict(by_alias=True, exclude_none=True, exclude_defaults=True)
    for path in default_paths
]
```

Specifically, the important part is adding:

`exclude_defaults=True
`

Not here:

`default=[GithubSkillPattern(path=path) for path in DEFAULT_SKILL_PATHS]
`

That default should stay as-is. The issue is only how that default is serialized into the generated schema.

The matching test change is here:

[integrations/github/tests/test_config_schema.py](https://github.com/port-labs/ocean/pull/3750/changes#diff-24979a025c4e829b4115894c38600a776a140a5d56731ed6f7564c0819396cfb)

Expected result should stay:

`assert defaults == [{"path": path} for path in DEFAULT_SKILL_PATHS]`

emekanwaoma: Hi @travisgosselin 

I resolved this in the commit here - https://github.com/port-labs/ocean/pull/3750/changes/a018e26c5b2a21f4d13fad9fcf262d6961813c75

emekanwaoma (approved):
LGTM

danamsalem-ux (approved):

PR #3793, opened 2026-08-19
[CI] Update ocean workflow to notify on external contributor

MPTG94:
# Description

What - Update the workflow that notifies on external contribution

Why - The workflow previously used `pull_request` as a trigger which meant it did not have access to the necessary repo secrets when an actual external contributor opened a PR. In addition the `author_association` check is not robust enough and does not correctly classify contributors so it was replaced with an API call to GH API

How - Update the workflow to use `pull_reqest_target`, update the workflow to use a fine grained access token to make the necessary calls to GH API, also update the workflow to never fail in order to not be a blocker for PRs

## Type of change


.github/workflows/notify-on-external-contribution.yml
```diff
@@ -1,7 +1,8 @@
 name: Notify on external contribution
 
 on:
-  pull_request:
+  # pull_request_target so fork PRs get secrets; safe since no PR code is checked out
+  pull_request_target:
     types: [opened, synchronize]
```
danamsalem-ux: it might be relevant to add 'reopened' type here for reopened PRs

MPTG94:
```suggestion
    types: [opened, reopened, synchronize]
```

matan84 (approved):

PR #3818, opened 2026-08-24
[Integration][GitHub] Add bulk external custom property values actions

omby8888:
## Summary
- Add `bulk_update_external_custom_property_values` action — sparse PATCH for one property across repositories (`repository_values`), with automatic 100-repo batching and per-item `org` support
- Add `bulk_delete_external_custom_property_values` action — destructive DELETE across a required `orgs` list
- Move `update_repo_external_custom_properties` into `external_custom_properties/` package with shared utils and a lightweight base executor (no redundant `_get_execution_client`)

Port task: https://app.getport.io/taskEntity?identifier=task_ti40mx

## Test plan
- [x] `make test` in `integrations/github` (1002 tests pass)
- [ ] Verify bulk update against GitHub sample API in a test org
- [ ] Verify bulk delete against GitHub sample API in a test org


integrations/github/github/actions/dispatch_workflow_executor.py
```diff
@@ -276,7 +279,12 @@ async def execute(self, run: IntegrationRun) -> None:
         inputs: dict[str, str] = self._parse_inputs(
             run.execution_properties.get("workflowInputs", {})
         )
-        rest_client = await self._get_execution_client(run)
+        clients = await self._get_execution_clients(run)
+        if not clients:
```
VenfiOranai: Don't think this'll ever be relevant, the function always returns exactly one.

integrations/github/github/actions/external_custom_properties/utils.py
```diff
@@ -0,0 +1,106 @@
...
+    return f"external_custom_properties/{run.execution_properties.get('propertyName')}"
+
+
+def external_properties_from_mapping(
+    external_properties_mapping: dict[str, Any],
+) -> list[dict[str, str | None]]:
+    return [
+        {
+            "property_name": name,
+            "value": None if value is None or value == "" else str(value),
```
VenfiOranai: Can you just do `if not value` or can value be a technically falsy value (0, empty list/object)?

integrations/github/github/actions/external_custom_properties/utils.py
```diff
@@ -0,0 +1,106 @@
...
+        f"{encoded_property_name}"
+    )
+
+
+def group_repository_values_by_org(
+    repository_values: Any,
+    *,
+    default_org: str | None,
+) -> dict[str, list[dict[str, str | None]]]:
+    if not repository_values:
```
VenfiOranai: Thoughts on creating a pydantic model and just instantiating that instead of running this custom validation?

integrations/github/github/actions/external_custom_properties/utils.py
```diff
@@ -0,0 +1,106 @@
...
+            }
+        )
+
+    return grouped
+
+
+def raise_external_custom_properties_action_error(
+    error: Exception, action_description: str
+) -> None:
+    error_message = str(error)
```
VenfiOranai: Really minor, but what do you think about running this in an else block?
Currently if the `isinstance` triggers this str conversion is redundant.

integrations/github/github/actions/external_custom_properties/bulk_update_external_custom_property_values_executor.py
```diff
@@ -0,0 +1,104 @@
...
+                endpoint = external_property_values_endpoint(
+                    rest_client.base_url, organization, str(property_name)
+                )
+
+                for repository_batch in batched(
+                    repository_values, REPOSITORY_VALUES_BATCH_SIZE
+                ):
+                    request_count += 1
+                    try:
+                        await rest_client.make_request(
```
VenfiOranai: Do we not want concurrency on these?

integrations/github/github/actions/external_custom_properties/bulk_update_external_custom_property_values_executor.py
```diff
@@ -0,0 +1,104 @@
...
+                            endpoint,
+                            method="PATCH",
+                            json_data={"repository_values": list(repository_batch)},
+                            ignore_default_errors=False,
+                        )
+                    except Exception as error:
+                        raise_external_custom_properties_action_error(
```
VenfiOranai: What do we expect to happen if say, request #3 out of 5 failed? Some will have been sent, but #4 and #5 will have been skipped. Is that acceptable behavior, or should we keep going? Alternatively would we perhaps want to try and revert the first two?

integrations/github/github/actions/external_custom_properties/bulk_delete_external_custom_property_values_executor.py
```diff
@@ -0,0 +1,73 @@
...
+            await create_github_client_for_org(organization)
+            for organization in organizations
+        ]
+
+    async def execute(self, run: IntegrationRun) -> None:
+        property_name = run.execution_properties.get("propertyName")
+        if not property_name:
+            raise InvalidActionParametersException("propertyName is required")
+
+        organizations = run.execution_properties.get("orgs")
```
VenfiOranai: Might be nice to have typing here, currently it marks it as any

integrations/github/github/actions/external_custom_properties/utils.py
omby8888: it can

integrations/github/github/actions/dispatch_workflow_executor.py
omby8888: wdym?

integrations/github/github/actions/external_custom_properties/utils.py
```diff
@@ -0,0 +1,104 @@
...
+class RepositoryValuesInput(BaseModel):
+
+    class RepositoryValueInput(RepositoryGithubValue):
+        org: str | None = None
+
+    org: str | None = None
+    repository_values: list[RepositoryValueInput] = Field(..., min_items=1)
+
+    @validator("repository_values")
+    def repository_values_must_not_be_empty(
```
VenfiOranai: Necessary if you have the fields with no default and a `min_items` validator?

integrations/github/github/actions/external_custom_properties/bulk_update_external_custom_property_values_executor.py
```diff
@@ -0,0 +1,142 @@
...
+                rest_client = await create_github_client_for_org(organization)
+                endpoint = external_property_values_endpoint(
+                    rest_client.base_url, organization, str(property_name)
+                )
+
+                for repository_batch in batched(
+                    repository_values, REPOSITORY_VALUES_BATCH_SIZE
+                ):
+                    patch_operations.append(
+                        lambda rest_client=rest_client,
```
VenfiOranai: This feels a bit hard to track. I asked Jimmy for alternatives and he suggested using partials from functools. What do you think?

integrations/github/github/actions/external_custom_properties/bulk_update_external_custom_property_values_executor.py
```diff
@@ -0,0 +1,142 @@
...
+                            organization,
+                            repository_batch,
+                        )
+                    )
+
+            outcomes = await throttle_batch_operation(
+                patch_operations,
+                MAX_CONCURRENT_BULK_REQUESTS,
+            )
+            failures = [outcome for outcome in outcomes if not outcome.success]
```
VenfiOranai: You've got a bunch of loops here when I'm pretty sure you could have just one

VenfiOranai (approved):

PR #3825, opened 2026-08-24
[Core] Convert IntegrationEventsCallbacks To Class

VenfiOranai:
# Description

What - Converted IntegrationEventsCallbacks from typed dict to dataclass 

Why - Makes it significantly more traceable within the codebase

How -

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Touches core event registration and resync/start dispatch. Behavior should be equivalent, but any external code that treated `event_strategy` as a dict will break.
> 
> **Overview**
> Converts `IntegrationEventsCallbacks` from a `TypedDict` to a dataclass so event listeners are accessed as attributes (`event_strategy.resync`) instead of dict keys.
> 
> `EventsMixin` now constructs the dataclass (still using `defaultdict(list)` for kind maps). Call sites in `BaseIntegration` and `SyncRawMixin` use attribute access; dynamic `"resync"` / `"incremental"` lookup goes through `getattr`. `resync_complete` hooks always run (the old `"resync_complete" in ...` guard is gone). Tests are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22b9899f7c0d50b02be9060962891db9aeb5a9f8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
+++ b/.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
@@ -1,3 +1,7 @@
 bump: minor
 changelog-type: improvement
-changelog: Converted IntegrationEventsCallback from TypedDict to dataclass for easier IDE tracing
+changelog: >-
+  Converted IntegrationEventsCallbacks from TypedDict to dataclass for easier IDE tracing.
+  Breaking (internal): custom integrations that mutate `integration.event_strategy` as a dict
+  (e.g. `event_strategy["resync"][kind]`) must use attribute access instead
+  (e.g. `event_strategy.resync[kind]`); `@ocean.on_resync()` and other registration APIs are unchanged.
--- a/.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
+++ b/.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
@@ -1,0 +1,3 @@
+bump: patch
+changelog-type: improvement
+changelog: Converted IntegrationEventsCallback from TypedDict to dataclass for easier IDE tracing
--- a/.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
+++ b/.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
@@ -1,3 +1,3 @@
-bump: patch
+bump: minor
 changelog-type: improvement
 changelog: Converted IntegrationEventsCallback from TypedDict to dataclass for easier IDE tracing
--- a/port_ocean/core/integrations/mixins/sync_raw.py
+++ b/port_ocean/core/integrations/mixins/sync_raw.py
@@ -86,17 +86,21 @@ class SyncRawMixin(HandlerMixin, EventsMixin):
         logger.info(f"Fetching {resource_config.kind} resync results")
 
         is_incremental = event.event_type == EventType.INCREMENTAL_RESYNC
-        strategy_key = "incremental" if is_incremental else "resync"
-        available_kinds = (
-            self.available_incremental_kinds
+        event_mapping = (
+            self.event_strategy.incremental
             if is_incremental
-            else self.available_resync_kinds
+            else self.event_strategy.resync
         )
 
-        if resource_config.kind not in available_kinds and None not in available_kinds:
-            return unsupported_kind_response(resource_config.kind, cast(list[str], available_kinds))
+        if not (
+            event_mapping.get(resource_config.kind) or event_mapping.get(None)
+        ):
+            return unsupported_kind_response(
+                resource_config.kind,
+                cast(list[str], list(event_mapping.keys())),
+            )
 
-        fns = self._collect_resync_functions(resource_config, strategy_key)
+        fns = self._collect_resync_functions(resource_config, event_mapping)
         logger.info(f"Found {len(fns)} resync functions for {resource_config.kind}")
 
         results, errors = await self._execute_resync_tasks(
@@ -106,9 +110,10 @@ class SyncRawMixin(HandlerMixin, EventsMixin):
         return results, errors
 
     def _collect_resync_functions(
-        self, resource_config: ResourceConfig, strategy_key: str = "resync"
+        self,
+        resource_config: ResourceConfig,
+        event_mapping: dict[str | None, list[Callable[[str], Awaitable[RAW_RESULT]]]],
     ) -> list[Callable[[str], Awaitable[RAW_RESULT]]]:
-        event_mapping = self.event_strategy.resync if strategy_key == "resync" else self.event_strategy.incremental
         fns = [
             *event_mapping[resource_config.kind],
             *event_mapping[None],
--- a/port_ocean/core/integrations/mixins/sync_raw.py
+++ b/port_ocean/core/integrations/mixins/sync_raw.py
@@ -95,7 +95,7 @@ class SyncRawMixin(HandlerMixin, EventsMixin):
         )
 
         if not is_resource_supported(
-            resource_config.kind, self.event_strategy[strategy_key]
+            resource_config.kind, getattr(self.event_strategy, strategy_key)
         ):
             return unsupported_kind_response(resource_config.kind, available_kinds)
 
@@ -111,9 +111,10 @@ class SyncRawMixin(HandlerMixin, EventsMixin):
     def _collect_resync_functions(
         self, resource_config: ResourceConfig, strategy_key: str = "resync"
     ) -> list[Callable[[str], Awaitable[RAW_RESULT]]]:
+        event_mapping = getattr(self.event_strategy, strategy_key)
         fns = [
-            *self.event_strategy[strategy_key][resource_config.kind],
-            *self.event_strategy[strategy_key][None],
+            *event_mapping[resource_config.kind],
+            *event_mapping[None],
         ]
 
         if self.__class__._on_resync != SyncRawMixin._on_resync:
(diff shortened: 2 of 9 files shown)


port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -94,10 +93,8 @@ async def _get_resource_raw_results(
             else self.available_resync_kinds
         )
 
-        if not is_resource_supported(
-            resource_config.kind, self.event_strategy[strategy_key]
-        ):
-            return unsupported_kind_response(resource_config.kind, available_kinds)
+        if resource_config.kind not in available_kinds and None not in available_kinds:
+            return unsupported_kind_response(resource_config.kind, cast(list[str], available_kinds))
```
raskinmaya: Removing `is_resource_supported` changes behavior. Checking `available_*_kinds` passes on empty lists and mutates `defaultdict`s, whereas the old check ensured non-empty handlers. Please restore `mapping.get(kind) or mapping.get(None)` to keep original semantics

raskinmaya: Another one here:
`available_*_kinds` includes `None`, so using `cast(list[str])` just masks the type issue and prints `Available kinds: [None]` on errors. Instead of using `cast`, either update `KindNotImplementedException` to accept `list[str | None]` or filter/format `None` when raising it

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -111,9 +108,10 @@ async def _get_resource_raw_results(
     def _collect_resync_functions(
         self, resource_config: ResourceConfig, strategy_key: str = "resync"
     ) -> list[Callable[[str], Awaitable[RAW_RESULT]]]:
+        event_mapping = getattr(self.event_strategy, strategy_key)
```
raskinmaya: `getattr` strips type checking and risks silent bugs on invalid strings. Since `strategy_key` is strictly `"resync" | "incremental"`, swap `getattr` for an explicit if/else or a typed method like `get_mapping(key: Literal["resync", "incremental"])`

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -94,10 +93,8 @@ async def _get_resource_raw_results(
             else self.available_resync_kinds
         )
 
-        if not is_resource_supported(
-            resource_config.kind, self.event_strategy[strategy_key]
-        ):
-            return unsupported_kind_response(resource_config.kind, available_kinds)
+        if resource_config.kind not in available_kinds and None not in available_kinds:
```
raskinmaya: The old path used O(1) dict lookups via `is_resource_supported`. The new path materializes `list(keys())` on every resync, then two O(n) in scans. Integrations like ADO register many kinds - this runs per resource. A `mapping.get(kind) or mapping.get(None)` check is cheaper and matches old semantics

port_ocean/core/integrations/mixins/sync_raw.py
```diff
@@ -1003,13 +1001,12 @@ async def _resync_reconciliation(
...
-                logger.info("Executing resync_complete hooks")
+            logger.info("Executing resync_complete hooks")
 
-                for resync_complete_fn in self.event_strategy["resync_complete"]:
-                    await resync_complete_fn()
+            for resync_complete_fn in self.event_strategy.resync_complete:
+                await resync_complete_fn()
 
-                logger.info("Finished executing resync_complete hooks")
+            logger.info("Finished executing resync_complete hooks")
```
raskinmaya, line 1014: DSP skips `_resync_reconciliation`, so this doesn't apply there, but for ocean-core processing mode, we now always log `Executing/Finished executing resync_complete hooks` even when `resync_complete` is empty if I understand correctly. Most integrations don't register this hook, so that's two info lines per successful full resync. Not too critical, but why not guard with `if self.event_strategy.resync_complete`?

port_ocean/core/integrations/mixins/sync_raw.py
raskinmaya: Please add test coverage for the missing edge cases around `is_resource_supported` since this is the riskiest part of the PR:
- Specific kind registered
- Catch-all only
- Unregistered kind
- Both catch-all and specific kind registered

port_ocean/core/ocean_types.py
```diff
@@ -53,9 +53,14 @@ class ETLPhase:
...
-class IntegrationEventsCallbacks(TypedDict):
-    start: list[START_EVENT_LISTENER]
-    resync: dict[str | None, list[RESYNC_EVENT_LISTENER]]
-    resync_start: list[BEFORE_RESYNC_EVENT_LISTENER]
-    resync_complete: list[AFTER_RESYNC_EVENT_LISTENER]
-    incremental: dict[str | None, list[INCREMENTAL_EVENT_LISTENER]]
+@dataclass
+class IntegrationEventsCallbacks:
+    start: list[START_EVENT_LISTENER] = field(default_factory=list)
+    resync: dict[str | None, list[RESYNC_EVENT_LISTENER]] = field(
```
raskinmaya, line 59: Not too critical, but `default_factory=lambda: defaultdict(list)` is duplicated. Extracting it to a named function or helper (e.g. `_default_listener_map`) would clean this up and eliminate the repetition. Non-blocking

.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
raskinmaya, line 1: Directly mutating `integration.event_strategy["resync"][kind]` will break after this PR is in. Since `@ocean.on_resync()` is the official API, this is low-risk, but it can still be a breaking change for private (customer-owned) integrations. WDYT about bumping a minor version instead of a patch?

raskinmaya (changes requested):

port_ocean/core/integrations/mixins/sync_raw.py
VenfiOranai: Not super sure I get the behavior change you're talking about here. If `available_kinds` is empty then it'll enter the if due to the first half of the condition (just like it used), and I'm not really seeing what part here can potentially mutate a defaultdict

VenfiOranai: But we have the `None not in available_kinds` check in the if, so we know for sure it's only strings. Only reason I added the casting is since the accursed IDE considers that a runtime check so it didn't recognize it.

port_ocean/core/integrations/mixins/sync_raw.py
VenfiOranai: Not sure the extra cognitive load is worth it. Going off a quick check, even if we have 1000 kinds (which I assume we don't), it takes less than a millisecond.
I can convert to a set if you think it'd be preferable, but not sure there's much of a point.

port_ocean/core/integrations/mixins/sync_raw.py
VenfiOranai: Cursor added that guard originally, but that actually changes behavior. Notice that the if we had there checked that the key exists, not that there's a value in it. The key of course always existed since it was added into the dict on init, so the logs always fired no matter what. I can add the if, but that's a change I'm not sure I want attributed to this PR (in case somebody's debugging and doesn't get why logs they used to have suddenly disappeared)

port_ocean/core/ocean_types.py
VenfiOranai: Feels a bit like overkill for a lambda that just calls a defaultdict

.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
VenfiOranai: I agree that it's a breaking change, thing is we should technically bump majors for those, so not sure there's too much of a diff between minor and patch. I'll add a mention in the changelog for sure, but from what I've seen we treat anything internal/bug as a patch, and features as a minor. Since this is fully internal, I went with patch.
I can change it if you think it's preferable, whatever you think

port_ocean/core/integrations/mixins/sync_raw.py
VenfiOranai: Nice catch, converted to if/else

port_ocean/core/integrations/mixins/sync_raw.py
VenfiOranai: Nice catch, added

.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
raskinmaya: I agree, but at least not a patch. Let's bump a minor version

port_ocean/core/integrations/mixins/sync_raw.py
raskinmaya: Customer mapping size drives the multiplier, not handler count - and `mapping.get(kind) or mapping.get(None)` avoids that on the success path. I'd rather we change that

VenfiOranai: Ah, okie, done (plus a bit more to avoid replicated ifs)

.ocean-release/core/task_tk9lzk_convert_integrationeventscallbacks_to_class.yaml
VenfiOranai: Done

raskinmaya (approved):

PR #3841, opened 2026-08-26
[Core] Task tk17js/b3 kind discovery from integration spec

VenfiOranai:
# Description

What - Utility function to extract kinds from spec file

Why - For test connection we need access to the full list of potential kinds

How - Loads up the local spec file and parses it

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Low Risk**
> Additive utility and exception types with no changes to existing spec loading behavior beyond new call sites that opt into `get_spec_kinds`.
> 
> **Overview**
> Adds **`get_spec_kinds`** in `port_ocean/utils/misc.py` so callers (e.g. test connection) can read the integration’s `.port` spec and return the **unique `kind` values** from `features[].resources[]`.
> 
> Spec loading reuses **`get_spec_file`**; missing spec files raise **`SpecNotFoundError`**, and specs missing expected structure raise **`MalformedSpecError`**. **`SpecFileError`** remains for parse/load failures when a spec file exists.
> 
> Two new exception types were added in **`port_ocean/exceptions/spec.py`** to separate “no spec”, “bad shape”, and existing load/parse errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f43ec9a621d6a8ba3c72b40827dfb2d1af935730. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


port_ocean/utils/misc.py
```diff
@@ -106,6 +110,28 @@ def get_spec_file(path: Path = Path(".")) -> dict[str, Any] | None:
...
+        raise SpecNotFoundError(
+            f"Failed to load spec from {spec_path / '.port'}: "
+            "expected spec.json, spec.yaml, or spec.yml"
+        )
+
+    kinds: set[str] = set()
+    try:
+        for feature in spec["features"]:
+            for resource in feature["resources"]:
+                kinds.add(resource["kind"])
```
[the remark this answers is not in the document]
VenfiOranai: Fixed, specifically seeks exporter in features

VenfiOranai:
@cursoragent please re-review

port_ocean/utils/misc.py
```diff
@@ -106,6 +110,31 @@ def get_spec_file(path: Path = Path(".")) -> dict[str, Any] | None:
...
+            "expected spec.json, spec.yaml, or spec.yml"
+        )
+
+    kinds: set[str] = set()
+    try:
+        for feature in spec["features"]:
+            if feature.get("type") != "exporter":
+                continue
+
+            for resource in feature["resources"]:
```
raskinmaya, line 128: `feature["resources"]` breaks configs that omit `resources`, like ADO and GCP. Use `feature.get("resources") or []` and skip empty blocks. Also, update the test file to assert that missing `resources` is allowed instead of throwing

port_ocean/utils/misc.py
```diff
@@ -106,6 +110,31 @@ def get_spec_file(path: Path = Path(".")) -> dict[str, Any] | None:
     return None
 
 
+def get_spec_kinds(path: str | Path = Path(".")) -> list[str]:
+    """Return the unique resource kinds declared by an integration spec."""
```
raskinmaya, line 114: The custom integration returns `["dynamic"]`, which breaks discovery when real kinds are dynamically loaded from port-app-config (`allow_custom_kinds=True`). We might need to handle that explicitly

port_ocean/utils/misc.py
```diff
@@ -106,6 +110,31 @@ def get_spec_file(path: Path = Path(".")) -> dict[str, Any] | None:
     return None
 
 
+def get_spec_kinds(path: str | Path = Path(".")) -> list[str]:
```
raskinmaya: Let's make sure that GH + Jira default mappings (here + provision-service) don't require updates to the `spec` files

port_ocean/utils/misc.py
```diff
@@ -106,6 +110,31 @@ def get_spec_file(path: Path = Path(".")) -> dict[str, Any] | None:
...
+            if feature.get("type") != "exporter":
+                continue
+
+            for resource in feature["resources"]:
+                kinds.add(resource["kind"])
+    except KeyError as e:
+        raise MalformedSpecError(
+            f"Missing expected key {e} when trying to extract kinds from spec"
+        ) from e
+    return list(kinds)
```
raskinmaya, line 134: Converting from `set` to `list` loses order, which can make outputs non-deterministic across runs. Why not use `dict.fromkeys(...)`?

port_ocean/utils/misc.py
raskinmaya: `get_spec_file` takes `Path` but this function takes `str | Path`. Any reason for that?
Since these are siblings, could you align the signatures to avoid callers converting unnecessarily?

port_ocean/tests/utils/test_misc.py
```diff
@@ -95,3 +100,136 @@ def test_get_spec_file_raises_when_spec_file_is_empty(tmp_path: Path) -> None:
     # Act + Assert
     with pytest.raises(SpecFileError, match="must contain a JSON object"):
         get_spec_file(tmp_path)
+
+
+def test_get_spec_kinds_returns_unique_kinds_from_yaml_spec(tmp_path: Path) -> None:
```
raskinmaya, line 105: Unit fixtures are great, but adding a parametrized test over `integrations/*/.port/spec.{yaml,yml,json}` would catch issues like ADO/GCP non-existing `resources` in spec and prevent future spec drift. It's a quick addition with high value IMO

port_ocean/utils/misc.py
VenfiOranai: As discussed, we'll bring those integrations in line with the rest and add kinds to the spec (separate PR)

port_ocean/utils/misc.py
VenfiOranai: As discussed, skipping over custom is outside the responsibility scope of this function, will be added in a future PR when adding the TC entrypoint

port_ocean/utils/misc.py
VenfiOranai: Don't think keeping the order is really needed (and honestly I wouldn't want anybody to depend on the order based on the spec anyways)

port_ocean/utils/misc.py
VenfiOranai: Done, they do indeed all exist in the spec (for both provision and ocean)

VenfiOranai: Done

port_ocean/tests/utils/test_misc.py
VenfiOranai: Not 100% sure I get what you mean here. How are you imagining this?
Do you mean a test that'll run on all actual real integrations? How could I write it in a way that it actually checks them all without breaking every time a kind is added? Also how do I handle the integrations that currently don't support this (ADO/GCP)?

raskinmaya: I wasn’t suggesting asserting exact kind lists (that would be brittle). The idea is a smoke test: parametrize over `integrations/*/.port/spec.{yaml,yml,json}`, call `get_spec_kinds(path)`, and only assert it returns a `list[str]` without raising. Adding a kind to a spec wouldn’t break it.

The ADO/GCP `exporter` blocks don’t have `resources`, so the test fails today regardless. I think that’s actually a signal that `get_spec_kinds` needs to treat missing resources as `[]` - especially given test connection needs kinds for every integration. Once that’s settled, the parametrized test is ~10 lines and mostly guards against future spec-shape regressions

port_ocean/utils/misc.py
raskinmaya: My concern was run-to-run determinism (`list(set)` can shuffle between processes), not semantic ordering. If we're treating this as an unordered bag of kinds, `list(kinds)` is fine. If we ever need stable output for logging/tests without implying spec order, `sorted(kinds)` would do that

raskinmaya (approved):

PR #3858, opened 2026-08-27
[Core] Task tk17ch/b1 result models and probescope abstraction

VenfiOranai:
# Description

What - Added context lifecycle logic

Why - Will use this to orchestrace/manage test connection

How -

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Low Risk**
> Adds new isolated probe scaffolding and tests with no changes to existing integration or resync paths; Port API reporting is not wired yet.
> 
> **Overview**
> Introduces a new **`port_ocean.core.probe`** package to support future **test connection** orchestration in Ocean core.
> 
> **`ProbeConfig`** holds runtime path, optional kind filter, and probe mode (currently only shallow). **`ProbeCheck`** and related enums model per-kind × scope check status and overall probe lifecycle stages.
> 
> **`ProbeContext`** drives the run: **`initialize`** loads available kinds from the integration spec via **`get_spec_kinds`**, **`add_scopes`** expands scope dicts into pending checks for every available kind (with copied scope data), **`build_request_body`** shapes timestamps and check results for Port reporting, and **`finalize`** / **`fail`** set completion state. **`update_progress`** is the reporting hook today—local runs (no **`probe_id`**) log the full request body at info; remote probes log stage messages at debug only (no HTTP yet in this diff).
> 
> Unit tests cover scope expansion, request body serialization, initialization, logging behavior, and finalize/fail.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3fbadc7609a4ef21a545427aedc9d43560e7efa6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+            for kind in self.available_kinds:
+                check = ProbeCheck(kind=kind, scopes=scope.copy())
+                new_checks.append(check)
+                self.checks.append(check)
+
+        self.update_progress()
+        return new_checks
+
+    def build_request_body(self) -> dict[str, Any]:
+        return {
```
raskinmaya, line 39: Why don't we include `status`, `probe_id` and `mode`? These are all present in the context/config. Won't we need these before the wiring?

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+                    "kind": check.kind,
+                    "scopes": check.scopes,
+                }
+                for check in self.checks
+            ],
+        }
+
+    def initialize(self, config: ProbeConfig | None = None) -> None:
+        self.config = config or ProbeConfig()
+        if self.config.kinds:
```
raskinmaya: `kinds=[]` is falsy, so it falls through to `get_spec_kinds()` and probes everything. Did you mean `kinds is not None`? Empty list from a CLI flag would silently become "probe all" which is not very intuitive IMO

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+            configured_kinds_set = set(self.config.kinds)
+            supported_kinds_set = set(get_spec_kinds(self.config.path))
+            if not configured_kinds_set.issubset(supported_kinds_set):
+                raise InvalidProbeKindsError(
+                    list(configured_kinds_set - supported_kinds_set)
+                )
+
+            self.available_kinds = list(configured_kinds_set)
```
raskinmaya: A little deja-vu maybe, but converting a set to list loses input order and varies between runs. For predictable reporting and reproducible integration tests, I'd preserve `config.kinds` order

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+        self.config = ProbeConfig()
+        self.started_at = datetime.now(timezone.utc)
+        self.ended_at = None
+        self.status = ProbeStatus.IN_PROGRESS
+        self.checks = []
+
+    def add_scopes(self, *scopes: dict[str, str]) -> list[ProbeCheck]:
+        logger.debug("Registering additional scopes", scopes=scopes)
+        new_checks: list[ProbeCheck] = []
+        for scope in scopes:
```
raskinmaya, line 29: If `add_scopes` runs before `initialize`, `available_kinds` is `[]` and nothing is registered - no error, no warning. Integrations might hit this easily. Can we guard with a clear error or at least or warning when `available_kinds` is empty?

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+            for kind in self.available_kinds:
+                check = ProbeCheck(kind=kind, scopes=scope.copy())
```
raskinmaya, line 31: Same `(kind, scopes)` can be appended multiple times if `add_scopes` is called again (can't this happen when an integration discovers scopes incrementally?) - Worth tracking a `(kind, frozenset(scopes.items()))` key

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+        self.started_at = datetime.now(timezone.utc)
+        self.ended_at = None
+        self.status = ProbeStatus.IN_PROGRESS
+        self.checks = []
+
+    def add_scopes(self, *scopes: dict[str, str]) -> list[ProbeCheck]:
+        logger.debug("Registering additional scopes", scopes=scopes)
+        new_checks: list[ProbeCheck] = []
+        for scope in scopes:
+            for kind in self.available_kinds:
```
raskinmaya, line 30: This is `O(kinds x scopes)`. A GitHub-style integration with 15 kinds and 200 orgs = 3,000 checks, and every `add_scopes` calls `update_progress` -> `build_request_body()` over the full list. For local mode that logs the entire body on every scope batch. Can we defer the progress until scope discovery finishes, or batch/throttle updates?

port_ocean/core/probe/models.py
```diff
@@ -0,0 +1,36 @@
...
+    FINALIZE = "finalize"
+    FAIL = "fail"
+
+
+@dataclass
+class ProbeCheck:
+    """Represents a singular probe check performed on a specific kind x scope."""
+
+    kind: str
+    scopes: dict[str, str] = field(default_factory=dict)
```
raskinmaya: Some integrations use numeric IDs IIRC. Forcing `str` values means every integration coerces at the boundary. `dict[str, str | int]` or `dict[str, Any]` with JSON-serializable validation would be more honest

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+
+from loguru import logger
+
+from port_ocean.core.probe.config import ProbeConfig
+from port_ocean.core.probe.models import ProbeCheck, ProbeReportStage, ProbeStatus
+from port_ocean.exceptions.probe import InvalidProbeKindsError
+from port_ocean.utils.misc import get_spec_kinds
+
+
+class ProbeContext:
```
raskinmaya, line 15: Why not a `@dataclass`? Manual `__init__` + class-level annotations duplicates what `EventContext` and other contexts do with `@dataclass`. That would be less boilerplate, same typing & easier to extend

port_ocean/exceptions/probe.py
```diff
@@ -0,0 +1,6 @@
+from port_ocean.exceptions.base import BaseOceanException
+
+
+class InvalidProbeKindsError(BaseOceanException):
+    def __init__(self, kinds: list[str]):
+        super().__init__(f"Invalid probe kinds: {kinds}")
```
raskinmaya: Error lists invalid kinds but not supported ones. When an integration dev typos `repositories` vs `repository`, showing `supported_kinds` saves a debug cycle

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+            ProbeReportStage.INIT: "Reporting probe start to Port for probe {probe_id}",
+            ProbeReportStage.UPDATE: "Reporting probe progress to Port for probe {probe_id}",
+            ProbeReportStage.FINALIZE: "Reporting final probe result to Port for probe {probe_id}",
+            ProbeReportStage.FAIL: "Reporting fatal probe error to Port for probe {probe_id}",
+        }[stage].format(probe_id=self.probe_id)
+
+        if self.probe_id is None:
+            logger.info(message, request_body=self.build_request_body())
+            return
+        logger.debug(message)
```
raskinmaya: Let's add a TODO here to report to port

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+            configured_kinds_set = set(self.config.kinds)
+            supported_kinds_set = set(get_spec_kinds(self.config.path))
+            if not configured_kinds_set.issubset(supported_kinds_set):
+                raise InvalidProbeKindsError(
+                    list(configured_kinds_set - supported_kinds_set)
+                )
+
+            self.available_kinds = list(configured_kinds_set)
+        else:
+            self.available_kinds = get_spec_kinds(self.config.path)
```
raskinmaya: `get_spec_kinds` returns `[]` when spec has no exporter resources. Probe initializes as "in progress" with zero checks - is that success or a misconfiguration? Third-party integrations with only gitops features would probe nothing silently

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.status = ProbeStatus.IN_PROGRESS
+        self.update_progress(ProbeReportStage.INIT)
+
+    def update_progress(
+        self, stage: ProbeReportStage = ProbeReportStage.UPDATE
+    ) -> None:
+        message = {
```
raskinmaya: The `ProbeReportStage` dict is rebuilt on every progress call. You can move the stage->message map to a module-level constant

raskinmaya (changes requested):

port_ocean/core/probe/context.py
VenfiOranai: Done

port_ocean/core/probe/context.py
VenfiOranai: Done

port_ocean/core/probe/context.py
VenfiOranai: It's needed in order to remove duplicates. I can add a validation earlier in the flow (at the entrypoint) and have it throw an error instead, let me know what you prefer.

port_ocean/core/probe/context.py
VenfiOranai: It'll never run before initialize unless somebody decides to be really silly. If you check out the next pr in the list, you'll see that it's called before entering the integration, so it isn't called within it at all. The validation becomes kinda pointless in that case.

port_ocean/core/probe/context.py
VenfiOranai: Not 100% sure I get the issue you're raising here, it's built to support multiple calls.
You mean if somebody calls it multiple times with the same data? If that's the case, that's kinda on them, I see no big reason to add validations on something like that.

port_ocean/core/probe/context.py
VenfiOranai: That's what we want though, no? To allow for the "streaming" effect rather than them just sitting and waiting for it to all finish at once.

port_ocean/core/probe/models.py
VenfiOranai: Added int

port_ocean/core/probe/context.py
VenfiOranai: Done

port_ocean/exceptions/probe.py
VenfiOranai: Done

port_ocean/core/probe/context.py
VenfiOranai: Not a huge fan of sprinkling todos in the codebase, especially seeing as I've got a task to do that within the MVP so it's not like it'll be there for long...

port_ocean/core/probe/context.py
VenfiOranai: Umm, not sure we'll ever really see that happen tbh, especially considering that I'll migrate this over to use `integration.py` in a future pr.

port_ocean/core/probe/context.py
VenfiOranai: Changed logic here a fair bit, don't think this is still relevant

port_ocean/core/probe/context.py
raskinmaya: Let's add `ordered` then

port_ocean/core/probe/context.py
raskinmaya: The concern isn't "don't update until the end", it's the cost of how we update today. Every `add_scopes` calls `update_progress()` -> `build_request_body()` over the full check list. In local mode (`probe_id` is None) that also logs the entire body at info on every batch.
For something like 15 kinds * 200 orgs discovered in batches, that's a lot of redundant serialization/logging even though only a small delta changed.

Would something like this work?
- Keep streaming updates for remote probes
- In local mode, log only the new checks from the batch (not the full body)
- A bit crazy but maybe even let integrators batch `add_scopes` and call `update_progress` explicitly when discovery finishes?

port_ocean/core/probe/reporters/file.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def report(self, report: dict[str, Any]) -> None:
+        reports_directory = self.config.path / PROBE_REPORTS_DIRECTORY
+        reports_directory.mkdir(parents=True, exist_ok=True)
+        timestamp = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%S_%fZ")
+        report_path = reports_directory / f"probe_report_{timestamp}.json"
+        report_path.write_text(
+            json.dumps(report, indent=2, default=str),
+            encoding="utf-8",
+        )
```
raskinmaya, line 26: Every `add_scopes` -> `update_progress` -> `report` creates a new JSON file. For a GitHub-style probe (15 kinds x 200 orgs, discovered in batches), that's hundreds of full snapshots on disk. Either overwrite a single `latest.json`, or append deltas. As written, `FILE` mode will spam the integration directory during scope discovery

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,98 @@
...
+    available_kinds: list[str] = field(default_factory=list)
+    config: ProbeConfig = field(default_factory=ProbeConfig)
+    started_at: datetime = field(default_factory=lambda: datetime.now(timezone.utc))
+    ended_at: datetime | None = None
+    status: ProbeStatus = ProbeStatus.IN_PROGRESS
+    message: str | None = None
+    checks: list[ProbeCheck] = field(default_factory=list)
+    reporter: ProbeReporter | None = None
+
+    def add_scopes(self, *scopes: dict[str, str]) -> list[ProbeCheck]:
```
raskinmaya: Signature is `dict[str, str]` but `ProbeCheck.scopes` is `dict[str, str | int]`. Align the signature or callers will fight the type checker or need casts

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,98 @@
...
+            supported_kinds = get_spec_kinds(self.config.path)
+            supported_kinds_set = set(supported_kinds)
+            if not configured_kinds_set.issubset(supported_kinds_set):
+                raise InvalidProbeKindsError(
+                    list(configured_kinds_set - supported_kinds_set), supported_kinds
+                )
+
+            self.available_kinds = sorted(list(configured_kinds_set))
+        else:
+            self.available_kinds = get_spec_kinds(self.config.path)
```
raskinmaya: Configured kinds are `sorted()` as discussed, but "probe all kinds" uses `get_spec_kinds()` which returns `list(set(...))`, which still results in nondeterministic order. Tests and reports will differ depending on path. Please sort in both branches for stable output

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,98 @@
...
+        return {
+            "probe_id": self.probe_id,
+            "status": self.status,
+            "mode": self.config.mode,
+            "started_at": self.started_at.isoformat(),
+            "ended_at": self.ended_at.isoformat() if self.ended_at else None,
+            "message": self.message,
+            "checks": [
+                {
+                    "status": check.status,
```
raskinmaya, line 48: This puts raw `StrEnum` objects in the dict (`status`, `mode`, `stage`). File reporter relies on `default=str`. Could you serialize with `.value` instead?

port_ocean/core/probe/reporters/base.py
```diff
@@ -0,0 +1,22 @@
+from abc import ABC, abstractmethod
+from typing import Any, ClassVar
+
+from port_ocean.core.probe.config import ProbeConfig
+from port_ocean.core.probe.models import ProbeReportingMode
+
+
+class ProbeReporter(ABC):
+    mode: ClassVar[ProbeReportingMode]
+    BY_MODE: ClassVar[dict[ProbeReportingMode, type["ProbeReporter"]]] = {}
```
raskinmaya: `__init_subclass__` registry for 3 fixed reporters is more machinery than this needs IMO. A plain `REPORTERS: dict[ProbeReportingMode, type[ProbeReporter]]` in `reporters/__init__.py` would be simpler, easier to test, and matches how little extensibility we actually have today. WDYT?

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def build_request_body(self) -> dict[str, Any]:
+        return {
+            "probe_id": self.probe_id,
+            "status": self.status,
+            "mode": self.config.mode,
+            "started_at": self.started_at.isoformat(),
+            "ended_at": self.ended_at.isoformat() if self.ended_at else None,
+            "message": self.message,
+            "checks": [
```
raskinmaya, line 46: Manual dict assembly duplicates `ProbeCheck` fields and will drift when we add fields (e.g. `duration`, error code). Why not use `dataclasses.asdict` + enum normalization, or a small `ProbeCheck.to_dict()`/shared serializer that will be used by all reporters?

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def update_progress(
+        self, stage: ProbeReportStage = ProbeReportStage.UPDATE
+    ) -> None:
+        if not self.reporter:
+            raise ValueError("Reporter is not initialized")
+
+        self.reporter.report(
+            {
+                "stage": stage,
```
raskinmaya: Payload has both `stage` (`init`/`update`/`finalize`/`fail`) and `status` (`IN_PROGRESS`/`COMPLETED`/`FAILED`). Consumers need a mapping doc - e.g. is `stage=fail` always paired with `status=FAILED`? Can we drop one or make the relationship explicit in a typed model?

port_ocean/core/probe/reporters/__init__.py
raskinmaya, line 1: Why isn't `ProbeReportingMode` exported from the package? It's part of the public config surface but not re-exported. Either export it or keep it internal, right now it's in an awkward middle ground

port_ocean/core/probe/context.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self.reporter = ProbeReporter.BY_MODE[self.config.reporting_mode](self.config)
+        self.status = ProbeStatus.IN_PROGRESS
+        self.update_progress(ProbeReportStage.INIT)
+
+    def update_progress(
+        self, stage: ProbeReportStage = ProbeReportStage.UPDATE
+    ) -> None:
+        if not self.reporter:
+            raise ValueError("Reporter is not initialized")
```
raskinmaya: Not too critical, but the rest of the codebase uses `BaseOceanException` subclasses for domain errors (`IntegrationNotStartedException`, etc) - a `ProbeNotInitializedError` would be more consistent and easier to catch than bare `ValueError`

port_ocean/core/probe/reporters/file.py
VenfiOranai: Done

port_ocean/core/probe/context.py
VenfiOranai: Done

port_ocean/core/probe/context.py
VenfiOranai: Done

port_ocean/core/probe/reporters/base.py
VenfiOranai: Done

port_ocean/core/probe/context.py
VenfiOranai: Not sure I see the point. This function is supposed to be the only place where we construct data about a probe

port_ocean/core/probe/context.py
VenfiOranai: I dropped stage, no longer needed now that we have status

port_ocean/core/probe/reporters/__init__.py
VenfiOranai: Added under the general probe init since it's in models

port_ocean/core/probe/context.py
VenfiOranai: Done

raskinmaya (approved):

PR #3861, opened 2026-08-27
[Core] Task tk17ht/b2 ocean on check decorator

VenfiOranai:
# Description

What - Decorator for probe mode

Why - To be able to trigger probes on integrations

How -

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Introduces a new integration execution path and event type without changing default server startup, but probe handlers will run real integration logic and future Port reporting could affect operational workflows.
> 
> **Overview**
> Adds **integration probing** to Ocean core so integrations can verify connectivity and report structured check results (by kind and scope) back to Port.
> 
> Integrations register an **`on_probe`** handler (via `ocean.on_probe()` / `EventsMixin.on_probe`), mirroring other lifecycle decorators. **`BaseIntegration.run_probe`** runs that listener inside an **`ON_PROBE`** event context, builds a **`ProbeContext`** (status, checks, timestamps), and raises **`ModeNotSupportedException`** when no probe handler is registered.
> 
> New **`port_ocean/core/probe`** package defines **`ProbeConfig`** (path, optional kinds, shallow mode), **`ProbeCheck`** / status enums, and **`ProbeContext`** helpers (`add_scopes`, `build_request_body`, `initialize` / `finalize` / `fail`). Progress reporting currently **logs** the payload (info when `probe_id` is absent, debug when remote).
> 
> **`load_ocean_app`** in bootstrap centralizes loading an integration’s `app` from `main.py` without starting the server; **`run()`** uses it, and new **`run_probe()`** in `run.py` loads the app and invokes `run_probe` asynchronously. Core release intent added in `.ocean-release/core/probe-context.yaml`; **`ProbeContext`** unit tests and a ruff **`F401`** ignore for probe `__init__.py` re-exports are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2006dac246ae03090b7558a8a0656c739fd1eeb7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


port_ocean/core/integrations/base.py
```diff
@@ -91,3 +95,35 @@ async def run_on_start_tasks() -> None:
...
         await event_listener.start()
+
+    async def run_probe(
+        self,
+        probe_id: str | None = None,
+        config: ProbeConfig | None = None,
+    ) -> ProbeContext:
+        """Invoke the registered ``on_probe`` listener."""
+        context = ProbeContext(probe_id)
+        context.initialize(config)
```
raskinmaya: Correct me if I'm wrong, but `initalize` reads the spec, validates kinds and emits an `IN_PROGRESS` report before we know an `on_probe` handler exists. For integrations without a probe handler, callers get a spurious progress report followed by `fail()` + `ModeNotSupportedException`. Why not check `self.event_strategy.on_probe` first, then call `initialize`?

port_ocean/core/integrations/base.py
```diff
@@ -91,3 +95,35 @@ async def run_on_start_tasks() -> None:
...
+            trigger_type="machine",
+        ):
+            try:
+                returned = await listener(context)
+            except Exception as error:
+                context.fail(str(error))
+                raise
+
+            final_context = returned or context
+            if final_context.status is ProbeStatus.FAILED:
```
raskinmaya, line 130: Failure handling is inconsistent: (a) no listener -> `fail()` + `ModeNotSupportedException`, (b) listener raises -> `fail()` + re-raise original exception, (c) listener sets `status=FAILED` -> `ProbeFailedError` with no `fail()`/`update_progress()`. Why should callers see different outcomes for the same semantic failure? Can't we consider one path? e.g. always call `fail(message)` and raise `ProbeFailedError`, or even always re-raise a single exception type

port_ocean/core/integrations/base.py
```diff
@@ -91,3 +95,35 @@ async def run_on_start_tasks() -> None:
...
+                raise ProbeFailedError(final_context.message or "Probe failed")
```
raskinmaya, line 131: If the listener sets `status=FAILED` without calling `context.fail()`, we raise immediately - no `ended_at`, no final reporter payload. The test encodes this. I think either require `fail()` for failure, or call `fail(message)` here before raising so Port/file reporters always get a terminal update, WDYT?

port_ocean/core/integrations/base.py
```diff
@@ -91,3 +95,35 @@ async def run_on_start_tasks() -> None:
...
+        self,
+        probe_id: str | None = None,
+        config: ProbeConfig | None = None,
+    ) -> ProbeContext:
+        """Invoke the registered ``on_probe`` listener."""
+        context = ProbeContext(probe_id)
+        context.initialize(config)
+        listener = self.event_strategy.on_probe
+        if listener is None:
+            error = ModeNotSupportedException(
```
raskinmaya, line 112: Why not use `NotImplementedError` instead?

port_ocean/core/integrations/mixins/events.py
```diff
@@ -72,3 +73,9 @@ def on_incremental_resync(
             logger.info(f"Registering incremental resync listener for kind {kind}")
             self.event_strategy.incremental[kind].append(function)
         return function
+
+    def on_probe(self, function: ON_PROBE_EVENT_LISTENER) -> ON_PROBE_EVENT_LISTENER:
+        """Register a function that probes the integration's connection."""
+        logger.info("Registering connection probe listener")
+        self.event_strategy.on_probe = function
```
raskinmaya, line 80: Unlike `on_resync` (list per kind), `on_probe` is a single slot that gets overwritten with no warning. A double `@ocean.on_probe()` (copy-paste, reload) silently drops the first handler. Just for kastach, I think it might be worth to add a `logger.warning` if already set, or raise like duplicate route registration

port_ocean/core/ocean_types.py
```diff
@@ -9,6 +9,7 @@
     NamedTuple,
 )
 from port_ocean.core.models import Entity
+from port_ocean.core.probe import ProbeContext
```
raskinmaya, line 14: This file is a low-level typing module; importing `ProbeContext` couples it to the probe package and risks circular imports. Maybe consider `TYPE_CHECKING` + string annotation, or a Protocol defined here with `ProbeContext` only referenced in the mixin?

port_ocean/bootstrap.py
```diff
@@ -22,3 +25,18 @@ def create_default_app(
...
+) -> Ocean:
+    """Load an integration's configured Ocean app without starting its runtime."""
+    spec = get_spec_file(Path(path))
+    config_factory = (
+        default_config_factory(spec.get("configurations", [])) if spec else None
+    )
+    default_app = create_default_app(path, config_factory, config_override)
+
+    app_module = load_module(str(Path(path) / "main.py"))
+    return {name: item for name, item in getmembers(app_module)}.get("app", default_app)
```
raskinmaya, line 42: Good extraction from `run()`, but the same `getmembers(...).get("app", default_app)` pattern still exists in cli/commands/defaults/dock.py, clean.py, and tests/helpers/ocean_app.py. Worth exporting a `load_app_from_module(path)` -> Ocean helper here and reuse it, or this refactor only moves duplication

port_ocean/core/integrations/base.py
VenfiOranai: That way it safely returns within the context of the flow, with a proper report. Without this users would just see it fail without a result where they expect.

port_ocean/core/integrations/base.py
VenfiOranai: The reason we don't call `fail()` in branch c is because in order for the condition to be true here, it means that `fail()` already was called inside the probe check. And not really seeing the benefit to always raising the same ProbeFailedException (specific exceptions help with traceability)

port_ocean/core/integrations/base.py
VenfiOranai: The listener isn't supposed to set it manually, only through `finalize()` or `fail()`. I can make it protected with a getter and no setter if you'd prefer, but not sure that's really necessary. Let me know what you think

port_ocean/core/integrations/base.py
VenfiOranai: I thought the convention was custom errors, no?

port_ocean/core/integrations/mixins/events.py
VenfiOranai: Feels a bit extra for a case that isn't really gonna be relevant ever

port_ocean/core/ocean_types.py
VenfiOranai: Used `TYPE_CHECKING`

port_ocean/bootstrap.py
VenfiOranai: Those spots seem similar, not identical, so I feel like that's a cleanup that's a touch out of this PR's scope.

port_ocean/core/integrations/base.py
raskinmaya: My concern is narrower: we currently emit two reports (`IN_PROGRESS`, then `FAILED`) for an integration that simply doesn't implement probe. Why even treat it as a probe lifecycle? We can check listener -> `initialize()` only when needed -> single `fail()` report. IMO `IN_PROGRESS` for probe not supported is weird

port_ocean/core/integrations/base.py
raskinmaya: On branch (c): `test_run_probe_raises_when_context_is_failed` reaches that branch by setting `context.status = FAILED` directly - without calling `fail()`, so the code path doesn't always imply `fail()` already ran.
I'm fine keeping distinct exception types for (a) vs (b). For (c), I'd either:
- Document that listeners must call `context.fail()` (and drop/replace that test), or
- Call `fail(message)` in `run_probe` before raising `ProbeFailedError` as a safety net

Does locking the contract to `fail()/finalize()` only make sense?

port_ocean/core/integrations/base.py
raskinmaya: In that case I'd update` test_run_probe_raises_when_context_is_failed` to call `context.fail(message)` instead of mutating status directly. That keeps the test aligned with how integrations should behave and guarantees `ended_at` + reporter payload

port_ocean/core/integrations/base.py
raskinmaya: Both patterns exist in core today:

- Custom `BaseOceanException` subclasses for runtime/config errors (`IntegrationAlreadyStartedException`, `InvalidProbeKindsError`, etc)
- `NotImplementedError` specifically for missing handler implementations - see `base.py` for `on_resync`
`ModeNotSupportedException` is defensible since it's user-facing ("github does not support probe mode"), but it's a different category than `on_resync`

Fine to keep if you prefer the typed error tho, not a blocker

port_ocean/core/integrations/base.py
VenfiOranai: I see, moved initialize down

port_ocean/core/integrations/base.py
VenfiOranai: I'll go with option 1 (if fail isn't called that means it's a successful completion, either full success or success_with_fails or whatever it's called). Where would you suggest documenting it?

port_ocean/core/integrations/base.py
VenfiOranai: Done

port_ocean/core/integrations/base.py
VenfiOranai: I think I'll keep it, big fan of custom errors

port_ocean/core/ocean_types.py
VenfiOranai: Apparently since it's a type definition and not just typing, Python gets mad. Currently there aren't really any circular imports, if anything comes up in the future I can make a separate file

port_ocean/core/integrations/base.py
raskinmaya: I think a comment above the `FAILED` check in `run_probe` should be enough for future readers/AI. Something like:

```
# listeners should use context.fail(), not set status directly
if final_context.status is ProbeStatus.FAILED:
  ...
```

Up to you if you want the same thing on `ProbeContext.fail()`/`finalize()` too

port_ocean/core/ocean_types.py
raskinmaya: I think we can keep `TYPE_CHECKING` and avoid the runtime error by using string forward refs in the `Callable`, something like:

```
if TYPE_CHECKING:
    from port_ocean.core.probe import ProbeContext

ON_PROBE_EVENT_LISTENER = Callable[
    ["ProbeContext"], Awaitable["ProbeContext | None"]
]
```

That way `ocean_types` doesn’t import probe at runtime, but type checkers still resolve `ProbeContext`. I think it's quite small and worth adding

raskinmaya (approved):

PR #3869, opened 2026-08-30
[Core] Base64-Encode Oversized Event Added To Queue Payloads

omby8888:
## Summary

Fixes [PORT-18426](https://getport.atlassian.net/browse/PORT-18426).

New Relic silently drops log events that exceed its 255-attribute limit. `Event Added To Queue` logs for large GitHub `pull_request` webhooks were being dropped because both the full payload and all headers were emitted in a single log entry.

This PR splits `Added To Queue` into **two correlated log entries**:
- `event_context=payload` — includes `trace_id`, `timestamp_type`, and the webhook payload
- `event_context=headers` — includes `trace_id`, `timestamp_type`, and the request headers

Both entries share the same `trace_id` so they can be correlated during investigation. Other timestamp types (`Started Processing`, `Finished Processing`, etc.) are unchanged.

## Test plan

- [x] Updated `test_setTimestamp_logsTraceIdAtTopLevelExtra` to assert two separate log records with correct `event_context` fields
- [x] `pytest port_ocean/tests/core/handlers/webhook/test_webhook_event.py` passes
- [x] Full `pytest -m 'not smoke'` suite passes (1112 tests)

[Slack Thread](https://getport.slack.com/archives/C0B1UK48WUU/p1787842467752659?thread_ts=1787842467.752659&cid=C0B1UK48WUU)

<div><a href="https://cursor.com/agents/bc-278887cd-01aa-5102-ac06-49934eb8807a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-278887cd-01aa-5102-ac06-49934eb8807a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

[PORT-18426]: https://getport.atlassian.net/browse/PORT-18426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


port_ocean/core/handlers/webhook/webhook_log_context.py
```diff
@@ -0,0 +1,80 @@
+import base64
+import json
+from typing import Any
+
+EventPayload = dict[str, Any]
+
+# Stay below New Relic's 255-attribute limit after JSON flattening.
+_MAX_FLAT_PAYLOAD_ATTRIBUTES = 200
+# Cap JSON size before base64 to stay within NR blob storage (~128KB).
+_MAX_BASE64_PAYLOAD_JSON_UTF8_BYTES = 120 * 1024
```
omby8888: drop newrelic mentions as this is agnostic to ocean

port_ocean/core/handlers/webhook/webhook_log_context.py
```diff
@@ -0,0 +1,80 @@
...
+        if isinstance(node, list) and node:
+            total = 0
+            for item in node:
+                total += _count(item)
+                if total >= stop_at:
+                    return stop_at
+            return total
+        return 1
+
+    return _count(value)
```
omby8888: value is o=always a dict imo, and u can use length instead of this for loops

emekanwaoma: Payload is always a dict, but len(payload) / JSON byte size only tell us how big the object is and not how many log attributes it becomes after nested JSON is flattened.

A GitHub PR payload can have over 20 top-level keys and still flatten to 300+ attributes. That’s why we walk nested leaves (and stop once we’re over the cap) instead of using length.

port_ocean/core/handlers/webhook/webhook_log_context.py
```diff
@@ -0,0 +1,81 @@
+import base64
+import json
+from typing import Any
+
+EventPayload = dict[str, Any]
```
VenfiOranai: This typing feels a bit extra considering it's just an alias for a standard dict.

port_ocean/core/handlers/webhook/webhook_log_context.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def count_flat_attributes(payload: EventPayload, *, limit: int | None = None) -> int:
+    """Estimate how many leaf attributes nested JSON would flatten into.
+
+    ``len(payload)`` and JSON byte size only measure how big the object is.
+    Nested values become extra attributes, so we walk leaves until ``limit``.
+    Defaults to one past the base64 threshold used for Added To Queue logs.
+    """
+    stop_at = _MAX_FLAT_PAYLOAD_ATTRIBUTES + 1 if limit is None else limit
```
VenfiOranai: I'm not sure about the way `stop_at` is calculated considering you allow limit to be injected. 
I think it might be preferable to always take limit if it's injected (with the default of None meaning it counts all of the attributes until it's done), and then you'd inject `_MAX_FLAT_PAYLOAD_ATTRIBUTES + 1` in the one place this function is currently called.
Alternatively just don't accept limit and always make `_MAX_FLAT_PAYLOAD_ATTRIBUTES + 1`
Whichever method you prefer, it's just that the current implementation might cause confusion if this function is ever used anywhere else.

port_ocean/core/handlers/webhook/webhook_log_context.py
```diff
@@ -0,0 +1,81 @@
...
+    while truncated and (truncated[-1] & 0b11000000) == 0b10000000:
+        truncated = truncated[:-1]
+    return truncated
+
+
+def _payload_json_bytes(payload: EventPayload) -> bytes:
+    return json.dumps(payload, default=str, separators=(",", ":")).encode("utf-8")
+
+
+def should_base64_payload(payload: EventPayload) -> bool:
```
VenfiOranai: This function feels a bit redundant considering it's pretty minimal and only has one call

port_ocean/core/handlers/webhook/webhook_log_context.py
```diff
@@ -0,0 +1,81 @@
...
+    if len(data) <= max_len:
+        return data
+    truncated = data[:max_len]
+    while truncated and (truncated[-1] & 0b11000000) == 0b10000000:
+        truncated = truncated[:-1]
+    return truncated
+
+
+def _payload_json_bytes(payload: EventPayload) -> bytes:
+    return json.dumps(payload, default=str, separators=(",", ":")).encode("utf-8")
```
VenfiOranai: This function feels a bit redundant considering it's pretty minimal and only has one call

VenfiOranai (approved):

PR #3893, opened 2026-09-01
[Core] Task tkf6n9/b8 discover kinds from integration py rather than spec

VenfiOranai:
# Description

What -

Why -

How -

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Probe initialization now depends on the integration config class instead of the spec file, which can change which kinds are considered valid at runtime; callers of `ProbeContext.initialize` must pass the config class.
> 
> **Overview**
> **Probe** now discovers supported resource kinds from the integration’s `PortAppConfig` / `ResourceConfig` Python types (`get_port_app_config_kinds`) instead of reading kinds from the integration spec path. `ProbeContext.initialize` takes a `port_app_config_class`, and `run_probe` passes `AppConfigHandlerClass.CONFIG_CLASS`.
> 
> Integrations can declare optional **`probe_permissions`** on each `ResourceConfig` (a tuple or a dict keyed by auth mode). **`get_kind_probe_permissions`** resolves those into a kind → permissions map for probes.
> 
> A new **`KindPermissionVerdict`** helper standardizes permission probe checks with **AND** or **OR** semantics, including handling for unmapped kinds, missing permission data, and grant/deny messaging. Unit tests cover the validators and verdict logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7bbb51ac77f06871b828e1daeaf1b1e442a839d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.ocean-release/core/discover_probe_kinds_from_integration.yaml
+++ b/.ocean-release/core/discover_probe_kinds_from_integration.yaml
@@ -1,0 +1,3 @@
+bump: patch
+changelog-type: feature
+changelog: Probe discovers supported resource kinds from the integration PortAppConfig and adds per-kind probe_permissions with KindPermissionVerdict for permission checks
--- a/.ocean-release/core/discover_probe_kinds_from_integration.yaml
+++ b/.ocean-release/core/discover_probe_kinds_from_integration.yaml
@@ -1,3 +1,3 @@
 bump: patch
 changelog-type: feature
-changelog: Probe discovers supported resource kinds from the integration PortAppConfig and adds per-kind probe_permissions with KindPermissionVerdict for permission checks
+changelog: Probe discovers supported resource kinds from the integration PortAppConfig
--- a/port_ocean/core/handlers/port_app_config/validators.py
+++ b/port_ocean/core/handlers/port_app_config/validators.py
@@ -10,6 +10,7 @@ from pydantic.v1 import BaseModel
 from port_ocean.core.handlers.port_app_config.models import (
     CUSTOM_KIND,
     PortAppConfig,
+    ProbePermissions,
     ResourceConfig,
     Selector,
 )
@@ -28,6 +29,71 @@ def validate_and_get_config_schema(
         "kinds": kinds,
         "advancedConfig": _get_advanced_config(config_class),
     }
+
+
+def get_port_app_config_kinds(config_class: Type[PortAppConfig]) -> list[str]:
+    """Return literal resource kind values declared on a PortAppConfig class."""
+    kinds = _build_kinds_mapping(
+        _get_resource_config_models(config_class),
+        config_class.allow_custom_kinds,
+    )
+    return sorted(kind for kind in kinds if kind != CUSTOM_KIND)
+
+
+def get_kind_probe_permissions(
+    config_class: Type[PortAppConfig],
+    *,
+    permission_key: str | None = None,
+) -> dict[str, tuple[str, ...]]:
+    """Return probe permission requirements keyed by resource kind.
+
+    Resource configs may declare ``probe_permissions`` as either a single tuple
+    (one permission namespace) or a dict keyed by auth mode / namespace.
+    When a dict is used, *permission_key* selects the active namespace.
+    """
+    kind_permissions: dict[str, tuple[str, ...]] = {}
+    for model in _get_resource_config_models(config_class):
+        if not (isinstance(model, type) and issubclass(model, ResourceConfig)):
+            continue
+
+        kind_field = model.__fields__.get("kind")
+        if kind_field is None:
+            continue
+
+        kind_value = _resolve_kind_value(
+            kind_field, model.__name__, config_class.allow_custom_kinds
+        )
+        if kind_value is None or kind_value == CUSTOM_KIND:
+            continue
+
+        permissions = _resolve_probe_permissions_for_model(model, permission_key)
+        if permissions:
+            kind_permissions[kind_value] = permissions
+
+    return kind_permissions
+
+
+def _resolve_probe_permissions_for_model(
+    model: type,
+    permission_key: str | None,
+) -> tuple[str, ...] | None:
+    permissions: ProbePermissions | None = getattr(model, "probe_permissions", None)
+    if permissions is None:
+        return None
+    if isinstance(permissions, tuple):
+        return permissions
+    if isinstance(permissions, dict):
+        if permission_key is None:
+            raise ValueError(
+                f"{model.__name__}.probe_permissions is a dict; "
+                "permission_key is required"
+            )
+        resolved = permissions.get(permission_key)
+        return resolved if resolved else None
+    raise TypeError(
+        f"{model.__name__}.probe_permissions must be a tuple or dict, "
+        f"got {type(permissions).__name__}"
+    )
 
 
 def _is_model(annotation: Any, model: type) -> bool:
--- a/port_ocean/core/handlers/port_app_config/validators.py
+++ b/port_ocean/core/handlers/port_app_config/validators.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import importlib
 import json
 import types
+from collections.abc import Iterable, Iterator
 from typing import Any, Literal, Type, Union, get_args, get_origin
 
 from pydantic.v1 import BaseModel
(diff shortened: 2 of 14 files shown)


port_ocean/core/probe/permissions.py
```diff
@@ -0,0 +1,85 @@
...
+class PermissionCombination(StrEnum):
+    AND = "and"
+    OR = "or"
+
+
+class KindPermissionVerdict(ABC):
+    _kind_permissions: Mapping[str, tuple[str, ...]] | None = None
+
+    @abstractmethod
+    def load_kind_permissions(self) -> Mapping[str, tuple[str, ...]]:
```
raskinmaya, line 17: We're introducing `probe_permissions` on `ResourceConfig` + an abstract `load_kind_permissions` here. Why? Integrations will end up maintaining this twice unless we wire the base class to call `get_kind_probe_permissions`. Can't we drop the abstract loader and have a concrete base that reads the `ClassVar`?

port_ocean/core/handlers/port_app_config/validators.py
```diff
@@ -30,6 +31,71 @@ def validate_and_get_config_schema(
     }
 
 
+def get_port_app_config_kinds(config_class: Type[PortAppConfig]) -> list[str]:
+    """Return literal resource kind values declared on a PortAppConfig class."""
+    kinds = _build_kinds_mapping(
```
raskinmaya: This routes through `_build_kinds_mapping`, which builds selector JSON schemas we don't need for probe. Probe init runs on every `run_probe` call - can't we use a lightweight `_iter_resource_kinds` that just resolves `Literal` values instead of constructing the full kinds dict?

port_ocean/core/handlers/port_app_config/validators.py
```diff
@@ -30,6 +31,71 @@ def validate_and_get_config_schema(
...
+    permission_key: str | None = None,
+) -> dict[str, tuple[str, ...]]:
+    """Return probe permission requirements keyed by resource kind.
+
+    Resource configs may declare ``probe_permissions`` as either a single tuple
+    (one permission namespace) or a dict keyed by auth mode / namespace.
+    When a dict is used, *permission_key* selects the active namespace.
+    """
+    kind_permissions: dict[str, tuple[str, ...]] = {}
+    for model in _get_resource_config_models(config_class):
```
raskinmaya: This is a third copy of "walk resource config models, resolve kind field". If we add another metadata field on `ResourceConfig`, we'll touch this loop again. Maybe extract a shared `_iter_resource_configs(config_class) -> Iterator[tuple[type, str]]` and reuse it in `_build_kinds_mapping` and here?

port_ocean/core/probe/permissions.py
```diff
@@ -0,0 +1,85 @@
...
+        if self._kind_permissions is None:
+            self._kind_permissions = self.load_kind_permissions()
+        return self._kind_permissions
+
+    @property
+    @abstractmethod
+    def combination(self) -> PermissionCombination:
+        """Whether all mapped permissions must be granted (AND) or any one (OR)."""
+
+    @abstractmethod
```
raskinmaya: Four required abstract methods just for status strings feels like a lot of boilerplate per integration. Could we provide sensible defaults (e.g. `f"Missing permissions: {', '.join(denied)}"`) and let integrations override only when they need custom copy?

port_ocean/core/probe/context.py
```diff
@@ -65,7 +77,9 @@ async def initialize(self, config: ProbeConfig | None = None) -> None:
 
             self.available_kinds = sorted(list(configured_kinds_set))
         else:
-            self.available_kinds = sorted(get_spec_kinds(self.config.path))
+            self.available_kinds = sorted(
```
raskinmaya: `get_port_app_config_kinds` already returns a sorted list. Wrapping it in `sorted()` again is a no-op

port_ocean/core/probe/context.py
```diff
@@ -65,7 +77,9 @@ async def initialize(self, config: ProbeConfig | None = None) -> None:
 
             self.available_kinds = sorted(list(configured_kinds_set))
```
raskinmaya: Isn't `sorted(configured_kinds_set)` enough here?

port_ocean/tests/core/handlers/port_app_config/test_get_port_app_config_kinds.py
```diff
@@ -0,0 +1,107 @@
...
+    PortAppConfig,
+    ResourceConfig,
+)
+from port_ocean.core.handlers.port_app_config.validators import (
+    get_kind_probe_permissions,
+    get_port_app_config_kinds,
+)
+
+
+def _resources() -> Any:
```
raskinmaya: Not too critical, but this is copy-pasted from `test_context`. Maybe pull it into a shared `conftest`/test helper module?

.ocean-release/core/discover_probe_kinds_from_integration.yaml
```diff
@@ -0,0 +1,3 @@
+bump: patch
+changelog-type: feature
+changelog: Probe discovers supported resource kinds from the integration PortAppConfig and adds per-kind probe_permissions with KindPermissionVerdict for permission checks
```
raskinmaya: This is a bit misleading. We state that probe discovers kinds and adds per-kind permission checks, but no integration actually uses `probe_permissions` or `KindPermissionVerdict` yet. Let's scope this to the kind-discovery change only

port_ocean/core/probe/permissions.py
VenfiOranai: It's technically possible, but I think it'll make the operation heavier since I'd need to scan the integration.py file to find the relevant class (the one that inherits from PortAppConfig) dynamically. 
I could make the implementation a bit simpler by having the get_kind_probe_permissions call in the base and adding a changing the abstractmethod expect the class ref, but that feels pretty minor.
What do you think?

port_ocean/core/probe/permissions.py
VenfiOranai: Smart, added

port_ocean/core/probe/context.py
VenfiOranai: Done

port_ocean/core/probe/context.py
VenfiOranai: Done

.ocean-release/core/discover_probe_kinds_from_integration.yaml
VenfiOranai: Done

port_ocean/core/probe/permissions.py
raskinmaya: Isn't the config class already available via `AppConfigHandlerClass.CONFIG_CLASS` in `run_probe`? I'm not sure that file scanning is needed. We can wire `KindPermissionVerdict` to call `get_kind_probe_permissions(config_class)` by default and drop the abstract `load_kind_permissions`, no? Integrations only need `combination` (and optional `message` overrides)

VenfiOranai: Using AppConfigHandlerClass.CONFIG_CLASS is definitely better than the dynamic way I was imagining, but I still feel the current method is easier. Using AppConfigHandlerClass.CONFIG_CLASS  would require me to wire it through the decorator, then from there into where the class is actually created. Even then though, I'd still need to override it at times in order to inject the permissionKey , like in GH. Plus more minor, but there's the bit of overhead of understanding what AppConfigHandlerClass.CONFIG_CLASS /where it comes from for people who aren't super familiar with it.
The implementation they'd need also isn't really complex, so I'm not too worried there, and since it's an abstract method it's not like anybody can really forget. Comes out to look something like this
```
    def load_kind_permissions(self) -> Mapping[str, tuple[str, ...]]:
        return get_kind_probe_permissions(GithubPortAppConfig, permission_key="app")
```

raskinmaya (approved):

PR #3898, opened 2026-09-02
[Integration][Jira] Task tk19xl/f3 jira check handler

VenfiOranai:
# Description

What -

Why -

How -

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Medium Risk**
> Probe initialization now depends on the integration config class rather than spec files, which affects any integration using probe mode; Jira probe adds live API calls and changes HTTP retry behavior for permission checks.
> 
> **Overview**
> **Probe** now derives supported resource kinds from the integration’s `PortAppConfig` (and optional `probe_permissions` on each `ResourceConfig`) instead of reading kinds from the on-disk spec. `ProbeContext.initialize` takes the config class; `KindPermissionVerdict` standardizes per-kind permission checks (AND/OR).
> 
> **Jira** registers an `@ocean.on_probe` handler that batches Jira permission lookups via new `get_current_user_permissions` (with `skip_retry` on probe calls), maps kinds using declared `probe_permissions`, and handles **team** separately (org ID + Teams API verification).
> 
> Tests cover kind/permission extraction, verdict logic, and updated probe initialization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d31d8ac0aa1d69ae1050b213c4bc80bead079b32. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/jira/jira/client.py
```diff
@@ -468,6 +474,35 @@ async def has_webhook_permission(self) -> bool:
...
+        response = await self._send_api_request(
+            "GET",
+            f"{self.api_url}/mypermissions",
+            params={"permissions": ",".join(permission_keys)},
+            skip_retry=True,
+        )
+        return {
+            key: bool(permission.get("havePermission"))
+            for key, permission in response.get("permissions", {}).items()
+        }
```
raskinmaya, line 493: `has_webhook_permission` and `get_current_user_permissions` duplicate the same API parsing - same `/mypermissions` endpoint, same response shape. Can't `has_webhook_permission` just return `(await self.get_current_user_permissions(["ADMINISTER"])).get("ADMINISTER", False)`? One parser, less drift

integrations/jira/jira/probe/probe.py
```diff
@@ -0,0 +1,120 @@
...
+        except (HTTPStatusError, RequestError):
+            await self.context.fail("Failed to verify Jira authentication.")
+            return
+
+        checks = await self.context.add_scopes({})
+
+        try:
+            permissions = await self.client.get_current_user_permissions(
+                permission_keys
+            )
```
raskinmaya, line 38: Not too critical, but when `permission_keys` is non-empty, `get_current_user_permissions` already proves auth — a 401 there fails the same way. `/myself` only adds value for team-only probes if I understand correctly. Why not skip `verify_current_user()` when you already have keys to check? This would save a call on every mixed-kind probe

integrations/jira/jira/overrides.py
```diff
@@ -48,6 +48,8 @@ class JiraIssueSelector(Selector):
 
 
 class JiraIssueConfig(ResourceConfig):
+    probe_permissions: ClassVar[tuple[str, ...]] = ("BROWSE_PROJECTS",)
```
raskinmaya, line 51: `BROWSE_PROJECTS` is a reasonable shallow check - Jira [docs](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-permissions/#api-rest-api-3-mypermissions-get) say global context is true if the user has it in any project. Just a small nit: "Permission(s) granted: `BROWSE_PROJECTS`" reads like a full sync guarantee; something like "Basic Jira access verified" might better match shallow intent

integrations/jira/jira/probe/probe.py
```diff
@@ -0,0 +1,120 @@
...
+            await self.context.update_progress()
+            return
+
+        try:
+            await self.client.verify_teams_access(org_id)
+        except HTTPStatusError as error:
+            check.status = ProbeCheckStatus.FAILURE
+            check.message = (
+                f"Jira returned HTTP {error.response.status_code} "
+                "while verifying teams access"
```
raskinmaya, line 108: This hits the Atlassian Teams gateway, not Jira REST. When OAuth is missing the teams scope you get a 403 here and the message says "Jira returned HTTP 403" — that's a painful red herring for customers. Something like "Atlassian Teams API returned HTTP..." would be more honest

integrations/jira/jira/probe/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+from port_ocean.core.probe import KindPermissionVerdict, PermissionCombination
+
+
+class JiraKindPermissionVerdict(KindPermissionVerdict):
+    def load_kind_permissions(self) -> dict[str, tuple[str, ...]]:
+        return get_kind_probe_permissions(JiraPortAppConfig)
+
+    @property
+    def combination(self) -> PermissionCombination:
+        return PermissionCombination.AND
```
raskinmaya, line 14: Couldn't this live at the top of `probe.py`? We could then drop `permissions.py`, `__init__.py` re-export and the ruff `F401` exception in `pyproject`

integrations/jira/jira/overrides.py
raskinmaya, line 51: `("BROWSE_PROJECTS",)` is copy-pasted on 8 resource configs. A module-level `_BROWSE_PROJECTS = ("BROWSE_PROJECTS",)` (or a small base class) would make future permission tweaks one edit instead of 8

integrations/jira/tests/probe/test_permissions.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+    denied_status, denied_message = verdict.verdict(
+        "user",
+        {"USER_PICKER": False},
+    )
+
+    assert granted_status is ProbeCheckStatus.SUCCESS
+    assert granted_message == "Permission(s) granted: BROWSE_PROJECTS"
+    assert denied_status is ProbeCheckStatus.FAILURE
+    assert denied_message == "Missing permission(s): USER_PICKER"
```
raskinmaya: The grant/deny assertions here are identical to core's permissions test. The only Jira-specific bit is "team is not in `kind_permissions`". I'd keep that one assertion and drop the rest. Less maintenance when core messages change

integrations/jira/jira/probe/probe.py
```diff
@@ -0,0 +1,120 @@
...
+            }
+        )
+
+        try:
+            await self.client.verify_current_user()
+        except (HTTPStatusError, RequestError):
+            await self.context.fail("Failed to verify Jira authentication.")
+            return
+
+        checks = await self.context.add_scopes({})
```
raskinmaya, line 33: `add_scopes({})` works, but it's not obvious that's intentional. It kinda looks like we're accidentally passing an empty scope.
Could we wrap it in something like:

```
_UNSCOPED: dict[str, str | int] = {}

async def _register_kind_checks(context: ProbeContext) -> list[ProbeCheck]:
    return await context.add_scopes(_UNSCOPED)
```

Then in `run()`:
```
checks = await _register_kind_checks(self.context)
```

It's the same behavior, but the intent is much clearer

integrations/jira/jira/probe/probe.py
```diff
@@ -0,0 +1,120 @@
...
+            await self._resolve_team_check(team_check)
+
+    async def _fail_permission_checks(
+        self,
+        checks: list[ProbeCheck],
+        message: str,
+    ) -> None:
+        for check in checks:
+            if check.kind == Kinds.TEAM:
+                continue
```
raskinmaya: Team is excluded in here, in `_resolve_permission_checks`, and handled separately in `_resolve_team_check` because it has no `probe_permissions`. That's fine, but fragile. Adding another non-Jira-permission kind means touching all three. Maybe a `NON_PERMISSIONS_KINDS` set would help?

integrations/jira/tests/probe/test_probe.py
```diff
@@ -0,0 +1,274 @@
...
+            }
+        ),
+    )
+    mock_ocean = MagicMock()
+    mock_ocean.integration_config = {"atlassian_organization_id": "test-org-id"}
+    monkeypatch.setattr("jira.probe.probe.ocean", mock_ocean)
+    monkeypatch.setattr(
+        "jira.probe.probe.get_or_create_jira_client",
+        lambda: client,
+    )
```
raskinmaya: Almost every test repeats the `get_or_create_jira_client` monkeypatch + optional ocean mock. A `mock_jira_probe` fixture would cut ~40 lines and make new cases easier to add :)

integrations/jira/jira/probe/probe.py
VenfiOranai: I feel that's a much more complex way to handle it. If you think it's an issue I can add another function in core that specifically populates all kinds with no scopes, but that'd be in a separate PR anyways cuz core

integrations/jira/jira/overrides.py
VenfiOranai: I mean, it's a string, comes out the exact same length so we're not improving readability, and I definitely wouldn't want a tweak on one kind to affect others (also not expecting that to ever happen since it'd be an API break on Jira's end)

integrations/jira/jira/probe/permissions.py
VenfiOranai: It definitely could, but the file feels a bit bloated with multiple classes in it

integrations/jira/jira/probe/probe.py
VenfiOranai: Done

integrations/jira/tests/probe/test_permissions.py
VenfiOranai: Done

integrations/jira/jira/probe/probe.py
VenfiOranai: Done

integrations/jira/tests/probe/test_probe.py
VenfiOranai: Done

integrations/jira/jira/client.py
VenfiOranai: Done

integrations/jira/jira/probe/probe.py
VenfiOranai: Jira's my permissions route doesn't return a 401, it'll treat an invalid token as anonymous and return default permissions. The verify_current_user is needed to tell if it's an actual valid token

integrations/jira/jira/overrides.py
VenfiOranai: Not 100% sure I get this one. You mean like override `granted_message` and do something like this?
```
    def granted_message(self, granted: tuple[str, ...]) -> str:
        if granted == ("BROWSE_PROJECTS",):
            return "Basic Jira access verified"
        return super().granted_message(granted)

```

raskinmaya: Yeah, that's the idea - but the override goes on `JiraKindPermissionVerdict` in `permissions.py`, not on the `ResourceConfig` here. That's what builds the success message via `verdict()`. Your snippet would work for browse kinds. `user` would still get the default `USER_PICKER` message unless we map that too

integrations/jira/jira/probe/probe.py
raskinmaya: Yeah, let's open a task for that and do it in a separate PR. I think it would read much better

raskinmaya (approved):

integrations/jira/jira/probe/probe.py
VenfiOranai: Opened, will handle it after we wrap this and GH up

integrations/jira/jira/overrides.py
VenfiOranai: Done

PR #3904, opened 2026-09-02
[Integration][Github] Task tk19pu/f1 gh check handler

VenfiOranai:
# Description

What -

Why -

How -

```yaml
bump: patch
changelog-type: bugfix
```

## Type of change

### Core testing checklist

- [ ] Docs PR link [here](#)

### Preflight checklist

- [ ] Handled rate limiting
- [ ] Handled pagination
- [ ] Support Multi account

## Screenshots

## API Documentation

---

> [!NOTE]
> **Low Risk**
> Additive metadata and optional token field; no change to sync or auth flow unless a separate check handler consumes these values.
> 
> **Overview**
> Adds **permission probing** support for the GitHub integration so configured resource kinds can be checked against PAT scopes or GitHub App installation permissions.
> 
> **`GitHubToken`** now optionally carries a **`permissions`** map from the installation access-token API response, and the app authenticator populates it when minting installation tokens.
> 
> Each **`Github*Config`** resource class gains a **`probe_permissions`** class variable mapping auth mode **`pat`** vs **`app`** to the required permission names (e.g. pull requests → `repo` / `pull_requests`, users → `read:org` / `members`). This plugs into Ocean’s existing `get_kind_probe_permissions` machinery for preflight validation of credentials against enabled mappings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 335a823e769f0ea51ce01f789025075b6ad8df74. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


integrations/github/github/probe/pat/probe.py
```diff
@@ -0,0 +1,151 @@
...
+        if granted_scopes is None:
+            return
+
+        checks = await self.context.add_scopes(*org_scopes(organizations))
+        permissions = {
+            scope: "granted" for scope in expand_pat_scopes(granted_scopes)
+        }
+        await self._resolve_checks(checks, permissions)
+
+    async def _get_organizations(
```
VenfiOranai, line 93: This logic is technically duplicated with the regular exporter. I've had cursor prep up a change that removes the duplication, but since that touches the regular integration and isn't scoped to just probe, let me know if you'd prefer I skip that/do it in a separate PR

integrations/github/integration.py
```diff
@@ -769,6 +844,11 @@ class GithubBranchConfig(ResourceConfig):
 
 
 class GithubOrganizationConfig(ResourceConfig):
+    probe_permissions: ClassVar[dict[str, tuple[str, ...]]] = {
+        "pat": ("repo",),
+        "app": ("metadata",),
+    }
```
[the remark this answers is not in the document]
raskinmaya: @VenfiOranai

integrations/github/github/probe/pat/probe.py
```diff
@@ -0,0 +1,151 @@
...
+            next_link = response.links.get("next")
+            url = next_link.get("url") if next_link else None
+            params = None
+        return organizations
+
+    async def _get_scopes(
+        self,
+        authenticator: AbstractGitHubAuthenticator,
+    ) -> set[str] | None:
+        response = await authenticator.client.get(
```
raskinmaya: We hit `/user` twice - once in `_get_organizations` & again here. One request can give login + `x-oauth-scopes`

integrations/github/github/probe/app/probe.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class GitHubAppPermissionProbe(GitHubPermissionProbeFlow):
+    @property
+    def _permission_verdict_class(self) -> type[KindPermissionVerdict]:
+        return AppKindPermissionVerdict
+
+    async def run(self) -> None:
+        try:
+            tokens = [
```
raskinmaya: Sequential `get_token()` with a broad `except AuthenticationException` means one bad installation fails the entire probe. So if I understand correctly, multi-org installs with a single suspended/misconfigured org get nothing. Can we do it per-installation?

integrations/github/github/probe/pat/probe.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+            return None
+        return {
+            scope.strip()
+            for scope in response.headers.get("x-oauth-scopes", "").split(",")
+            if scope.strip()
+        }
+
+
+def expand_pat_scopes(scopes: set[str]) -> set[str]:
```
raskinmaya, line 135: Let's add some tests for this, fine-grained PAT detection, app permission levels, org slicing. Should be easy to mock, especially scope expansion

integrations/github/integration.py
```diff
@@ -295,6 +300,11 @@ class GithubFileSelector(Selector, IncludedFilesConfig):
 
 
 class GithubFileResourceConfig(ResourceConfig):
```
raskinmaya, line 302: Similar to one my comments on Jira's PR - I'll test my luck here too. `probe_permissions` is copy-pasted ~23 times. Same `{"pat": ("repo",), "app": ("contents",)}` block appears on file/skill/plugin/folder/branch/release/tag/workflow/etc. A few module-level constants would cut noise

integrations/github/github/probe/pat/permissions.py
```diff
@@ -0,0 +1,14 @@
+from typing import Mapping
+
+from integration import GithubPortAppConfig
+from port_ocean.core.handlers.port_app_config.validators import get_kind_probe_permissions
+from port_ocean.core.probe import KindPermissionVerdict, PermissionCombination
+
+
+class PatKindPermissionVerdict(KindPermissionVerdict):
```
raskinmaya, line 10: Do we expect `PatKindPermissionVerdict` and `AppKindPermissionVerdict` to diverge soon?
If not, we could make one parameterized verdict class (key + optional level checker), since these two classes only differ by `permission_key` and app's `is_granted` override

integrations/github/github/clients/auth/abstract_authenticator.py
```diff
@@ -20,6 +20,7 @@
 class GitHubToken(BaseModel):
     token: str
     expires_at: Optional[str] = None
+    permissions: dict[str, str] | None = None
```
raskinmaya: `permissions` here is only populated for installation tokens; PAT always leaves it `None`. Not wrong, but it spreads app-install concepts onto the shared token type. Keeping permissions on the probe path only might be cleaner

integrations/github/github/probe/pat/probe.py
```diff
@@ -0,0 +1,151 @@
...
+        if not user_response.is_success:
+            await self.context.fail(
+                f"Failed to fetch authenticated user: GitHub API returned {user_response.status_code}"
+            )
+            return None
+        user = user_response.json()
+        if isinstance(user, dict) and isinstance(login := user.get("login"), str):
+            organizations.append(login)
+
+        url: str | None = f"{github_host}/user/orgs"
```
raskinmaya, line 109: When `organizations` is configured we still paginate all of `/user/orgs` instead of probing only configured orgs. Extra API calls and checks for orgs that won’t be synced anyways

integrations/github/github/probe/pat/probe.py
raskinmaya: Shallow probe doesn't need exporter parity, and keeping probe decoupled from resync logic makes sense for this PR. You can make a small shared-helper follow-up later if you'd like :)

integrations/github/github/clients/auth/github_app/app_authenticator.py
raskinmaya, line 1: Please add an intent file :)

integrations/github/integration.py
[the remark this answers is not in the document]
VenfiOranai: Truth, done

integrations/github/github/probe/pat/probe.py
VenfiOranai: Done

integrations/github/github/probe/app/probe.py
VenfiOranai: While feasible, we'd encounter an issue where they'd be missing orgs in the probe with no indication that they're missing any. Any thoughts on how we could handle that? Current idea I've got is to continue but to add a message on the general probe, something like "Failed to fetch all orgs, displaying for orgs that were successfully found"
Thoughts?

integrations/github/integration.py
VenfiOranai: Got the same issue here as I did there. We're saving something pretty minimal, and the duplication here is something I actively want. If they ever change their scopes, I wouldn't want the changes to have side effects on unrelated kinds.

integrations/github/github/probe/pat/permissions.py
VenfiOranai: Not sure I really get what you're saying, adding an if/else instead of the current method feels like it would harm readability compared to how it is now

integrations/github/github/clients/auth/abstract_authenticator.py
VenfiOranai: Not sure I get the suggestion here. Like, the alternative I've got would be to duplicate a bunch of code which I imagine isn't what you're aiming for

integrations/github/github/probe/pat/probe.py
VenfiOranai: Added

integrations/github/github/clients/auth/github_app/app_authenticator.py
VenfiOranai: Done

integrations/github/github/probe/pat/probe.py
VenfiOranai: Currently to my understanding we're not supposed to use mapping at all, even for post install (if you recall I asked Maya/Geva about enrichments and they said to keep it minimal). Seeing as organizations comes from that, we won't support that for now.

integrations/github/github/clients/auth/abstract_authenticator.py
raskinmaya: I don't think we need to duplicate the token type. `permissions` only gets set for installation tokens, and only the app probe reads it - PAT always leaves it `None`. Wouldn't it be cleaner to keep that on the app path (e.g. a `GitHubInstallationToken` or read it from the `access_token`s response in the probe) so shared `GitHubToken` stays token + expiry?

integrations/github/github/probe/app/probe.py
raskinmaya: Yeah, skipping silently is what I want to avoid. Could we still add scopes for every org, and for the installation that failed `get_token` mark those checks as `FAILURE` with the error (including the org name)? Then the other orgs still get permission results, and the broken one is visible instead of a generic “failed to fetch all orgs” banner. Does it make sense to only `fail()` the whole probe if every installation fails?

raskinmaya (approved):

integrations/github/github/probe/app/probe.py
VenfiOranai: Fixed

PR #3918, opened 2026-09-03
[Core] Support status labels on action and workflow node runs

dormeiri-port:
# Description

**What** - Adds an optional `status_label` to run logs and status updates, for both action runs and workflow node runs.

**Why** - A long-running action hands work off to an external system, leaving the Port run sitting in progress with no indication of the stage it reached. Port supports a `statusLabel` on runs for exactly this, but Ocean had no way to set it.

**How** -

- `post_action_run_log` sends `statusLabel` in the log body. `post_run_log`, `update_run_started` and `report_run_completed` take an optional `status_label`, omitted from the payload when absent, so existing callers are unaffected.
- The two run kinds accept the label in different places and the client hides that: action runs take it on the log entry, workflow node runs only on the run. So `post_run_log` follows the log with a `statusLabel` patch for workflow node runs, while `update_run_started` and `report_run_completed` fold it into the patch they already send.
- Failures are reported by the execution manager rather than the executor, so `ActionExecutionError` carries a `status_label` with an overridable `DEFAULT_STATUS_LABEL`, and the manager forwards it.
- Adds a `Waiting for rate limit` label to the manager's existing backoff log.

## Stack

This is the base of a stack. The integrations adopt the new API in sibling PRs, each based on this branch:

- **#3918 (this PR)** — core `status_label` support
- #3919 — Github
- #3920 — Gitlab
- #3921 — Azure Devops
- #3922 — Cursor Cloud Agents
- #3923 — Claude Managed Agents

The integration PRs depend on this one being released, so merge this first.

Release intent in `.ocean-release/core/action-run-status-labels.yaml` (`minor` / `feature`).

## Type of change

### Core testing checklist

- [ ] Verified on a testing org that labels appear on the run and update as the action progresses

### Preflight checklist

- [x] Handled rate limiting (label added to the existing backoff path)


port_ocean/core/handlers/actions/execution_manager.py
```diff
@@ -539,13 +546,19 @@ async def _execute_run(self, run: IntegrationRun) -> None:
                 except ActionExecutionError as e:
                     logger.warning("Action run failed: {}", str(e))
                     error_summary = str(e)
+                    error_status_label = e.status_label or DEFAULT_FAILURE_STATUS_LABEL
                 except Exception as e:
                     logger.exception("Error executing run", error=str(e))
                     error_summary = f"Failed to execute run: {str(e)}"
+                    error_status_label = DEFAULT_FAILURE_STATUS_LABEL
```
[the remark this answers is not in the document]
dormeiri-port: Good catch that this makes the failure label unconditional, but I've kept the fallback deliberately and documented why instead.

A run keeps the last label it was given — both completion paths only include `statusLabel` when it's truthy, so passing `None` omits the field and preserves the stored value. An executor that logs `Dispatching workflow` before handing off and then hits an unexpected exception would end up with a run whose status is failed but whose label still reads `Dispatching workflow`. A redundant generic label is noise; a stale in-progress label on a failed run is wrong, so the generic fallback is the safer default.

On the behavior-change concern: all five integrations that implement actions adopt labels in the PRs stacked on this one, so the set of integrations that would newly show a generic failure label is effectively empty.

The client contract you're referring to is unchanged — `post_run_log`, `update_run_started` and `report_run_completed` still omit `statusLabel` when it's absent. What was missing is that the manager's policy of always supplying one on failure was intentional but undocumented, which is what made this look like an accident. Fixed in 5072e28:

- a comment at the fallback explaining the stale-label reasoning
- `ActionExecutionError`'s docstring now says an unset label falls back to a generic one, so authors know to set `DEFAULT_STATUS_LABEL` for something more useful
- the same note in `docs/framework-guides/docs/framework/features/actions.md`
- a named test, `test_execute_run_labels_failures_that_supply_no_label`, covering both the unlabeled `ActionExecutionError` and the unexpected-exception paths, so the intent is pinned rather than incidental

port_ocean/clients/port/mixins/actions_and_workflow_runs.py
```diff
@@ -137,35 +154,44 @@ async def update_run_started(
         link: str,
         external_id: str,
         extra_output: dict[str, Any] | None = None,
+        status_label: str | None = None,
     ) -> None:
         """Update a run to indicate it has started, setting the link and external ID."""
+        patch: dict[str, Any]
```
omby8888: no need for that to be outside the closure really

port_ocean/core/handlers/actions/execution_manager.py
```diff
@@ -529,6 +535,10 @@ async def _execute_run(self, run: IntegrationRun) -> None:
                     raise
 
                 error_summary: str | None = None
+                # A run keeps the last label it was given, so a failure always
+                # carries one: omitting it would leave an in-progress label such
+                # as "Dispatching workflow" on a run that has already failed.
+                error_status_label: str | None = None
```
omby8888: why not having this variable default to `DEFAULT_FAILURE_STATUS_LABEL`, then this comment is not needed

omby8888 (changes requested):

port_ocean/core/handlers/actions/execution_manager.py
dormeiri-port: Agreed, done in 238def13.

`error_status_label` is now `str` initialized to `DEFAULT_FAILURE_STATUS_LABEL`, so the invariant is in the type rather than in a comment, and the unexpected-exception branch no longer needs an assignment at all. The `ActionExecutionError` branch just overrides the default when the error carries its own label.

The stale-label rationale still lives in `ActionExecutionError`'s docstring, the framework docs, and the name of `test_execute_run_labels_failures_that_supply_no_label`, so it is not lost with the comment gone.

port_ocean/clients/port/mixins/actions_and_workflow_runs.py
dormeiri-port: Done in 7913e4fd — the annotation moved inline onto the first assignment, so the standalone line is gone.

It could not be dropped entirely: the two branches build differently shaped dicts, so mypy binds `patch` to the workflow-node branch's inferred type and then rejects the action-run assignment. Annotating the first assignment as `dict[str, Any]` covers both, which is what `report_run_completed` a few lines below already does — this method was just inconsistent with it.

omby8888 (approved):

PR #3924, opened 2026-09-03
[Skill] Add create-integration-action skill

dormeiri-port:
# Description

**What** - Adds a `create-integration-action` project skill, for adding an integration action to an integration that already exists. Sits alongside the existing `create-ocean-integration` skill, which covers building a whole integration.

**Why** - Five integrations now implement actions, and the conventions live only in those implementations. More importantly, the actions framework has a few behaviors that are easy to get wrong and that `AbstractExecutor`'s own docstring actively contradicts, so anyone (or any agent) starting from that docstring writes a broken action.

**How** - A ten-step workflow in `SKILL.md`, with the contract stated up front rather than buried:

- **The manager reports failures, not successes.** `_execute_run` calls `report_run_completed(success=False)` when `execute` raises, and does nothing when it returns. A synchronous action must therefore complete its own run, and an asynchronous one must deliberately leave it in progress for its webhook. The docstring instead shows success being reported via `patch_run` with `RunStatus.SUCCESS`.
- **`WEBHOOK_PROCESSOR_CLASS` is only an annotation** on `AbstractExecutor` with no default, and `register_executor` reads it directly, so a sync executor that omits it raises `AttributeError` at registration. It has to be set to `None`.
- **The docstring's `PARTITION_KEY` attribute does not exist** — the real API is an async `_get_partition_key` method — and it refers to `.port/spec.yaml` when all five integrations with actions use `.port/spec.json` with `actionsProcessingEnabled`.
- Status labels are held to two words, matching the convention in #3918.

Two reference files cover the fiddly areas: `async-completion.md` for the external-id correlation between an executor and its completion webhook (the shared `build_external_id` helper, the four guards in `handle_event`, and how `ACTION` processors bypass kind and mapping resolution), and `testing.md` for the mocking approach, which patches `ocean` on the module under test rather than globally because each module holds its own reference from the import.

Every framework claim was checked against source rather than taken from the docstrings, and every referenced path and internal link resolves.

Not applicable - this adds only `.cursor/skills/`, and touches no integration or `port_ocean/` code, so there is no release intent file.

## Type of change

- [x] Documentation (added/updated documentation)

## Notes for reviewers

Worth a skim by anyone who has written an action: if a convention here does not match what you would do, the skill is what future authors will follow, so I would rather correct it now.

---

> [!NOTE]
> **Low Risk**
> Documentation-only under `.cursor/skills/`; no integration or `port_ocean` runtime changes.
> 
> **Overview**
> Adds a new **create-integration-action** Cursor skill under `.cursor/skills/`, complementing **create-ocean-integration** for teams that only need a new action on an existing integration.
> 
> **SKILL.md** is a ten-step workflow (placement, sync vs webhook completion, `.port/spec.json`, executors, registration, tests, release intent) with an upfront **contract** section: success completion is the executor’s job, failures should be raised (not double-reported), `WEBHOOK_PROCESSOR_CLASS` must be set (including `None` for sync), stale `AbstractExecutor` docstring pitfalls, and two-word status labels.
> 
> **async-completion.md** documents webhook-based completion: shared `build_external_id`, `ACTION` vs catalog processors, guard patterns in `handle_event`, registration via the executor, and the known race with `update_run_started`.
> 
> **testing.md** documents executor and action webhook processor tests, especially patching `ocean` on the module under test and asserting external-id strings on `update_run_started`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e57729f926d744102b09f46c422523c830e49fb3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


omby8888 (approved):
LGTM

PR #3925, opened 2026-09-03
[CI] Fix notify-on-external-contribution workflow

MPTG94:
## Summary
- Skip notifying when the PR author, reviewer, or commenter is an org member (their activity doesn't need external-contribution attention)

## Test plan
- [ ] Confirm `ORG_MEMBERSHIP_TOKEN` secret is set (fine-grained PAT, org resource, Members: Read-only)
- [ ] Open a PR from a fork and confirm the webhook fires
- [ ] Have an org member comment/review on an external PR and confirm no webhook fires
- [ ] Have an external contributor comment on their own PR and confirm the webhook fires


.github/workflows/notify-on-external-contribution.yml
raskinmaya: This comment is wrong now - we do care about the commenter/reviewer for non-`pull_request_target` events

.github/workflows/notify-on-external-contribution.yml
```diff
@@ -82,6 +86,22 @@ jobs:
               return;
             }
 
+            if (eventName !== 'pull_request_target') {
+              const actor =
+                eventName === 'pull_request_review' ? payload.review.user.login : payload.comment.user.login;
+              let actorIsMember;
+              try {
+                actorIsMember = await isOrgMember(actor);
```
raskinmaya: If `actor === user`, we already know they're not a member from the author check above, no? Maybe add a `if (actor !== user)` guard here? External contributors commenting on their own PR is probably the most common comment path, and this will save an API call every time

.github/workflows/notify-on-external-contribution.yml
```diff
@@ -58,21 +58,25 @@ jobs:
...
+              try {
+                await orgClient.rest.orgs.checkMembershipForUser({ org, username });
+                return true;
+              } catch (err) {
+                if (err.status === 404) return false;
+                throw err;
+              }
+            }
+
             let isExternal;
```
raskinmaya: `isExternal` only describes the author at this point, but we also check the actor below. `authorIsExternal` would make the two checks easier to follow IMO

MPTG94: Changed to `prAuthorIsExternal`

.github/workflows/notify-on-external-contribution.yml
MPTG94: Good call, added

.github/workflows/notify-on-external-contribution.yml
MPTG94: Changed to `prAuthorIsExternal`

MPTG94 (commented):
Fixed CR comments

raskinmaya (approved):

